PR #4, opened 2021-05-27
Fixed patient form editing not working.

manuelroemer:
The patient registration form was not functional as it didn't display the current values of a patient at the moment.
The changes in here (re-)enable that functionality, at least partially. 

Required for future work on making the edit form available offline.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
@@ -25,16 +25,15 @@ import {
 import { DummyDataInput } from './input/dummy-data/dummy-data-input.component';
 import { useTranslation } from 'react-i18next';
 import { getSection } from './section/section-helper';
-import { cancelRegistration, parseAddressTemplateXml, scrollIntoView } from './patient-registration-utils';
+import {
+  cancelRegistration,
+  parseAddressTemplateXml,
+  scrollIntoView,
+  getFormValuesFromFhirPatient,
+  getAddressFieldValuesFromFhirPatient,
+  getPatientUuidMapFromFhirPatient,
+} from './patient-registration-utils';
 import { ResourcesContext } from '../offline.resources';
-
-const initialAddressFieldValues = {};
-
-const patientUuidMap: PatientUuidMapType = {
-  additionalNameUuid: undefined,
-  patientUuid: undefined,
-  preferredNameUuid: undefined,
-};
 
 const blankFormValues: FormValues = {
   givenName: '',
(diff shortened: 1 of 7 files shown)


manuelroemer:
There are already some preparations for offline editing in here. These are not the focus of this PR and can absolutely change again in following PRs.

The focus of this one is mainly making the form work again when _online_.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -126,123 +131,62 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
...
-        });
-      }
-
-      if (patient.deceasedBoolean || patient.deceasedDateTime) {
-        initialFormValues.isDead = true;
-        initialFormValues.deathDate = patient.deceasedDateTime ? patient.deceasedDateTime.split('T')[0] : '';
-      }
+    if (patient) {
+      setPatientUuidMap({ ...patientUuidMap, ...getPatientUuidMapFromFhirPatient(patient) });
+      setInitialAddressFieldValues({ ...initialAddressFieldValues, ...getAddressFieldValuesFromFhirPatient(patient) });
```
manuelroemer, line 136: Honestly, I'm not quite sure whether all of this object merging in this file always leads to the correct form being displayed in the end.
Data gets fetched/merged all over the place. I think we should consider reworking this in the future.

For the moment, the current version should reflect the *_previous_* version of the code, just extracted into different files, i.e. I tried to not actually change (too much) behavior here.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -126,123 +131,62 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
...
     const { addressFieldValues, addressValidationSchema } = parseAddressTemplateXml(addressTemplateXml);
+    setValidationSchema((validationSchema) => validationSchema.concat(addressValidationSchema));
 
-    for (const { name, defaultValue } of addressFieldValues) {
-      if (!initialAddressFieldValues[name]) {
-        initialAddressFieldValues[name] = defaultValue;
+    // `=== false` is here on purpose (`inEditMode` is a triple state value).
+    // We *only* want to set initial address field values when *creating* a patient.
+    // We must wait until after loading for this info.
+    if (inEditMode === false) {
```
manuelroemer, line 174: This is the only exception to my previous comment.
Overwriting patient address data from the backend with _default_ values seemed wrong to me. I can remove the `if` here though if that's a wrong assumption on my side.

FlorianRappl (approved):
LGTM - thanks! :beers:

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
FlorianRappl: Yeah, personally, I think this MF has a lot of dodgy behavior. Cleaning it up would be wonderful. If we break something on the way its fine. We can then repair it.

PR #5, opened 2021-05-31
Offline Patient Editing

manuelroemer:
Adapts the offline code to support editing patients while offline.

* Patients which have been created while offline can be edited.
* Patients which have been created while online can be edited after being cached.

### DIFF
--- a/packages/esm-patient-registration-app/package.json
+++ b/packages/esm-patient-registration-app/package.json
@@ -34,7 +34,6 @@
     "url": "https://github.com/openmrs/openmrs-esm-patient-management/issues"
   },
   "dependencies": {
-    "@types/uuid": "^8.3.0",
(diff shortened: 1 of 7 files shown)


packages/esm-patient-registration-app/package.json
```diff
@@ -34,9 +34,11 @@
     "url": "https://github.com/openmrs/openmrs-esm-patient-management/issues"
   },
   "dependencies": {
+    "@types/uuid": "^8.3.0",
```
FlorianRappl, line 37: Not a `dependency`. Its a `devDependency`.

FlorianRappl (commented):
I think this is almost done. Great job!

FlorianRappl (approved):

PR #6, opened 2021-06-15
add offline patient list data

nmalyschkin:


packages/esm-patient-list-app/src/offline/offlineData.ts
```diff
@@ -0,0 +1,232 @@
...
+// Declare Database
+class OfflinePatientDatabase extends Dexie {
+  private offlinePatients: Dexie.Table<OfflinePatient, number>; // id is number in this case
+  private currentUser = 'unauthorized_user'; // will be automatically set to current user if logged in
+  private changeSubscribers = new Set<(data: Array<OfflinePatient>) => void>();
+  private broadcastChange = () => {}; // is overwritten if the Broadcastchannel is available
+
+  public constructor() {
+    super('OfflinePatientDatabase');
+    this.version(4).stores({
```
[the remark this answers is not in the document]
nmalyschkin: This was incremented while developing. I will reset it before it gets merged.

packages/esm-patient-list-app/src/offline/offlineData.ts
```diff
@@ -0,0 +1,232 @@
...
+      offlinePatients: '++id, uuid',
```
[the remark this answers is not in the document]
nmalyschkin: I was wrongfully under the impression, that the id as the key path is obligatory. Removed the id.

packages/esm-patient-list-app/src/offline/types.ts
```diff
@@ -0,0 +1,43 @@
+export enum LOAD_STATE_TYPE {
```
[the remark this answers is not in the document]
nmalyschkin: Not a big fan of the implicity. Also it bugs me to have strings all over my switch cases and stored data.
```ts
switch (prevState.type) {
  case 'LOADING':
    newLoadState = { type: 'LOADING_ERROR', error };
    break;

  case 'RELOADING':
    newLoadState = { type: 'RELOADING_ERROR', error, date: prevState.date };
    break;

  default:
    return;
}
```
But I changed it to your preferred style.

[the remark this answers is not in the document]
FlorianRappl: 1. You don't need a string.
2. The enum you use is a lot less efficient (its a runtime type, not a compile-time type) and refactors the same. After all - its still JS. Strings are everywhere (was before TS, and is still the preferred / used style even with TS).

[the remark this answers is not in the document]
FlorianRappl: Also just to emphasize: The given construct was a suggestion, not an obligation. That's why I started with "Personally".

packages/esm-patient-list-app/src/Modal.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import style from './style.scss';
+
+const Modal: React.FC<{ close: () => void }> = ({ children, close }) => {
```
FlorianRappl, line 5: This is generic / React independent enough to actually make this implementation w/o React and put it in styleguide.

Maybe it should also work a bit like toasts. There should be a global container and a global function to just add / show some modal dialog.

nmalyschkin: Toasts in `styleguide` are implemented with React, is there a reason not to use React for the modal as well?

Would it make sense to use the extension system to attach/detach a modal to a slot in a global container?

FlorianRappl: They are now. But they should not be.

> is there a reason not to use React for the modal as well?

The reason is that `styleguide` is framework independent. And for such elementary things we should actually be agnostic, too.

> use the extension system to attach/detach a modal to a slot in a global container?

We could, but I think it would not give us much advantages (this is, after all, done for handling workspaces, but here we required a loosely coupled mechanism). Why not mirror `showToast` and have `showModal` as an import from `esm-framework`?

nmalyschkin: -----BEGIN OPINION-----
We should commit to a main framework (probably react) that we use in central components like toast and modals. Staying impartial just for the sake of it just slows down progress. As long as other frameworks can be integrated it should not make any difference to the integration developer. I mean `esm-framework` is already bringing the react-utils.
-----END OPINION-----

> Why not mirror showToast and have showModal as an import from esm-framework?

I was thinking about that but because the modal has slightly different requirements, it gets a little more complicated. Mainly the problem is, that the content in the modal should be able to close the modal (e.g. a cancel button). If we stick to react, we could make use of a the context API, otherwise we need to pass props upwards making it more complicated to use than it already is.

```tsx
const MyModal = withModalContext(({ close, data }) => {
  return (
    <div>
      <p>content...</p>
      <button onClick={close}>finish</button>
    </div>
  );
});

const TestComponent = () => {
  const [data, setData] = React.useState<any>();
  const openModal = React.useCallback(() => {
    showModal(<MyModal data={data} />);
  }, [data]);

  return <button onClick={openModal}>open here</button>;
};
```
What do you think of this kind of interface?

FlorianRappl: The modal needs to be usable from any framework.

I don't see the problem with a cancel button. When you `showModal` you get back a disposer that can cancel it. Problem solved. There is no need for a React specific solution here.

> We should commit to a main framework (probably react) that we use in central components like toast and modals.

The problem is that we don't have React there. The architecture of this application is different. This is not a Piral instance and there is no main framework. `react-utils` is just a lib, there is no React DOM yet. Any React DOM is spawned at the position of a component. There is no root React DOM, which is why things are inefficient already and why we should not make things slower (which is why `Toast` should not remain written in React).

nmalyschkin: Okay, how do you propose to make it usable from any framework, how do you imagine the interface to look like? With toasts it's rather easy as long as I only pass down strings. With modals this doesn't fly.

FlorianRappl: `render(container: HTMLElement): Dispose` sounds good to me.

nmalyschkin: Could you give an example like above, with how it would integrate in a minimal use-case with let's say react (as it is our most used framework).

FlorianRappl: If it helps you:

```jsx
const renderMyModal = (container, { close, data }) => {
  ReactDOM.render(
    <div>
      <p>content...</p>
      <button onClick={close}>finish</button>
    </div>,
  container);
  return () => {};
};

const TestComponent = () => {
  const [data, setData] = React.useState();
  const openModal = React.useCallback(() => {
    const props = { data };
    props.close = showModal(container => renderMyModal(container, props));
  }, [data]);

  return <button onClick={openModal}>open here</button>;
};
```

nmalyschkin: Don't you think that this I a little too over complicated for the average developer? There is so much area for error here, for example in the above code the close button does not work because at the moment of rendering `close` is undefined (assuming `showModal` does not defer internally). You and I know that we can defer the rendering here to solve this but others wouldn't.

FlorianRappl: OK so you ask for *an* example but then just to take that as a final implementation (w/o any abstraction that we usually introduce for React).

> You and I know that we can defer the rendering here to solve this but others wouldn't.

I think you have too many assumptions here - please don't make assumptions if you don't know the implementation. The problem you describe will not be present as rendering always happens *after* you call *showModal*. Therefore, the close function is available and nothing needs to be deferred.

All in all I feel that this discussion has stopped being productive a while ago. I think already after I presented the render signature the deal was sealed; you just wanted to see some arbitrary example to take it as a final implementation (with assumptions) for actually doing remarks.

That being said the wrapper we have here is essentially already a non-React code. Using context or anything similar makes things worse. The dialog handling should be generic and don't rely on a particular framework. It's as simple as that.

If you have further questions or want to discuss things don't hesitate to call me!

PR #7, opened 2021-06-17
MF-623: desktop search bar

seankwon:
# Description

Styles for desktop search bar (if row reverse)

Screen Shot 2021-06-16 at 8 27 32 PM


FlorianRappl (approved):
LGTM! Thanks :beers:

PR #9, opened 2021-06-17
Fix search input height

donaldkibet:
### Description

Provide carbon `Search` component with the prop size. By default, it should assume `size='xl'` seems like by default it assumes the smallest size as shown [here](https://www.carbondesignsystem.com/components/search/code) am setting it explicitly to `xl`

### Screenshoot

before 

Screenshot 2021-06-17 at 16 01 24

after 

Screenshot 2021-06-17 at 16 00 29


FlorianRappl (approved):
LGTM :beers:

PR #13, opened 2021-06-25
Fixed the spacing of the 'Create New Patient header' and the navbar

vasharma05:
In this PR, I have fixed the top margin of the left column in the patient registration.

Before: 
image

After: 
image

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.scss
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.scss
@@ -28,9 +28,11 @@
     margin: $spacing-05 0 $spacing-05 0;
 }
 
-.fixedColumn {
-    position: fixed;
+.stickyColumn {
+    position: sticky;
     margin-top: 2rem;
+    // 3rem for the nav height and 2rem for top margin
+    top: 5rem;
 }
 
 $color-gray-70: #525252;
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.scss
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.scss
@@ -28,8 +28,9 @@
     margin: $spacing-05 0 $spacing-05 0;
 }
 
-.fixedPosition {
+.fixedColumn {
     position: fixed;
+    margin-top: 2rem;
 }
 
 $color-gray-70: #525252;
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
@@ -224,7 +224,7 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
             <Grid>
               <Row>
                 <Column lg={2} md={2} sm={1}>
-                  <div className={styles.fixedPosition}>
+                  <div className={styles.fixedColumn}>
                     <h4>
                       {inEditMode ? t('edit', 'Edit') : t('createNew', 'Create New')} {t('patient', 'Patient')}
                     </h4>
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
@@ -223,8 +223,8 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
             />
             <Grid>
               <Row>
-                <Column lg={2} md={2} sm={1}>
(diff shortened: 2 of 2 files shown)


HerbertYiga:
cc @FlorianRappl

packages/esm-patient-registration-app/src/patient-registration/patient-registration.scss
```diff
@@ -28,8 +28,9 @@
     margin: $spacing-05 0 $spacing-05 0;
 }
 
-.fixedPosition {
+.fixedColumn {
     position: fixed;
```
FlorianRappl, line 32: Should not be position fixed. Please remove.

vasharma05: Sure!

vasharma05: @FlorianRappl the left column needs to be `position:fixed`, because the right column, i.e. the form is scrollable and the left column is kind of navigator for sections on the right form.
image

FlorianRappl: I don't think it does.

1. There is `position: sticky`
2. You can give the outer container full height,but limited to full height (i.e., not scrollable) - then have the right box full the height with an auto scrollable

This way there is no position fixed and it just works.

vasharma05: > There is position: sticky
I just tried applying position sticky, but it didn't work. 
Btw, `position: fixed` was already there in the code, I didn't add it.
> You can give the outer container full height,but limited to full height (i.e., not scrollable) - then have the right box full the height with an auto scrollable

In this way, it will give a scrollbar to the right container, which is not expanded to full width and won't that look bad?
image
As you can see in the image, the outer container of the right column is only limited to the grey part. What I can do is expand it to full width. Do you want me to go with this?

FlorianRappl: > Btw, position: fixed was already there in the code, I didn't add it.

Tell me something I don't know :)

FlorianRappl: Sorry - the gray box you see is not the right element I refer to. I always refer to logical elements, not physical ones...

So when I talk about the right box I mean the container that hosts the gray box. And, of course, this container takes the remaining space (vertical and horizontal) from the outer container. Therefore, for me, there was no question about whether or not take the full width. It will always be taken.

vasharma05: Ohh okay, got it.  I'll do it.
Thanks!

vasharma05: https://user-images.githubusercontent.com/51502471/123639650-0b0e7180-d83e-11eb-982e-60f589b5d9ba.mp4

vasharma05: The blank space above is of the navigation bar, which is not visible due to some error.
Is this good @FlorianRappl ?

FlorianRappl: Hm, in the video I don't see a real scrollbar. The scrollbar I see is not scrolling and is at 100%.

vasharma05: image
This is the scrollbar. Actually `Grid` is being used  which doesn't take up the full width in the desktop view and this is what I asked that time, to extend it to full width,

What I can do is remove the Grid and add my own flex to the container. What you say about it?

vasharma05: It wasn't looking good and hence I removed the scrollbar.

FlorianRappl: > What I can do is remove the Grid and add my own flex to the container. What you say about it?

Sounds good to me.

FlorianRappl: > It wasn't looking good and hence I removed the scrollbar.

There should be just one scrollbar - and it should look as expected (i.e., if there an overflow the scrollbar should indicate that).

vasharma05: Got it! I'll do it.
Thanks!

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #14, opened 2021-06-28
MF-638: Fix design issues in the patient registration form

vasharma05:
### What does this PR do?
This PR works on the design fixups in the patient-registration form.

1. Adding Breadcrumbs
2. Fixing the structure for the form as per the mockup designed.
3. Add space below the form.
4. Changing the label of postcode


vasharma05:
The new design for the patient registration looks as follows:
image
image

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
```diff
@@ -27,8 +27,8 @@ export const esmPatientRegistrationSchema = {
       },
     },
     _default: {
-      demographics: { name: 'Basic Info', fields: ['name', 'gender', 'dob', 'id'] },
-      contact: { name: 'Contact Details', fields: ['address', 'phone', 'email'] },
+      demographics: { name: 'Basic Info', fields: ['name', 'gender & dob', 'id'] },
```
FlorianRappl, line 30: What if I want only gender? Or only date of birth?

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
```diff
@@ -27,8 +27,8 @@ export const esmPatientRegistrationSchema = {
...
+      contact: { name: 'Contact Details', fields: ['address', 'phone & email'] },
```
FlorianRappl, line 31: In general, I think the configuration option(s) here seem to make almost no sense - at least when looking at the very static and fixed design.

We need to ping this back.

What makes sense here regarding configurability? How is this reflected in the design? We need this to be answered.

packages/esm-patient-registration-app/src/patient-registration/input/basic-input/select/select-input.component.tsx
```diff
@@ -17,7 +17,7 @@ export const SelectInput: React.FC<SelectInputProps> = ({ name, options, label }
   ];
 
   return (
-    <div>
+    <div style={{ maxWidth: '360px' }}>
```
FlorianRappl, line 20: Belongs to CSS.

The max-width should rather be 100%. The width may be 360px.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -222,6 +224,13 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
               }}
             />
             <Grid>
+              <Row>
+                <ExtensionSlot extensionSlotName="breadcrumbs-slot">
+                  <div className={styles.breadcrumbsContainer}>
```
FlorianRappl, line 229: Not sure - why do we need this container?

FlorianRappl: (The breadcrumbs should look OK on any page when placed there; if not we should alter the default styling.)

FlorianRappl (commented):
Changes look fine so - just the questions that came up need to be addressed imho asap.

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
vasharma05: The mockup wanted them to be in the same section under `Sex & Birth`. I'll confirm from Ciaran about this and will respond back asap.

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
vasharma05: These are being used to show the fields in the form. The form is dynamic in relation to the fields on the basis of these configuration. You might know about this before, so maybe this is not what you wanted as an answer. 
Who should I contact to clarify this properly?

packages/esm-patient-registration-app/src/patient-registration/input/basic-input/select/select-input.component.tsx
vasharma05: Ohh okay, got it! I'll push the changes.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
vasharma05: The breadcrumbs have a grey background, and I wanted the background to be transparent. 
Yes I also believe that the breadcrumbs should be transparent in nature, the grey background should not be default. 
Also about the padding, the default padding is 2rem, but I believe the padding should be `2rem 0` to be default, so that the user can align it on it's own, wherever needed.

FlorianRappl: The padding we should keep as is (after all, this is the choice from the Carbon design language; which we inherit). The background, however, does not make much sense. We should override this in the `esm-core` repository and avoid wrapping the breadcrumbs here.

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
FlorianRappl: This is not about the *current* design, but about the *configurability*.

FlorianRappl: Just to be clear: I am not questioning the *current* design, I am questioning why we have *configurability* here - my questions above should only point out that these configuration options make no sense as they don't even allow such basics things to be done.

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
FlorianRappl: > These are being used to show the fields in the form. The form is dynamic in relation to the fields on the basis of these configuration.

Tell me something I don't know.

> Who should I contact to clarify this properly?

Every stakeholder. This should be a general discussion point about configuration options. Right now, some options are either

1. too far fetched and will never be used
2. too complicated and will never be used
3. too limited and will result in frustration
4. conflicting with the design choices (this is a good example)

The problem is - and I've mentioned that several times already - that the design is done as a monolith. It does not care about

- the micro frontend nature of the application (i.e., everything can grow; even without us doing something)
- the configuration possibilities
- the potential design overrides (ex. Brandon's custom navbar)

But these things need to be on the table. They need to be very present. They are at the heart of the application and dismissing them from the design is an issue.

cc @jonathandick

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
vasharma05: I am sorry, but I don't have much idea about the configurations, so should I keep it to discuss it on the talk on Thursday?

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
vasharma05: Duly Noted, I'll push the required changes in the `esm-core`.
Thanks alot!

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
```diff
@@ -24,6 +24,7 @@ export const AddressField: React.FC = () => {
 
   useEffect(() => {
     const templateXmlDoc = parseString(addressTemplateXml);
+    console.log(templateXmlDoc);
```
FlorianRappl, line 27: Remove please

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
FlorianRappl: Let's just leave it like that. I'll take care of it later.

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
vasharma05: Sorry. Removed it.

packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts
vasharma05: Then I should keep it the way it is or revert back all the config changes?

FlorianRappl: Keep it as you have it now.

FlorianRappl (approved):

PR #15, opened 2021-06-30
MF-615: Active Visits Widget

vasharma05:
### What does this PR include?
This PR comprises of 2 sections:

1. Creating a new module `@openmrs/esm-active-visits-app` in the monorepo.
2. Active Visits widget

Before merging this PR, please merge [PR on home app to add extension slot for home page widget](https://github.com/openmrs/openmrs-esm-home/pull/66).

PR #17, opened 2021-07-10
Improvements in Active VIsits Widget

vasharma05:
### What does this PR include?

This PR is in continuation to #15 with some improvements mentioned below.
1. Visual Improvement
2. Search added
3. Empty state added

image
image
image

### DIFF
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useMemo, useEffect, useState } from 'react';
 import DataTable, {
   TableContainer,
   Table,
@@ -7,9 +7,12 @@ import DataTable, {
   TableHeader,
   TableBody,
   TableCell,
+  TableToolbar,
+  TableToolbarContent,
 } from 'carbon-components-react/es/components/DataTable';
 import DataTableSkeleton from 'carbon-components-react/es/components/DataTableSkeleton';
 import Pagination from 'carbon-components-react/es/components/Pagination';
+import Search from 'carbon-components-react/es/components/Search';
 import { useLayoutType, useConfig, usePagination } from '@openmrs/esm-framework';
 import { ActiveVisitRow, fetchActiveVisits } from './active-visits.resource';
 import styles from './active-visits.scss';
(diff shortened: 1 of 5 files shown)


packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -83,56 +90,69 @@ const ActiveVisitsTable = (props) => {
...
     });
     return () => activeVisits.unsubscribe();
   }, []);
-  return (
+
+  return !loading ? (
     <div className={styles.activeVisitsContainer}>
       <div className={styles.activeVisitsDetailHeaderContainer}>
-        <h4 className={styles.productiveHeading02}>{t('activeVisits', 'Active Visits in Clinic')}</h4>
+        <h4 className={styles.productiveHeading02}>{t('activeVisits', 'Active Visits')}</h4>
```
FlorianRappl, line 104: Changed here but not in the actual language file.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.scss
```diff
@@ -2,7 +2,8 @@
 
 .activeVisitsContainer {
   background-color: $ui-background;
-  width: 100%;
+  // feedback needed
```
FlorianRappl, line 5: Yes. Why 90? Why `vw`?

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
vasharma05: Ohh yeah, Thanks!

packages/esm-active-visits-app/src/active-visits-widget/active-visits.scss
vasharma05: If you see the current widget on the home screen, the width changes according to the content in different pages and hence I wanted to fix the size. If I fix the size to 768px(tab width) then the date divided into 2 lines. I discussed it with JJ as in to what to do, he said, you can help in it and thought of keeping to the screen width and hence `90vw` .

FlorianRappl (approved):

packages/esm-active-visits-app/src/active-visits-widget/active-visits.scss
FlorianRappl: Yes, but 90vw does not make much sense, e.g., on mobile. You shouldn't fix the width and you should try to make a progressive design, i.e., one that pretty much *always* works.

So what if the nav bar is shown on the left side? What if the user has a wide screen (like 4k pixels wide)? 90vw does not sound right,.

vasharma05: Sure, I get you, let me fix that too.
Thanks!

FlorianRappl (approved):

PR #20, opened 2021-07-14
MF-647: Introduce PatientTable component

samuelmale:
An example usage of the PatientTable extension in a different MF:

### DIFF
--- a/packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
+++ b/packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
@@ -7,57 +7,99 @@ import DataTable, {
   TableHeader,
   TableRow,
 } from 'carbon-components-react/lib/components/DataTable';
-import React from 'react';
+import Pagination from 'carbon-components-react/lib/components/Pagination';
+import Search from 'carbon-components-react/lib/components/Search';
+import { debounce } from 'lodash';
+import React, { useMemo } from 'react';
+import styles from './patient-table.component.scss';
 
-const PatientTable: React.FC<{ patients: Array<Object>; columns: Array<PatientTableColumn> }> = ({
-  patients,
-  columns,
-}) => {
-  const rows = patients.map((patient) => {
-    const row = {};
-    columns.forEach((column) => {
-      row['id'] = patient['id'] || patient['uuid'];
-      row[column.key] = column.getValue ? column.getValue(patient) : patient[column.key];
-    });
-    return row;
-  });
+const PatientTable: React.FC<PatientTableProps> = ({ patients, columns, search, pagination }) => {
+  const rows: Array<any> = useMemo(
+    () =>
+      patients.map((patient) => {
+        const row = {};
+        columns.forEach((column) => {
+          row['id'] = patient['id'] || patient['uuid'];
+          row[column.key] = column.getValue ? column.getValue(patient) : patient[column.key];
+        });
+        return row;
+      }),
+    [patients, columns],
+  );
+  const handleSearch = React.useMemo(() => debounce((searchTerm) => search.onSearch(searchTerm), 300), []);
+
   return (
-    <TableContainer>
-      <DataTable rows={rows as any} headers={columns as any} isSortable={true} size="normal" useZebraStyles={true}>
+    <div>
+      <div id="table-tool-bar" style={{ display: 'flex', flexDirection: 'row-reverse' }}>
+        <Search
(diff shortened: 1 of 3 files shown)


samuelmale:
@nmalyschkin I think we can later on introduce the 'advance filters' and 'column(s) controller' widgets.

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
```diff
@@ -0,0 +1,140 @@
...
+import React, { useMemo, CSSProperties } from 'react';
+import styles from './patient-table.component.scss';
+
+const PatientTable: React.FC<PatientTableProps> = ({ patients, columns, search, pagination, isLoading, autoFocus }) => {
+  const rows: Array<any> = useMemo(
+    () =>
+      patients.map((patient) => {
+        const row = {};
+        columns.forEach((column) => {
+          row['id'] = patient['id'] || patient['uuid'];
```
nmalyschkin, line 25: This does not have to be done for every column.

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
```diff
@@ -0,0 +1,140 @@
...
+          const value = column.getValue ? column.getValue(patient) : patient[column.key];
```
nmalyschkin, line 26: `column.getValue?(patient) || patient[column.key]`

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
```diff
@@ -0,0 +1,140 @@
...
+        });
+        return row;
+      }),
+    [patients, columns],
+  );
+  const handleSearch = React.useMemo(() => debounce((searchTerm) => search.onSearch(searchTerm), 300), []);
+
+  return (
+    <>
+      {!isLoading ? (
```
nmalyschkin, line 47: You could use this notation:
```tsx
if(isLoading)
  return <Skeleton />

return <Table />
```

Not a big difference, but make it easier to read

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <>
```
nmalyschkin, line 48: Right now the PatientTable returns two separate components, should be one.

nmalyschkin (commented):
Great work! 👍

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
samuelmale: Did you mean like the Skeleton and the Table should be under the same parent component :

```jsx

return (
<div> 
   {isLoading && <Skeleton />}
   {!isLoading && <Table/>}
</div>)

```

Or you suggest that it should have the Skeleton?

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
samuelmale: Nice catch!

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
FlorianRappl: Not sure what Nikita was after here. You can leave it as is for now.

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
samuelmale: @nmalyschkin somehow my typescript compiler will complain. I think it's confusing this with being a ternary operator.

Screenshot 2021-07-19 at 10 27 49 (2)

samuelmale:
@FlorianRappl and @nmalyschkin updated PR

packages/esm-patient-list-app/src/PatientTable/patient-table.component.tsx
FlorianRappl: No. You have a syntax error. I think you meant `getValue?.(patient)`. What you wrote misses the alternative branch specified after an `:` (so TS is right).

samuelmale: Resolved!

FlorianRappl (approved):
LGTM - thanks! :beers:

PR #21, opened 2021-07-14
Minor fixups in active visits widget

vasharma05:
### What does this PR do?

I did a mistake regarding the patient link in my last PR #17 which is improved in this.
Along with that I changed the link color to blue.

image

### DIFF
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -124,12 +124,13 @@ const ActiveVisitsTable = (props) => {
                 </TableRow>
               </TableHead>
               <TableBody>
-                {rows.slice(lowerBound, upperBound).map((row) => (
+                {rows.slice(lowerBound, upperBound).map((row, ind) => (
                   <TableRow key={row.id} style={{ height: desktopView ? '2rem' : '3rem' }}>
-                    {row.cells.map((cell, ind) => (
+                    {row.cells.map((cell) => (
                       <TableCell key={cell.id}>
                         {cell.info.header === 'name' ? (
-                          <ConfigurableLink to={`\${openmrsSpaBase}/patient/${activeVisits[ind].patientUuid}/chart/`}>
+                          <ConfigurableLink
+                            to={`\${openmrsSpaBase}/patient/${activeVisits[lowerBound + ind].patientUuid}/chart/`}>
                             {cell.value}
                           </ConfigurableLink>
                         ) : (
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -74,11 +74,23 @@ const ActiveVisitsTable = (props) => {
   const pageSizes = config?.activeVisits?.pageSizes ?? [10, 20, 50];
   const [loading, setLoading] = useState(true);
   const [activeVisits, setActiveVisits] = useState<ActiveVisitRow[]>([]);
-  const { goTo, currentPage } = usePagination(activeVisits, currentPageSize);
-  const [lowerBound, upperBound] = useMemo(
-    () => [(currentPage - 1) * currentPageSize, (currentPage + 0) * currentPageSize],
-    [currentPage, currentPageSize],
-  );
+  const [searchString, setSearchString] = useState('');
+
+  const searchResults: ActiveVisitRow[] = useMemo(() => {
+    if (searchString && searchString.trim() != '') {
+      return activeVisits.filter((activeVisitRow) =>
+        Object.keys(activeVisitRow).some((header) => {
+          if (header === 'patientUuid') {
+            return false;
+          }
+          return `${activeVisitRow[header]}`.toLowerCase().includes(searchString.toLowerCase());
+        }),
+      );
+    } else {
+      return activeVisits;
+    }
+  }, [searchString, activeVisits]);
+  const { goTo, currentPage, results } = usePagination(searchResults, currentPageSize);
 
   useEffect(() => {
     const activeVisits = fetchActiveVisits().subscribe((data) => {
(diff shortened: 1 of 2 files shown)


packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -124,12 +124,13 @@ const ActiveVisitsTable = (props) => {
                 </TableRow>
               </TableHead>
               <TableBody>
-                {rows.slice(lowerBound, upperBound).map((row) => (
+                {rows.slice(lowerBound, upperBound).map((row, ind) => (
```
FlorianRappl, line 127: Why anyway this whole thing with `slice`? You have `usePagination`. It gives you already the subset. Use this.

vasharma05: I need all of the results when I will be applying searching. That's why I pass all the activeVisits in the line 106 to the datatable and whenever search happens, the corresponding rows are output which are then sliced.

If I use `usePagination` I get the sliced data which if passed in the datatable will implement the search through that specific sliced rows only.

FlorianRappl: No. You should have the search to look up in the activeVisits data set and supply the filtered data set to the `usePagination` hook.

vasharma05: But search is handled by the dataTable. If I give it a try by myself, I'll have to iterate through all the column to find the results.

vasharma05: If you want by that way, I'll implement it too. But my recommendation would be to go through the pre-built search by dataTable. Slicing by usePagination does the same thing it is implemented right now.

FlorianRappl: This is unnecessary and (as we can see) a source of bugs. Just implement it the right away.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -74,11 +74,23 @@ const ActiveVisitsTable = (props) => {
...
   const [activeVisits, setActiveVisits] = useState<ActiveVisitRow[]>([]);
-  const { goTo, currentPage } = usePagination(activeVisits, currentPageSize);
-  const [lowerBound, upperBound] = useMemo(
-    () => [(currentPage - 1) * currentPageSize, (currentPage + 0) * currentPageSize],
-    [currentPage, currentPageSize],
-  );
+  const [searchString, setSearchString] = useState('');
+
+  const searchResults: ActiveVisitRow[] = useMemo(() => {
+    if (searchString && searchString.trim() != '') {
```
FlorianRappl, line 80: `!==`

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -74,11 +74,23 @@ const ActiveVisitsTable = (props) => {
...
+      return activeVisits.filter((activeVisitRow) =>
+        Object.keys(activeVisitRow).some((header) => {
+          if (header === 'patientUuid') {
+            return false;
+          }
+          return `${activeVisitRow[header]}`.toLowerCase().includes(searchString.toLowerCase());
```
FlorianRappl, line 86: You could also make `const search = searchString.toLowerCase()` before  iterating over everything.  This way you only compute that part once.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -124,13 +136,12 @@ const ActiveVisitsTable = (props) => {
...
               <TableBody>
-                {rows.slice(lowerBound, upperBound).map((row, ind) => (
+                {rows.map((row, ind) => (
                   <TableRow key={row.id} style={{ height: desktopView ? '2rem' : '3rem' }}>
                     {row.cells.map((cell) => (
                       <TableCell key={cell.id}>
                         {cell.info.header === 'name' ? (
-                          <ConfigurableLink
-                            to={`\${openmrsSpaBase}/patient/${activeVisits[lowerBound + ind].patientUuid}/chart/`}>
+                          <ConfigurableLink to={`\${openmrsSpaBase}/patient/${results[ind]?.patientUuid}/chart/`}>
```
FlorianRappl, line 144: Isn't `results[index]` the same as `row`? Why not use `row`?

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
vasharma05: It wan't a bug, it was a mistake

FlorianRappl: Isn't almost every bug a mistake? But in any case - that's exactly my point.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
vasharma05: Because `row` contains those values of the headers. As you can see activeVisits have a property patientUuid, but since there's no header for patientUuid, hence value of patientUuid doesn't exist in `row`.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -103,15 +115,15 @@ const ActiveVisitsTable = (props) => {
...
+        {({ rows, headers, getHeaderProps, getTableProps, getBatchActionProps }) => (
           <TableContainer title="" className={styles.tableContainer}>
             <TableToolbar>
               <TableToolbarContent>
                 <Search
                   tabIndex={getBatchActionProps().shouldShowBatchActions ? -1 : 0}
-                  placeHolderText="Filter table"
-                  onChange={onInputChange}
+                  placeholder="Filter table"
+                  onChange={(e) => setSearchString(e.target.value)}
```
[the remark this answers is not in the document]
vasharma05: Sure!

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
vasharma05: Sure!

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -74,11 +74,23 @@ const ActiveVisitsTable = (props) => {
...
   const [loading, setLoading] = useState(true);
   const [activeVisits, setActiveVisits] = useState<ActiveVisitRow[]>([]);
-  const { goTo, currentPage } = usePagination(activeVisits, currentPageSize);
-  const [lowerBound, upperBound] = useMemo(
-    () => [(currentPage - 1) * currentPageSize, (currentPage + 0) * currentPageSize],
-    [currentPage, currentPageSize],
-  );
+  const [searchString, setSearchString] = useState('');
+
+  const searchResults: ActiveVisitRow[] = useMemo(() => {
```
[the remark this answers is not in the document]
vasharma05: Sure!

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
vasharma05: So sorry

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
FlorianRappl: Alright - thanks for the clarification.

Alternatively, you could have put the `patientUuid` in a hidden column; but I think that way is also fine. Works for me!

FlorianRappl (approved):
LGTM - thanks!

PR #27, opened 2021-07-27
Fix: Store local patient lists information on a per-user basis

manuelroemer:
A leftover from the last PR: The local patient list metadata (e.g. starred lists) should be considered on a per-user basis on the current device. The last PR only stored the info globally.

This one adds a `userId` to the stored data.


FlorianRappl (approved):
LGTM - thanks! :beers:

PR #30, opened 2021-08-03
Full Integration of Local Patient Lists Into Current Feature Set | Patient List Project Refactoring & Cleanup

manuelroemer:
This is a bigger one - apologies for that! 🙏🏻  The changes went all hand-in-hand.

## Functional Changes

The main goal of this PR is the full integration of the local/offline patient list(s) into the current feature set, i.e. to enable the following (previously not working) things:

* Support actually having/storing **patients** in **local** patient lists
  * Display these patient list's **member count** in the overview
  * Allow **adding patients** from the patient chart, just like with online patient lists
  * Allow **displaying these patients** in the overview
* Support **searching** also for local patient lists
* **Starring** local patient lists already worked, but the **UX** was bad - this PR changes this, such that users **immediately see the new `starred` state** when clicking on the star.
* Simplify the way offline patients are stored locally

#### Selected Screenshots (Most things aren't screenshotable sadly) 

image
Local starred patient list with member count on it.

image
Support for adding patients to local patient list.

image
Displaying members of a local patient list is prepared.

## Refactoring

Some changes required refactoring. Here I changed the following:

* Added a simple `useAsync` hook for declaratively managing query/mutation lifecycles.  
  Replaces the copy&pasted `FetchState` which was present before in several locations.
* Restructured the API requests into `api-remote` (for remote/online patient list requests) and `api-local` (for local patient list queries). Both are consolidated in `api` which decides the right call to make (i.e. remote vs local).
* Added a much simpler global store for notifying offline patient list changes. This one is more both simpler and more generic and will be much easier to extend in the future.  
  Replaces the entire previous `offlineData` folder (together with patients now being tracked in the local patient list DB).
* Renamed files and folders in the patient-list repo so that they are consistent with the other projects (e.g. use the `kebab-case.component.tsx` naming scheme).


packages/esm-patient-list-app/src/api/api-local.ts
```diff
@@ -0,0 +1,144 @@
+import { isOfflineUuid, offlineUuidPrefix } from '@openmrs/esm-framework';
+import Dexie, { Table } from 'dexie';
+import { PatientList, PatientListMember, PatientListUpdate, PatientListType, PatientListFilter } from '.';
```
FlorianRappl, line 3: Imports from adjacent `index` are always problematic because they may be cyclic.

FlorianRappl: (In this case I think it may only be for types? Remember an `enum` in TS is not a compile-time only construct, but will have runtime impact, too)

packages/esm-patient-list-app/src/api/api.ts
```diff
@@ -0,0 +1,84 @@
+import { isOfflineUuid } from '@openmrs/esm-framework';
+import {
```
FlorianRappl, line 2: Same as the comment above; avoid imports from adjacent index modules.

FlorianRappl (commented):
You, sir, are a god. I have nothing to teach you.

packages/esm-patient-list-app/src/api/api.ts
manuelroemer: Done!

manuelroemer:
> I have nothing to teach you.

Well, I learnt a lot of stuff from you already - I'm sure there will be a lot more 🤗

FlorianRappl (approved):
LGTM! :beers:

PR #36, opened 2021-08-10
Move Global Store Creation Out of setupOpenMRS | Localization String Updates

manuelroemer:
After looking into how global store changes could be queued, I came to the realization that it would be much simpler to create the global store during module setup (this way it's also consistent with, e.g., the way breadcrumbs are handled). I'd much prefer this solution to a custom queueing strategy since it's just so much simpler to understand/implement. Let me know if that's acceptable. ;)

I also addressed some localization/naming issues mentioned [here](https://openmrs.slack.com/archives/C024GLJ0W2H/p1628176628001700).


brandones (approved):
Makes sense to me!

PR #37, opened 2021-08-11
Update patient list styling

herbertUG:
**What does this PR do?**
This PR adds styling to the Patient list tables to match the zeplin designs.

**ScreenShoots**
Screenshot 2021-08-11 at 6 40 46 AM
Screenshot 2021-08-11 at 6 41 00 AM

Design:

- https://app.zeplin.io/project/60d5947dd636aebbd63dce4c/screen/60ddb15b6c18f03cea663424
- https://app.zeplin.io/project/60d5947dd636aebbd63dce4c/screen/60dda7534fd8c43dda252de9

### DIFF
--- a/packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
+++ b/packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
@@ -3,7 +3,8 @@
 @import "~@openmrs/esm-styleguide/src/vars";
 
 .searchOverrides {
-    width: 250px;
+    width: 100%;
+    max-width: 250px;
     padding-top: $carbon--spacing-05;
     border: 0px !important;
 }
--- a/packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
+++ b/packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
@@ -1,3 +1,48 @@
 .searchOverrides {
-    width: 50% !important;
+    width: 250px !important;
+    padding-top: 17px !important;
+    border: 0px !important;
+}
+
+.searchOverrides > input {
+    height: 32px !important;
+}
+
+.searchOverrides > div {
+   top: 70%;
+}
+
+.searchOverrides > button {
+    height: 32px !important;
+    top: 10px;
+}
+
+div.table1  {
+    background-color: #f4f4f4;
+    margin: -16px;
+}
+
+div.table1 > div {
+    white-space:nowrap;
+    padding: 0px !important;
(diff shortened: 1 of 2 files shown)


packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
```diff
@@ -1,3 +1,48 @@
 .searchOverrides {
-    width: 50% !important;
-}
\ No newline at end of file
+    width: 250px ;
```
manuelroemer, line 2: I'd prefer not having the fixed `250px` here. Could this be solved otherwise (e.g. via an SCSS variable or a percentage)? 

**Disclaimer:** I don't have access to the designs on Zeplin right now, so I can only make a guess. If this exact width is actually a requirement of the design, let's keep this as-is.

manuelroemer (changes requested):
Hey, thanks a lot for your work! It generally looks good, but we should try to avoid having hardcoded values and custom CSS rules whenever possible. I made some comments at some lines, but the feedback generally applies to the entire SCSS file. 

You can make the mentioned variables available by importing them like this:

```scss
@import "~carbon-components/scss/globals/scss/vendor/@carbon/elements/scss/type/styles";
@import "~carbon-components/src/globals/scss/vendor/@carbon/layout/scss/generated/spacing";
@import "~@openmrs/esm-styleguide/src/vars";
```

At the end of the day it would be great if there are no hardcoded values anymore and if there's no `px` unit in the file (except for border widths).

herbertUG:
Thanks for the feedback @manuelroemer. I shall be implementing what is required.

packages/esm-patient-list-app/src/patient-table/patient-table.component.scss
herbertUG: Currently, the fixed `250px` is required to remain constant on different screen sizes. Since percentages keep adjusting the width based on the parent div. This handles the requirements well.

FlorianRappl: I'd say it should rather be `width: 100%; max-width: 250px` to avoid having an horizontal overflow in case of smaller canvas.

herbertUG: @FlorianRappl I have tested `width: 100%; max-width: 250px` and adjusted accordingly.

manuelroemer (approved):
LGTM, thanks a lot!

PR #38, opened 2021-08-11
MF-549: Patient Registration: Offline Patient List Integration

manuelroemer:
In patient registration, precaches patients when they are added to an offline list.


FlorianRappl (approved):
LGTM - thanks!

PR #41, opened 2021-08-24
Offline Tools Integration: Offline Patients Table

manuelroemer:
## Requirements

## Summary
**⚠ Important:** This PR is dependent on https://github.com/openmrs/openmrs-esm-core/pull/173 and depends on new functions in `esm-framework`. Let's merge it after the other one is merged.

It makes the offline patients table in offline-tools available via extensions. The reason for managing the table here is that the table should only display the user's on a user's offline list (which is inherently managed by the patient-list MF).

## Screenshots

See https://github.com/openmrs/openmrs-esm-core/pull/173.

## Related Issue
https://issues.openmrs.org/browse/MF-748
https://issues.openmrs.org/browse/MF-749

## Other

*None.*


packages/esm-patient-list-app/src/offline-patient-table/last-updated-table-cell.component.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { navigate, OfflinePatientDataSyncState } from '@openmrs/esm-framework';
+import { useTranslation } from 'react-i18next';
+import PendingFilled16 from '@carbon/icons-react/es/pending--filled/16';
+import WarningAltFilled16 from '@carbon/icons-react/es/warning--alt--filled/16';
+import CheckmarkOutline16 from '@carbon/icons-react/es/checkmark--outline/16';
+import styles from './last-updated-table-cell.scss';
+import { TableCell } from 'carbon-components-react/lib/components/DataTable';
+import Link from 'carbon-components-react/lib/components/Link';
```
FlorianRappl, line 9: Sure you want to use `/lib` and not `/es`? Usually the latter is more optimized for our use case.

packages/esm-patient-list-app/src/offline-patient-table/offline-patient-table.component.tsx
```diff
@@ -0,0 +1,226 @@
...
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableHeader,
+  TableRow,
+  FilterRowsData,
+  TableSelectAll,
+  TableSelectRow,
+} from 'carbon-components-react/lib/components/DataTable';
```
FlorianRappl, line 13: Same here; I think all of these `/lib` references may be `.../es/` ones. Maybe this is a general repo problem. Would be superb if you could check this!

packages/esm-patient-list-app/src/offline-patient-table/last-updated-table-cell.component.tsx
manuelroemer: The reason why I used the lib import is that this is the default import that VS Code suggests. But it does have one other advantage: In comparison to es, lib imports give you type checking and IntelliSense. That's not the case with the es imports. 

But sure, I can change that!

FlorianRappl: Hm I might be wrong, but lib is an ES5 build with commonjs as format and ES may be ES6 with the ES Module system. So the latter is "easy" to handle from the bundler while the former will result in some unnecessary bloat. If I am wrong here (please check and report) then we can / should stay at "lib".

manuelroemer: Sure, I’ll have a look! If es is the preferred one I’ll also have to update the code in yesterday’s esm-core PR.

FlorianRappl (approved):

PR #42, opened 2021-09-02
MF-749: Migrated patient-list and patient-registration to the new patient sync/SW headers API.

manuelroemer:
## Requirements

## Summary
Migrated the patient-list and patient-registration code to the newer patient synchronization APIs and SW header API for caching API requests.

## Screenshots

*None.*

## Related Issue
https://issues.openmrs.org/browse/MF-749

## Other

*None.*


packages/esm-patient-registration-app/src/constants.ts
```diff
@@ -1,2 +1,8 @@
+import { OmrsOfflineHttpHeaders } from '@openmrs/esm-framework';
+
 export const moduleName = '@openmrs/esm-patient-registration-app';
 export const patientRegistration = 'patient-registration';
+
+export const cacheForOfflineHeaders: OmrsOfflineHttpHeaders = {
+  'x-omrs-offline-caching-strategy': 'network-first',
```
FlorianRappl, line 7: One suggestion that I have is that this header name (along the others) might be exported as a constant from `@openmrs/esm-framework`. The benefit would be, that we could change the value of the constant later (making refactoring easier).

FlorianRappl (approved):
Great!

packages/esm-patient-registration-app/src/constants.ts
manuelroemer: Good point, that makes a lot of sense. I'll get on it right away!

PR #43, opened 2021-09-03
Add 'Actions' in patient search results and action to add a patient to a list

samuelmale:
Quick demo:

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-result/patient-search-result.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-result/patient-search-result.component.tsx
@@ -1,35 +1,43 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import dayjs from 'dayjs';
 import styles from './patient-search-result.scss';
-import { interpolateString, ExtensionSlot, useConfig, navigate } from '@openmrs/esm-framework';
+import { interpolateString, ExtensionSlot, useConfig, navigate, attach, detach } from '@openmrs/esm-framework';
 import { SearchedPatient } from '../types/index';
+import CustomOverflowMenuComponent from '../ui-components/overflow-menu.component';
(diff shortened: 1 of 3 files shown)


packages/esm-patient-search-app/src/patient-search-result/patient-search-result.component.tsx
```diff
@@ -1,35 +1,43 @@
...
 import { SearchedPatient } from '../types/index';
+import CustomOverflowMenuComponent from '../ui-components/overflow-menu.component';
+import OverflowMenuVertical16 from '@carbon/icons-react/es/overflow-menu--vertical/16';
+import { useTranslation } from 'react-i18next';
 
 const PatientSearchResults: React.FC<PatientSearchResultsProps> = ({ patients, hidePanel }) => {
   const config = useConfig();
+  const { t } = useTranslation();
+
+  useEffect(() => {
```
FlorianRappl, line 14: Why this effect? I think it that the extension should always be attached here.

FlorianRappl: (in general, why is this a new slot instead of taking the existing slot name? does it really matter if the action button is shown on the patient chart or on the search?)

packages/esm-patient-search-app/src/ui-components/overflow-menu.component.tsx
```diff
@@ -0,0 +1,78 @@
+import React, { useState, useEffect, useRef } from 'react';
```
FlorianRappl, line 1: Hm this was just copied from patient chart, right? Maybe we should make this reusable, e.g., by making it an extension or placing it in the components lib.

samuelmale:
@FlorianRappl closing in favour of https://github.com/openmrs/openmrs-esm-patient-management/pull/46

PR #45, opened 2021-09-21
MF-760: Changes in registration form for tablet view.

vasharma05:
## Requirements

- [ Yes] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [ Yes] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary

I have done the changes as per the design requirements given in the [design here](https://zpl.io/VDNg9BO)  . The changes are done for both tablet and desktop views. 

## Screenshots

https://user-images.githubusercontent.com/51502471/134116003-b4623f8c-d34d-460c-81b6-832f24708ec4.mp4

## Related Issue

None

## Other

*None.*


packages/esm-patient-registration-app/src/patient-registration/field/gender/gender-field.component.tsx
```diff
@@ -20,22 +20,24 @@ export const GenderBirthField: React.FC = () => {
...
-          <RadioButton id="female" labelText={t('femaleLabelText', 'Female')} value="Female" />
-          <RadioButton id="other" labelText={t('otherLabelText', 'Other')} value="Other" />
-        </RadioButtonGroup>
-        {meta.touched && meta.error && (
-          <>
-            <input type="hidden" data-invalid="true" />
-            <div className="bx--form-requirement">{t(meta.error)}</div>
-          </>
-        )}
+        <div className={styles.ageField}>
```
HerbertYiga, line 23: cc @FlorianRappl

brandones:
Nice, thanks @vasharma05 ! Great video also; that's super helpful to see the live comparison, especially when talking about responsiveness.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
```diff
@@ -31,37 +33,51 @@ export const NameField = () => {
...
+        <h4 className={styles.productiveHeading02Light}>{t('fullNameLabelText', 'Full Name')}</h4>
+        <div>
+          {!desktop && (
+            <div style={{ margin: '1rem 0' }}>
+              <ExtensionSlot
+                extensionSlotName="capture-patient-photo-slot"
+                state={{ onCapturePhoto, initialState: currentPhoto }}
+              />
+            </div>
+          )}
```
brandones, line 47: The most important thing we need to fix about this PR is that we shouldn't have these duplicated sections that are in different places based on the screen size. Instead, the component should exist only once in code, and be reordered using the [Flex Order property](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Ordering_Flex_Items).

brandones (changes requested):
Just that one change above (though it will probably be a little tricky to figure out). Otherwise I think this looks good

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: Hi @brandones! So sorry for replying here so late, actually if we see the designs, I don't have any other option to go for other than this. The designs for the laptop and the table view is as follows: 

Desktop: 
image

Tablet:
image

I'll have a look at it, if I can get some good news!

brandones: Not sure I understand. Why can't you use the Flex Order property to achieve the reordering of elements?

vasharma05: Hey @brandones , I know it's very late, but I have worked upon and now it's working good.

vasharma05:
https://user-images.githubusercontent.com/51502471/137124748-fcf18755-aa23-45d2-a8f8-df4495b0b971.mp4

brandones (approved):
Nice work!

PR #54, opened 2021-10-12
Refactor carbon component imports

denniskigen:
## Requirements

## Summary

This PR refactors carbon components imports in this repo so that they source components from the `carbon-components-react` package.

## Related Issue

Partially related to https://issues.openmrs.org/browse/MF-781

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
@@ -58,7 +58,6 @@ export default class FormManager {
       },
       {
         id: patientUuid,
-        // @ts-ignore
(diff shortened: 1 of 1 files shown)


packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -58,6 +58,7 @@ export default class FormManager {
       },
       {
         id: patientUuid,
+        // @ts-ignore
```
manuelroemer, line 61: This should probably not be a TS error. It is a fairly new API in esm-framework though. Could you check whether updating the packages solves the issue? (`rm -rf node_modules && rm yarn.lock && npx lerna bootstrap`)

PR #63, opened 2021-11-09
O3-823: Editing patient details continually adds the patient address(es)

NischithJShetty:
## Requirements

- [x] I have checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary

Quoting from the [O3-823](https://issues.openmrs.org/browse/O3-823) page:
"Every time a patient is edited using the *Edit patient details* action, a new address is appended to the patient's list of addresses in the backend."

This PR fixes the stated issue.

## Screenshots

*None.*

## Related Issue

[O3-823](https://issues.openmrs.org/browse/O3-823)

## Other

*None.*

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
@@ -117,6 +117,38 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
   }, [patient, config]);
 
   const onFormSubmit = async (values: FormValues, helpers: FormikHelpers<FormValues>) => {
+    const checkIfEdited = (obj1, obj2) => {
+      const objectKeys = Object.keys(obj1);
+
+      for (const objKey of objectKeys) {
+        const val1 = obj1[objKey];
+        const val2 = obj2[objKey];
+        const isAnObject = val1 != null && typeof val1 === 'object';
+        if (
+          (isAnObject && Object.keys(val1).length !== Object.keys(val2).length) ||
+          (isAnObject && !checkIfEdited(val1, val2)) ||
+          (!isAnObject && val1 !== val2)
+        ) {
+          return false;
+        }
+      }
+      return true;
+    };
+
+    const obj1 = JSON.parse(JSON.stringify(initialFormValues));
+    const obj2 = JSON.parse(JSON.stringify(values));
+    const formNotEdited = checkIfEdited(obj1, obj2);
+
+    if (formNotEdited && inEditMode) {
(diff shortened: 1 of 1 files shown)


brandones:
@NischithJShetty , could you explain how your fix fixes the problem where duplicate addresses get created? It looks like all you're doing is conditionally changing the redirect URL.

NischithJShetty:
@brandones 
As soon as the `Edit patient details` form is submitted the `onFormSubmit` function gets called and I have just added a few lines to the very begining of this function. 

I have created a new function called `checkIfEdited` which takes in two parameters:

- `initialFormValues` : An object containing the values all the fields of the form prior to any edits
- `values` : An object containing the values of all the fields after edits (if any)

The `checkIfEdited` function meticulously compares the above two objects (parsed into JSONs) to varify if they are equal or not. If the two objects are equal, it would mean that no edits have been made. Only if any edits have been made the `onFormSubmit` will continue to execute as before and otherwise the function will exit preventing any unnecessary call to the Backend (which actually is the root cause of this issue).

Thanks.

brandones:
Sorry, I didn't need you to explain the code to me; I can read code. What I didn't understand is why your code solves the problem. All I was looking for was

> the function will exit preventing any unnecessary call to the Backend (which actually is the root cause of this issue).
>
> Earlier even if a user just opened up the form to make edits but submitted it without making any edits, a needless Backend request was being sent which was creating the duplicate addresses.

Which does answer my question nicely.

Does the problem still occur if the registration form is edited? Say, if I edit the patient's age, will it still create a duplicate address?

NischithJShetty:
Yes, in that scenario it does create duplicate addresses. My assumption is that the logic for that needs to be handled by the Backend.

brandones:
Okay, I'm going to close this, since it doesn't really address the problem. Thanks @NischithJShetty .

NischithJShetty:
@brandones 
At the [ticket page](https://issues.openmrs.org/browse/O3-823), I was pointed to this repo to fix the problem and only so much can be done here. Could you please point me to the correct repo to be worked with to fix this issue?

brandones:
The problem is most likely in this repo. I'd recommend identifying the exact backend call that creates duplicate addresses and changing it so that it doesn't create duplicate addresses.

NischithJShetty:
Thanks! In that case I will submit another PR soon which will take care of adreess duplication when there are form edits as well (this PR was only taking care of the specific situation of duplication when no edits are made as specified in the Ticket)

NischithJShetty:
A follow-up PR has been submitted at #65 which fills the gaps of this PR and covers all the use cases of address duplication. Thanks.

PR #65, opened 2021-11-17
O3-823: Editing patient details continually adds the patient address(es)

NischithJShetty:
## Requirements

- [x] I have checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary

Quoting from the [O3-823](https://issues.openmrs.org/browse/O3-823) page:
"Every time a patient is edited using the *Edit patient details* action, a new address is appended to the patient's list of addresses in the backend."

This PR fixes the stated issue.

This is a follow-up PR to #63 

## Screenshots

*None.*

## Related Issue

[O3-823](https://issues.openmrs.org/browse/O3-823)

## Other

*None.*

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
@@ -188,6 +188,25 @@ export default class FormManager {
     return [];
   }
 
+  static checkIfEdited = (obj1, obj2) => {
+    const objectKeys = Object.keys(obj1);
+
+    for (const objKey of objectKeys) {
+      const val1 = obj1[objKey];
+      const val2 = obj2[objKey];
+      const isAnObject = val1 != null && typeof val1 === 'object';
+      if (
+        (isAnObject && Object.keys(val1).length !== Object.keys(val2).length) ||
+        (isAnObject && !FormManager.checkIfEdited(val1, val2)) ||
+        (!isAnObject && val1 !== val2)
+      ) {
+        return false;
+      }
+    }
+
+    return true;
+  };
+
   static getPatientToCreate(
     values: FormValues,
     personAttributeSections: any,
@@ -195,6 +214,12 @@ export default class FormManager {
     initialAddressFieldValues: Record<string, any>,
     identifiers: Array<PatientIdentifier>,
   ): Patient {
+    let address = FormManager.getPatientAddressField(values, initialAddressFieldValues);
+
+    if (FormManager.checkIfEdited(initialAddressFieldValues, address)) {
+      address = {};
+    }
+
     return {
       uuid: patientUuidMap['patientUuid'],
       person: {
@@ -204,7 +229,7 @@ export default class FormManager {
         birthdate: values.birthdate,
         birthdateEstimated: values.birthdateEstimated,
         attributes: FormManager.getPatientAttributes(values, personAttributeSections),
-        addresses: [FormManager.getPatientAddressField(values, initialAddressFieldValues)],
+        addresses: [address],
         ...FormManager.getPatientDeathInfo(values),
       },
       identifiers,
--- a/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
@@ -17,6 +17,7 @@ import {
   savePatientPhoto,
   saveRelationship,
 } from './patient-registration.resource';
+import isEqual from 'lodash-es/isEqual';
 
 export type SavePatientForm = (
   patientUuid: string | undefined,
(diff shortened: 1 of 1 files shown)


NischithJShetty:
@brandones 
This is a follow-up PR to #63 and takes care of the deficiencies the earlier PR had. The logic used is as follows:
A comparison is made between the earlier address and the address contained in the form used for editing patient details, if they are similar, an empty address array is sent during the backend call preventing address duplication. Thanks.

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -188,13 +188,38 @@ export default class FormManager {
...
     personAttributeSections: any,
     patientUuidMap: PatientUuidMapType,
     initialAddressFieldValues: Record<string, any>,
     identifiers: Array<PatientIdentifier>,
   ): Patient {
+    let address = FormManager.getPatientAddressField(values, initialAddressFieldValues);
+
+    if (FormManager.checkIfEdited(initialAddressFieldValues, address)) {
+      address = {};
+    }
```
brandones, line 221: Did you try

```suggestion
    if (isEqual(initialAddressFieldValues, address)) {
      address = {};
    }
```

using [lodash `isEqual`](https://lodash.com/docs/4.17.15#isEqual)?

brandones:
If the user changes the patient's address, does it invalidate or remove the old address? Or does it just create another address for the patient, so that the patient has multiple addresses?

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
NischithJShetty: Thanks! I will replace the custom function with the 'lodash' function and make another commit.

NischithJShetty:
If the user changes the patient's address, the old address is not invalidated or removed. The earlier address is preserved as an 'old' address for future reference.

NischithJShetty:
@brandones 
A second commit has been made which uses the 'loadash' function instead of the earlier one. Thanks.

brandones:
@jonathandick @ibacher , are we trying to keep the old addresses around?

@NischithJShetty could you share a screenshot of what it looks like with old addresses?

NischithJShetty:
When I use an API client to send a request of the type `/ws/rest/v1/patient/<patient uuid>?v=custom:(addresses) or /ws/fhir2/R4/Patient/<patient uuid>?_elements=address` the resposponse will be like:

`{
    "resourceType": "Patient",
    "id": "a281b92d-2b64-4a24-9c70-523b1ba94dbb",
    "meta": {
        "tag": [
            {
                "system": "http://terminology.hl7.org/CodeSystem/v3-ObservationValue",
                "code": "SUBSETTED",
                "display": "Resource encoded in summary mode"
            }
        ]
    },
    "address": [
        {
            "id": "8f4d4a1b-9b54-4338-8b7c-8431fde27600",
            "extension": [
                {
                    "url": "http://fhir.openmrs.org/ext/address",
                    "extension": [
                        {
                            "url": "http://fhir.openmrs.org/ext/address#address1",
                            "valueString": "aa"
                        }
                    ]
                }
            ],
            "use": "home",
            "city": "cccc",
            "state": "rrrrr3",
            "country": "countryyy"
        },
        {
            "id": "3171e4c3-a3ba-4b4f-8aa3-1b725fbadb35",
            "extension": [
                {
                    "url": "http://fhir.openmrs.org/ext/address",
                    "extension": [
                        {
                            "url": "http://fhir.openmrs.org/ext/address#address1",
                            "valueString": "aa"
                        }
                    ]
                }
            ],
            "use": "old",
            "city": "aa",
            "state": "aa",
            "country": "tr"
        }
    ]
}`

NischithJShetty:
When I use a API client to make a request of the type `/ws/rest/v1/patient/<patient uuid>?v=custom:(addresses) or /ws/fhir2/R4/Patient/<patient uuid>?_elements=address` the response I get is:

image

brandones:
Sorry @NischithJShetty , I meant a screenshot of the UI. Does the UI show multiple addresses? Could you share what that looks like?

ibacher:
@brandones Well, the backend supports the notion of multiple addresses per patient and I think in an ideal world we'd have a registration mechanism that supported maintaining more than one address for cases where that made sense. However, there needs to be a way to remove a previous address.

NischithJShetty:
@brandones 
The UI doesn't show multiple addresses, only the newest address will be displayed (even in the form used to edit the patient details). The old address will only be stored in the Database and one could have access to it, if and when needed (this is as far as I understand).

brandones (approved):
Awesome, I think we're in good shape then! Thanks @NischithJShetty .

PR #68, opened 2021-11-23
O3-940: Redesign patient avatars

denniskigen:
## Requirements

- [x]  My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).

## Summary

Redesign patient avatars to match the designs outlined in https://app.zeplin.io/project/61434fa756474d5545f65cf4/screen/614362b21dd66d20a276d8b8.

Summary of the work done:

- Install [react-avatar](https://github.com/ambassify/react-avatar).
- Install [geopattern](https://github.com/btmills/geopattern).
- Add a `usePatientPhoto` hook that leverages SWR to fetch the patient photo (nice, but not strictly required).
- Refactor DisplayPhotoComponent to use the `Avatar` component from react-avatar as follows: 
   - Pass in the patient name as the value of the `name` prop. In the absence of an image `src`, `react-avatar` derives the patient's initials and renders them inside a placeholder SVG as a fallback.
   - Set the `color` prop value to a transparent colour so that the background image pattern provided by geopattern appears in the avatar.
   - Finally, use an inline style attribute to set the generated pattern as the background image of the avatar.

Related PR: https://github.com/openmrs/openmrs-esm-patient-chart/pull/464.

## Screenshots

Screenshot 2021-11-23 at 13 24 56

Screenshot 2021-11-23 at 13 25 12

Screenshot 2021-11-23 at 13 27 33

## Related Issue

https://issues.openmrs.org/projects/MF/issues/O3-940


packages/esm-patient-registration-app/src/widgets/display-photo.component.tsx
```diff
@@ -1,26 +1,28 @@
...
+import Avatar from 'react-avatar';
+import GeoPattern from 'geopattern';
+import { usePatientPhoto } from '../patient-registration/patient-registration.resource';
 import styles from './display-photo.scss';
 
-export default function DisplayPatientPhoto(props: { patientUuid: string }) {
-  const [photo, setPhoto] = useState(placeholder);
-  const config = useConfig();
+interface DisplayPatientPhotoProps {
+  patientName: string;
```
brandones, line 8: Given that we're already fetching the patient photo data using the `patientUuid`, can't we just get the name from the same SWR-cached fetch data, and avoid adding it to the extension props?

denniskigen: It's entirely possible - the patient's person object includes the `display` property which contains the primary identifier concatenated with the patient name. Remember that we're rendering the patient's name as a fallback. This strategy is contingent on having the patient name available at the time the Avatar component gets rendered. Relying on a network request would mean having to wait for the request to resolve and needing to show some loading state in the meantime. I don't know if that'd be an ideal experience.

brandones: But don't we wait for the request to resolve anyway, to find out whether there's a patient picture?

denniskigen: The real problem is the endpoint we're using to fetch the patient photo doesn't return anything if there's no observation data for that particular patient. It only resolves the patient's name when there's a picture stored on the server.

brandones: Okay, thanks for checking @denniskigen

brandones (approved):

PR #69, opened 2021-11-25
03 938 : Should be able to edit patient details from the patient search page

donaldkibet:
## Requirements

## Summary

At the moment, when you try to navigate to patient-registration in edit mode, from the patient-search, patient search result panel doesn't close. This PR adds that ability.

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-registration-app/src/widgets/edit-patient-details-button.component.tsx
+++ b/packages/esm-patient-registration-app/src/widgets/edit-patient-details-button.component.tsx
@@ -1,20 +1,40 @@
 import React from 'react';
-import { ConfigurableLink } from '@openmrs/esm-framework';
+import { navigate } from '@openmrs/esm-framework';
 import { useTranslation } from 'react-i18next';
 import styles from './edit-patient-details-button.scss';
 
-export default function EditPatientDetailsButton({ patientUuid }) {
+interface EditPatientDetailsButtonProps {
+  closePatientSearchResultsPanel?: () => void;
+  patientUuid: string;
+}
+
+const EditPatientDetailsButton: React.FC<EditPatientDetailsButtonProps> = ({
+  patientUuid,
+  closePatientSearchResultsPanel,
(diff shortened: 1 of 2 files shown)


packages/esm-patient-registration-app/src/widgets/edit-patient-details-button.component.tsx
```diff
@@ -1,20 +1,40 @@
...
 
-export default function EditPatientDetailsButton({ patientUuid }) {
+interface EditPatientDetailsButtonProps {
+  closePatientSearchResultsPanel?: () => void;
+  patientUuid: string;
+}
+
+const EditPatientDetailsButton: React.FC<EditPatientDetailsButtonProps> = ({
+  patientUuid,
+  closePatientSearchResultsPanel,
```
brandones, line 13: This button seems to know too much about its context. I think we should establish a pattern to deal with this kind of situation, where we have a function prop that should be called on click/transition. Maybe we should call this prop `onTransition`? Or `afterClick`?

packages/esm-patient-registration-app/src/widgets/edit-patient-details-button.component.tsx
```diff
@@ -1,20 +1,40 @@
...
-        title={t('editPatientDetails', 'Edit Patient Details')}>
+      <button
+        className="bx--overflow-menu-options__btn"
+        role="menuitem"
+        title={t('editPatientDetails', 'Edit Patient Details')}
+        data-floating-menu-primary-focus
+        onClick={handleClick}
+        style={{
+          maxWidth: '100vw',
+        }}>
```
brandones, line 31: This seems like it shouldn't ever do anything?

brandones (changes requested):
Let's just change the prop name at least, and call it good. All this will get refactored anyway, once we have the Link system.

brandones (approved):

PR #77, opened 2022-01-13
O3-1002: Registration: Support for Unknown gender

jnsereko:
## Requirements

## Summary

adds support for "Unknown" gender

## Screenshots
image

## Related Issue

https://issues.openmrs.org/browse/O3-1002

## Other

*None.*

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/field/gender/gender-field.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/field/gender/gender-field.component.tsx
@@ -28,8 +28,8 @@ export const GenderBirthField: React.FC = () => {
           <RadioButtonGroup name="gender" orientation="vertical" onChange={setGender} valueSelected={field.value}>
             <RadioButton id="male" labelText={t('maleLabelText', 'Male')} value="Male" />
             <RadioButton id="female" labelText={t('femaleLabelText', 'Female')} value="Female" />
-            <RadioButton id="unknown" labelText={t('unknownLabelText', 'Unknown')} value="Unknown" />
             <RadioButton id="other" labelText={t('otherLabelText', 'Other')} value="Other" />
+            <RadioButton id="unknown" labelText={t('unknownLabelText', 'Unknown')} value="Unknown" />
           </RadioButtonGroup>
           {meta.touched && meta.error && (
             <>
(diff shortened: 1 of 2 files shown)


packages/esm-patient-registration-app/src/patient-registration/field/gender/gender-field.component.tsx
```diff
@@ -28,6 +28,7 @@ export const GenderBirthField: React.FC = () => {
           <RadioButtonGroup name="gender" orientation="vertical" onChange={setGender} valueSelected={field.value}>
             <RadioButton id="male" labelText={t('maleLabelText', 'Male')} value="Male" />
             <RadioButton id="female" labelText={t('femaleLabelText', 'Female')} value="Female" />
+            <RadioButton id="unknown" labelText={t('unknownLabelText', 'Unknown')} value="Unknown" />
```
jnsereko, line 31: @ibacher 
Is this sufficient, or i need to think of passing this to the server

manuelroemer: This should be fine from the frontend side (the code as-is will send the value `"gender": "U" to the backend).
If you want to end-to-end-test that this works, we'll have to get patient registration working again. Though you could probably also just look into the network tab to look at the request and see if the gender field is set correctly.

manuelroemer (approved):
I think that this is fine and will work as-is, but I'm not merging yet in case @ibacher has some additional input here. He definitely knows the API better than me. 👀

packages/esm-patient-registration-app/src/patient-registration/field/gender/gender-field.component.tsx
ibacher: I'd put "unknown" after "other", but aside from that, this looks good to me.

ibacher: Ok... [we use](https://github.com/openmrs/openmrs-esm-patient-management/blob/68c760b1033a38162a6469c8dcaa35bd0412dfdb/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts#L225) `value.gender.charAt(0)`, i.e., the first character as the payload to the backend, so that means we're sending the correct code....

PR #88, opened 2022-01-31
03-1034 : (enhancement) Add support for branding on active visits

donaldkibet:
## Requirements

## Summary

- Add branding for active-visit header 

## Screenshots
brading

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.scss
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.scss
@@ -94,7 +94,8 @@
     display: block;
     width: 2rem;
     padding-top: 3px;
-    border-bottom: 0.375rem solid $brand-teal-01;
+    border-bottom: 0.375rem solid;
+    @include brand-03(border-bottom-color);
   }
 }
 


packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -175,10 +175,11 @@ const ActiveVisitsTable = () => {
                           <th colSpan={headers.length + 2}>
                             <ExtensionSlot
                               className={styles.visitSummaryContainer}
-                              extensionSlotName="visit-summary-slot"
+                              extensionSlotName="visit-detail-overview-slot"
```
brandones, line 178: Yeah this seems like a worse name

brandones:
Screenshot from 2022-02-01 09-32-39
love to see it

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -175,10 +175,11 @@ const ActiveVisitsTable = () => {
...
                               state={{
-                                visitUuid: paginatedActiveVisits[index]?.visitUuid,
+                                visit: paginatedActiveVisits[index]?.visit,
                                 patientUuid: paginatedActiveVisits[index]?.patientUuid,
+                                listViewOverride: true,
                               }}
                             />
```
brandones, line 184: Is this a single-use extension slot (i.e., only one extension ever goes in it, and it should always go in it)? Can we replace it with just using the component directly?

jonathandick: I realize now, we made this into an extension slot as this gives orgs the flexibility to use their own widget to show this data. At AMPATH, we're using obs to recording medication orders as well as concepts that may be specific to us.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
jonathandick: I'm not exactly sure what we're putting in this slot so hard for me to recommend a different name.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
brandones: Where does that stuff show up? Don't we expect that kind of customization to be needed for most implementations? Can we accomplish that by making the widget configurable?

jonathandick: I feel like this would be comlicated to do via configuration as this is based on context. The visit type, program encrollment, the role of the user may all be implementation level factors that really require logic to sort out. I'd like to merge this for now as I believe it's blocking us from moving forward. If we change the approach later it won't be a big deal to refactor this

jonathandick:
@donaldkibet can you fix conflicts and we can then merge.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
jonathandick: let's just call this the av-visit-overview-slot for now. @brandones do we need to start a convention to name-space slots? Calling this the "visit-overview-slot" could eventually be used somewhere else which would be problematic. calling this the active-visits-widget-visit-overview-slow seems a tad long.

brandones: Oh that's an interesting question. Right now slot names do need to be globally unique. I would rather we lift that restriction and make them only have to be unique per module rather than establish a namespacing convention.

For this one, I would advocate calling it `visit-row-details-slot`. [A slot is a place](http://o3-dev.docs.openmrs.org/#/main/extensions?id=naming-slots).

jonathandick (approved):

PR #89, opened 2022-01-31
Fix visit start time

denniskigen:
## Requirements

## Summary
The active visits widget currently uses the visit `startDate` to compute the time when a visit started. This property doesn't exist on the visit object. We should instead use the `startDatetime` property. 

This commit fixes the ActiveVisits component logic to use the `startDatetime`.

## Screenshots

> Before 

Screenshot 2022-01-31 at 21 43 46

> After

Screenshot 2022-01-31 at 21 43 14

> Error when the visit is `null`

Screenshot 2022-01-31 at 21 42 57

> Fixed 

Screenshot 2022-01-31 at 21 44 19

### DIFF
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
@@ -19,12 +19,15 @@ export function useActiveVisits() {
   const { data: currentUserSession } = useCurrentSession();
   const startDate = dayjs().format('YYYY-MM-DD');
   const sessionLocation = currentUserSession?.data?.sessionLocation?.uuid;
+
   const customRepresentation =
     'custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),' +
     'visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,' +
-    'stopDatetime)&fromStartDate=';
-  const url =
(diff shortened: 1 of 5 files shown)


denniskigen:
Two interesting outcomes following this change:

1. I've realized that the active visits widget can show visits with start times in the future.
2. The active visits widget can show visits started for voided patients (like the test patient in the screenshot above).

packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
```diff
@@ -17,21 +17,15 @@ export interface ActiveVisit {
...
-    '&location=' +
-    sessionLocation;
-
-  const { data, error, isValidating } = useSWR<{ data: { results: Array<Visit> } }, Error>(
-    `/ws/rest/v1/visit?includeInactive=false&v=${customRepresentation}`,
-    openmrsFetch,
-  );
+    'stopDatetime)&fromStartDate=';
+  const url =
+    `/ws/rest/v1/visit?includeInactive=false&v=${customRepresentation}` + startDate + '&location=' + sessionLocation;
```
brandones, line 27: I find the previous version more readable. It seems strange to say `fromStartDate=` and then leave it hanging.

brandones (approved):
One readability concern, but generally LGTM.

denniskigen:
Thanks for the review, @brandones!

PR #95, opened 2022-02-04
O3-1050: Patient Search should display a Clear button instead of X inside text field

jnsereko:
## Requirements

## Summary
Instead of having a second X inside of the search bar to clear the text, the design has a ghost button saying Clear to separate it from the X above it to close the search bar entirely. See the design here: https://zpl.io/2jq39Xx

## Screenshots

*None.*

## Related Issue
https://issues.openmrs.org/browse/O3-1050

## Other

*None.*

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-icon/patient-search-icon.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-icon/patient-search-icon.component.tsx
@@ -125,6 +125,7 @@ const PatientSearchLaunch: React.FC<PatientSearchLaunchProps> = () => {
               className={styles.patientSearchInput}
               placeholder={t('searchForPatient', 'Search for a patient by name or identifier number')}
               labelText=""
+              closeButtonLabelText={t('clearSearch', 'Clear')}
               onKeyUp={handleEnterKeyPressed}
               onChange={(event) => handleChange(event.target.value)}
               autoFocus={true}
--- a/packages/esm-patient-search-app/translations/en.json
+++ b/packages/esm-patient-search-app/translations/en.json
@@ -1,4 +1,5 @@
 {
+  "clearSearch": "Clear",
   "error": "Error",
   "errorCopy": "Sorry, there was a an error. You can try to reload this page, or contact the site administrator and quote the error code above.",
   "errorText": "An error occurred while performing search",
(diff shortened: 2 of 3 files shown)


jnsereko:
@ibacher 
i am still learning how to debug this, so for a moment i might not be in the best position to tell the cause of the error.

At the same time adding the property `closeButtonLabelText={t('clearSearch', 'Clear')}`  to the `<Search .../>` is not working but i can see the lable with the developer tools as shown in the screenshot below.
Is this a problem with carbon or am doing it wrongly?
o 9 openmrs

ibacher:
@jnsereko Yes, I can see why this wasn't initially implemented... the Carbon components just don't support this use case (notice how the text appears in the `aria-label` element, so effectively as a tooltip.

packages/esm-patient-search-app/src/patient-search-icon/patient-search-icon.component.scss
```diff
@@ -55,3 +55,11 @@ $inverse-02:#393939;
...
   background-color: $inverse-02;
 }
+
+.bx--search .bx--search-close::before{
+  content: "Clear";
+}
+
+.bx--search .bx--search-close svg{
+  display: none;
+}
```
jnsereko, line 65: @ibacher atleast plain css had to do the trick but it has also failed.  Do you have any other options i should use?

jnsereko: Thank you, @ibacher 
I think `Internal` are accessible through released carbon packages 

I am thinking of a way we can do some sort of extending the Search-Component and override the returned/rendered fields  
Is this possible?

brandones: I don't think it would be worth it to fork the Search component just for this.

My suggestion: create an element containing the text and use CSS to hack it into place. It will be hacky but I think it is the best way forward. `display` and probably `position` (or maybe `margin`) will be the key CSS properties to mess with here.

ibacher: I'm actually thinking we maybe just leave the X and do what @jnsereko already did to have a label saying "Clear"

PR #96, opened 2022-02-04
chore:CI: Bump patch version for pre-releases

ibacher:
## Requirements

## Summary

This is the *best* strategy, but currently we're releasing what claim to be pre-releases of the current release. So let's adopt the "finalise version on release" strategy.

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


brandones (approved):
This brings it into agreement with the way esm-core is versioned, which I think is correct. Thank you @ibacher !

PR #185, opened 2022-04-25
O3-1075:Update the patient search feature to leverage the FHIR API

jwnasambu:
## Requirements

## Summary

I update the patient search feature to leverage the FHIR API instead of the REST API.

## Screenshots

https://user-images.githubusercontent.com/33891016/166296331-5e0b4c7b-d26f-404e-91f9-b65ec92733e7.mp4

## Related Issue
[O3-1075](https://issues.openmrs.org/browse/O3-1075)

## Other

*None.*


packages/esm-patient-search-app/src/patient-search/patient-search.resource.tsx
```diff
@@ -1,7 +1,7 @@
 import { openmrsFetch } from '@openmrs/esm-framework';
 
 export function performPatientSearch(query: string, objectVersion: string, ac: AbortController, includeDead: boolean) {
-  const url = `/ws/rest/v1/patient?q=${query}&v=${objectVersion}&includeDead=${includeDead}`;
+  const url = `/ws/fhir2/R4//Patient?name:contains=<q>&identifier=<q>${objectVersion}&includeDead=${includeDead}`;
```
vasharma05, line 4: @jwamalwa, only changing the URL works?
The FHIR response is different than the REST one. Have you tried to run the change?

vasharma05 (changes requested):

ibacher:
@vasharma05 said, the FHIR response is completely different from the REST response and likely we'll need several other changes here.

packages/esm-patient-search-app/src/patient-search/patient-search.resource.tsx
```diff
@@ -1,7 +1,7 @@
 import { openmrsFetch } from '@openmrs/esm-framework';
 
 export function performPatientSearch(query: string, objectVersion: string, ac: AbortController, includeDead: boolean) {
-  const url = `/ws/rest/v1/patient?q=${query}&v=${objectVersion}&includeDead=${includeDead}`;
+  const url = `/ws/fhir2/R4/Patient/_search?_name=${query}=${objectVersion}&includeDead=${includeDead}`;
```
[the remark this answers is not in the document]
jwnasambu: @ibacher I have fixed it after learning that  the `_search end` point is  primarily to allow search via POST request and shouldn’t be used for GET requests.

packages/esm-patient-search-app/src/patient-search/patient-search.resource.tsx
```diff
@@ -1,7 +1,7 @@
 import { openmrsFetch } from '@openmrs/esm-framework';
 
 export function performPatientSearch(query: string, objectVersion: string, ac: AbortController, includeDead: boolean) {
-  const url = `/ws/rest/v1/patient?q=${query}&v=${objectVersion}&includeDead=${includeDead}`;
+  const url = `/ws/fhir2/R4//Patient?name:contains=<q>&identifier=<q>`;
```
ibacher, line 4: So at a minimum we need this:

```suggestion
  const url = `/ws/fhir2/R4//Patient?name:contains=${query}&identifier=${query}`;
```

We also need to handle the `includeDead` parameter, which is a little weird. `includeDead` means include both deceased and living people, so we should have something like:

```ts
`${includeDead ? '&deceased=true,false' : ''}
```

Also, requests to the FHIR API should usually have the parameter `&_summary=data` at the end to avoid having to generate the FHIR narrative.

Maybe to handle that. The biggest part of this ticket, however, it to update the places that call `performPatientSearch` to ensure they continue to work when we switch to using the FHIR API, since FHIR API data is in a different format from data from the REST API.

packages/esm-patient-search-app/src/hooks/usePatients.tsx
```diff
@@ -0,0 +1,37 @@
...
+  isLoading: boolean;
+  error: any;
+}
+/**
+ * React hook that performs patient search using FHIR endpoint
+ * @param query The search term which can be identifier or name
+ * @param includeDead Whether to include dead patients in search results
+ * @returns Object {patients,isLoading,error}
+ */
+const usePatients = (query: string, includeDead: boolean = false): PatientReturnType => {
```
ZacButko, line 25: Please change this to `usePatientSearch`

ZacButko (changes requested):
The code seems to work. However I'm only seeing 10 patients returned for searching "Test" vs 18 returned when searching with the Rest endpoint? Which patients are missing?

packages/esm-patient-search-app/src/hooks/usePatients.tsx
jwnasambu: @ZacButko have fixed the proposed changes. Thanks for the review.

PR #190, opened 2022-04-28
O3-1268:Default patient identifier types should not be required

jwnasambu:
## Requirements

## Summary

When new default identifiers are configured to be shown on the registration screen, whether or not they are required should still be driven by the backend. So, for example, if I configure things so that an optional identifier appears by default on the registration screen, it should still not be required to submit the registration.

## Screenshots

Uploading screencast 2022-04-28 4 PM-14-36.mp4…

## Related Issue

*None.*

## Other

*None.*


jwnasambu:
@vasharma05 I tried it out before but didn't work as expected but I wouldn't mind to make the proposed changes.

jwnasambu:
@jwamalwa  When I try to change the required property in the identifier field on [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/input/basic-input/input/input.component.tsx#L43) line by replacing it with `invalid={false}` it not only fix the issue by it also affects other  required fields.

PR #196, opened 2022-05-05
O3-1179: Performance improvement in patient registration page loading

vasharma05:
## Requirements

## Summary
This PR improves the loading time of the patient registration page. 
Before: Resources such as current session, relationship types, identifier types with sources, and address template were required beforehand for the page to load. Since these network requests were made synchronously, hence the page takes time to load.

Now: These requests are now handled using SWR, preserving the prev concept of fetching the resources. The page has a few sections which are loading till the data is not loaded, but the whole page's loading is not stuck due to the network requests. The components are rendered without requiring the said resources.

## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/O3-1179

## Other

*None.*


packages/esm-patient-registration-app/src/index.ts
```diff
@@ -58,7 +58,6 @@ function setupOpenMRS() {
           savePatientForm: FormManager.savePatientFormOffline,
           isOffline: true,
         },
```
manuelroemer, line 60: I cannot comment at that location, but there is this piece of code above which is now obsolete and can be removed:
grafik

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
```diff
@@ -23,42 +24,48 @@ export const AddressField: React.FC = () => {
...
-      const name = property.getAttribute('name');
-      /*
+    if (addressTemplateXml) {
+      const templateXmlDoc = parseString(addressTemplateXml);
+      const nameMappings = getTagAsDocument('nameMappings', templateXmlDoc);
+      const elementDefaults = getTagAsDocument('elementdefaults', templateXmlDoc);
+      const properties = nameMappings.getElementsByTagName('property');
+      const propertiesObj = Array.prototype.map.call(properties, (property: Element) => {
+        const name = property.getAttribute('name');
+        /*
```
manuelroemer, line 37: I know that this is not your code, but I'm wondering whether the following comment means anything to anyone. Maybe we can just straight up remove it? 

grafik

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -24,6 +24,7 @@ import { getSection } from './section/section-helper';
 import { cancelRegistration, parseAddressTemplateXml, scrollIntoView } from './patient-registration-utils';
 import { useInitialAddressFieldValues, useInitialFormValues, usePatientUuidMap } from './patient-registration-hooks';
 import { ResourcesContext } from '../offline.resources';
+import { add } from 'lodash-es';
```
manuelroemer, line 27: I think that this is unused and can be removed (according to a simple CTRL+F) - otherwise I'd import it via `import add from 'lodash-es/add';`.

packages/esm-patient-registration-app/src/offline.resources.ts
```diff
@@ -7,16 +7,16 @@ import { PatientIdentifierType, FetchedPatientIdentifierType } from './patient-r
 import { cacheForOfflineHeaders } from './constants';
 
 export interface Resources {
-  addressTemplate: any;
-  currentSession: Session;
+  addressTemplate: undefined | any;
+  currentSession: Session | undefined;
   relationshipTypes: any;
-  identifierTypes: Array<PatientIdentifierType>;
+  identifierTypes: Array<PatientIdentifierType> | undefined;
```
manuelroemer, line 13: Nitpicky, I know, but I'd prefer to use a `?` instead of the `| undefined` here. For the `any` types it's not really required anyway (though it also doesn't hurt to add it, the `any` could become more concrete eventually and then having the `undefined` type in place already is a good thing).

packages/esm-patient-registration-app/src/root.component.tsx
```diff
@@ -1,33 +1,41 @@
...
+  const { data: addressTemplate } = useSWR('patientRegistrationAddressTemplate', fetchAddressTemplate);
+  const { data: relationshipTypes } = useSWR('patientRegistrationRelationshipTypes', fetchAllRelationshipTypes);
+  const { data: identifierTypes } = useSWR(
+    'patientRegistrationPatientIdentifiers',
+    fetchPatientIdentifierTypesWithSources,
+  );
 
   return (
-    <ResourcesContext.Provider value={resources}>
+    <ResourcesContext.Provider
```
manuelroemer, line 32: From my side, this doesn't need to happen in this PR, I just want to mention the following:
Since we do migrate to SWR, a good follow-up refactoring/issue to this PR would be to remove this `ResourcesContext` in favor of direct SWR calls in the respective components which require the data that's currently passed down via the context. That would free this `Root` component from doing any data fetching.

But again, from my side, this absolutely doesn't have to happen in this PR as the focus here really can remain on removing the `resources` usage and thus improving UX perf.

manuelroemer (commented):
Looks good to me generally, thanks a lot for refactoring towards SWR @vasharma05! 🚀

packages/esm-patient-registration-app/src/index.ts
vasharma05: Oh yes, totally forgot about it.
I'll just remove this and check for more redundant code places.

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
vasharma05: Yes, I'll remove this comment block too.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
vasharma05: Ohh I added this unintentionally. I'll just remove this too.

ZacButko (approved):

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
ibacher: @vasharma05 Would you mind re-adding that comment? Comments like this are how we tell i18next that we still need translations for these keys even though they aren't included here (because we literally call `t(name, ...)` to load the label text).

vasharma05: Sure @ibacher!

packages/esm-patient-registration-app/src/root.component.tsx
vasharma05: Sure!

denniskigen:
@vasharma05 hey, could you fix the conflicts?

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
```diff
@@ -23,42 +24,49 @@ export const AddressField: React.FC = () => {
...
+  const addressTemplateXml = addressTemplate?.results[0].value;
 
   useEffect(() => {
-    const templateXmlDoc = parseString(addressTemplateXml);
-    const nameMappings = getTagAsDocument('nameMappings', templateXmlDoc);
-    const elementDefaults = getTagAsDocument('elementdefaults', templateXmlDoc);
-    const properties = nameMappings.getElementsByTagName('property');
-    const propertiesObj = Array.prototype.map.call(properties, (property: Element) => {
-      const name = property.getAttribute('name');
+    if (addressTemplateXml) {
```
ZacButko, line 30: ewww... xml...

packages/esm-patient-registration-app/translations/en.json
```diff
@@ -29,32 +23,23 @@
   "estimatedMonthsLabelText": "Estimated Months",
   "estimatedYearsLabelText": "Estimated Years",
   "familyNameLabelText": "Family Name",
-  "familyNameRequired": "Family name is required",
```
ZacButko, line 32: Are you sure we don't need these translations anymore?

ZacButko (commented):
I think some of the translations may have been deleted accidentally. Also I'm not understanding where the performance improvement comes from?

packages/esm-patient-registration-app/translations/en.json
vasharma05: These translations got removed in the previous commit, after I removed the commented translations from address field.
Let me just put these all back.

Thank you for noticing it.

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
vasharma05: Yes, the adress fields fetched comes in the XML format.

packages/esm-patient-registration-app/src/index.ts
```diff
@@ -16,11 +16,8 @@ const backendDependencies = {
   'webservices.rest': '^2.24.0',
 };
 
-const resources = {
-  currentSession: fetchCurrentSession,
-  addressTemplate: fetchAddressTemplate,
-  relationshipTypes: fetchAllRelationshipTypes,
-  identifierTypes: fetchPatientIdentifierTypesWithSources,
+const frontendDependencies = {
+  '@openmrs/esm-framework': process.env.FRAMEWORK_VERSION,
```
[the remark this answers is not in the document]
vasharma05: It was added in by other changes, after I pulled the code from main.

packages/esm-outpatient-app/translations/en.json
```diff
@@ -36,7 +36,6 @@
   "fields": "of the following fields",
   "firstName": "First name",
   "firstNameSort": "First name (a-z)",
-  "futureScheduledVisits": " {count, plural, one{{count} visit} other{{count} visits}} scheduled for dates in the future",
```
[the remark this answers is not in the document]
vasharma05: Sure!

ZacButko (approved):

packages/esm-patient-registration-app/src/index.ts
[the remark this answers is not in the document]
ibacher: Looks like a merge error. This [isn't in the current version](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/index.ts) and shouldn't be here (we dropped support for the feature this was for).

packages/esm-patient-registration-app/src/offline.resources.ts
```diff
@@ -7,16 +7,16 @@ import { PatientIdentifierType, FetchedPatientIdentifierType } from './patient-r
 import { cacheForOfflineHeaders } from './constants';
 
 export interface Resources {
-  addressTemplate: any;
+  addressTemplate: undefined | any;
```
ibacher, line 10: This seems like a weird change? Surely `undefined` is covered by `any` and, in any case, could be represented by `addressTemplate?: any`

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
```diff
@@ -23,42 +24,49 @@ export const AddressField: React.FC = () => {
...
+      });
+      setAddressFields(propertiesObj);
+      // Do not remove the comments below
       /*
         t('postalCode')
         t('address1')
         t('stateProvince')
         t('cityVillage')
         t('country')
       */
```
ibacher, line 54: We should move these comments with the `labelText` line, since these comments document the translation keys we expect to support.

packages/esm-patient-registration-app/src/root.component.tsx
```diff
@@ -1,33 +1,41 @@
...
-  isOffline,
-}: RootProps) {
-  const resources = {
-    currentSession,
-    addressTemplate,
-    relationshipTypes,
-    identifierTypes,
-  };
+export default function Root({ savePatientForm, isOffline }: RootProps) {
+  const { data: currentSession } = useSWR<Session, Error>('patientRegistrationCurrentSessionInfo', fetchCurrentSession);
```
ibacher, line 23: Is there some reason for separately fetching the session in the registration? It seems like it would be better to just use the `useSession()` hook from esm-framework.

packages/esm-patient-registration-app/src/index.ts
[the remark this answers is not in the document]
vasharma05: Okay, I'll remove it.

packages/esm-patient-registration-app/src/patient-registration/field/address/address-field.component.tsx
vasharma05: Hi @ibacher!
Will the place matter? The reason is because whenever the translation key is called by the `t` hook, it looks into the `en.json` file, hence if the key is present in the `en.json`, there won't be any reason left for these keys to be placed near the labelText, what do you say?

packages/esm-patient-registration-app/src/root.component.tsx
vasharma05: Hi @ibacher! The reason for using the seprate session function is because the session value is cached in here and is then used in the offline functionality of the patient registration page.

I looked through the `useSession` hook and couldn't find, if it's being stored for offline functionality too.

@manuelroemer, can you please help me here.
Thanks!

ZacButko: Oh very interesting. Would it be appropriate to make the hook take care of both offline and online mode? That way the api is as clean as possible? What do you guys think?

ibacher: So, rather than have a separate hook to do the whole thing, I think the default pattern is just to tell the service worker to cache the path if the user has opted-in to offline mode as is done [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/ead9335b6328b9d7bd121605a2b6e0fea11f1af8/packages/esm-form-entry-app/src/index.ts#L25). Then the calls to `useSession()` should still be cached and we reduce the number of calls we're making to the session endpoint.

vasharma05:
I'm sorry everyone for not updating this PR. This PR is older than my work on the refactoring of the patient identifier in the patient registration, and hence I will have to redo the whole work. 
I'm closing this PR for the time being, but will come up soon, most probably before 23rd July.

Thanks for all the reviews @ZacButko @ibacher!

PR #201, opened 2022-05-11
(chore) CI workflow modifications

denniskigen:
## Requirements

## Summary

This PR makes the following changes to the CI workflow:

- Renames the workflow from `Node.js CI` to `OpenMRS CI`.
- Adds more descriptive names to steps in the `build` job.
- Amends the `verify` script in the lock file to remove the `test` job. Because `test` is run at the top level instead of per package, we cannot leverage the benefits of Turborepo. Running `turbo run test` doesn't do anything; you'd need to run `yarn run test` to run tests. I've added a test step using `yarn run test` to the CI that actually runs the tests.

Additionally, it amends the `pre-push` hook to leverage Turborepo.

## Screenshots

> Pre-push hook not working
Screenshot 2022-05-11 at 11 55 40

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Node.js CI
+name: OpenMRS CI
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


donaldkibet (approved):
LGTM Thanks @denniskigen

PR #212, opened 2022-05-19
(feat): More configurability for registration page

ibacher:
## Requirements

- [x] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [x] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [x] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

PR #215, opened 2022-05-20
BREAKING: Refactor and overhaul config schema

brandones:
## Requirements

## Summary

The main things:
- changing [Type.Object configs](https://github.com/openmrs/openmrs-esm-patient-management/blob/263e01753b5d38384b689f5e2eb91f66f53e58c7/packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts#L13) to Type.Array. So `fieldDefinitions: { foo: { label: "Foo" }}` would become `fieldDefinitions: [{ id: 'foo', label: 'Foo'}]`
- integrating the [`codedPersonAttributes`](https://github.com/openmrs/openmrs-esm-patient-management/blob/263e01753b5d38384b689f5e2eb91f66f53e58c7/packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts#L89) and `textBasedAttributes` sections into `fieldDefinitions`. I think there was a misunderstanding of how that was intended to be structured. I'm not sure what `fieldDefinitions` is supposed to do as written, but the idea is that `fieldDefinitions` is where you should define additional fields (whether coded person attribute, text person attribute, obs, whatever). So any configurations of `codedPersonAttributes` or `textBasedAttributes` should be rephrased into `fieldDefinitions`
pursuant to the above, `codedAttributes` and `textBasedAttributes` will no longer be built-in field names that can be referred to in `sectionDefinitions` as such
- [`sectionDefinitions`](https://github.com/openmrs/openmrs-esm-patient-management/blob/263e01753b5d38384b689f5e2eb91f66f53e58c7/packages/esm-patient-registration-app/src/config-schemas/openmrs-esm-patient-registration-schema.ts#L12) should be made to be piece-wise overridable, just like `fieldDefinitions`. What is currently the `_default` is made an independent `builtInSectionDefinitions` array, parts of which can be overridden using the `sectionDefinitions` config element

## Screenshots

Peek 2022-05-20 11-38

## Related Issue

*None.*

## Other

Supersedes https://github.com/openmrs/openmrs-esm-patient-management/pull/212


packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -0,0 +1,173 @@
...
+        _default: null,
+        _description: 'The label of the input. By default, uses the metadata `display` attribute.',
+      },
+      placeholder: {
+        _type: Type.String,
+        _default: '',
+        _description: 'Placeholder that will appear in the input.',
+      },
+      validation: {
+        required: { _type: Type.Boolean, _default: false },
```
ibacher, line 114: Maybe doesn't need to be dealt with in this PR, but it would be nice to have this handled at some point.

packages/esm-patient-registration-app/src/patient-registration/field/person-attributes/coded-person-attribute-field.component.tsx
```diff
@@ -0,0 +1,43 @@
...
+    <div className={`${styles.attributeField} ${styles.halfWidthInDesktopView}`}>
+      {!isLoadingConceptAnswers && conceptAnswers?.length ? (
+        <Select
+          id={`person-attribute-${personAttributeType.uuid}`}
+          name={`attributes.${personAttributeType.uuid}`}
+          labelText={label ?? personAttributeType?.display}
+          light>
+          {conceptAnswers.map((answer) => (
+            <SelectItem key={answer.uuid} value={answer.uuid} text={answer.display} />
+          ))}
```
ibacher, line 31: It would be a good idea to add a default blank option here. Otherwise it appears as if the first concept answer is always selected. Cf. how [Carbon-React implements placeholders](https://react.carbondesignsystem.com/?path=%2Fdocs%2Fcomponents-select--default).

packages/esm-patient-registration-app/src/patient-registration/section/section-wrapper.component.tsx
```diff
@@ -2,19 +2,29 @@ import React from 'react';
...
+
+export const SectionWrapper = ({ sectionDefinition, index }: SectionWrapperProps) => {
   const { t } = useTranslation();
+
   return (
-    <div id={id}>
+    <div id={sectionDefinition.id}>
       <h3 className={styles.productiveHeading02} style={{ color: '#161616' }}>
-        {index + 1}. {name}
+        {index + 1}. {sectionDefinition.name}
```
ibacher, line 19: This seems like a pain for i18n, but I don't really have a better suggestion off-hand. I suppose one "nice" feature of a fixed list of sections is it can just be pushed into the translations.

packages/esm-patient-registration-app/src/patient-registration/field/field.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+    case 'gender':
+      return <GenderField />;
+    case 'dob':
+      return <DobField />;
+    case 'address':
+      return <AddressField />;
+    case 'id':
+      return <IdField />;
+    case 'phone & email':
+      return <PhoneEmailField />;
```
ibacher, line 43: I suppose I forgot to mention this: one of the reasons I was trying to do away with the separate fields here is that they actually don't work as-is (mostly because they need to be submitted as attributes but aren't).

ibacher (approved):
Thanks @brandones! One small thing in the comments below, but I think this looks good!

packages/esm-patient-registration-app/src/config-schema.ts
brandones: Yeah, I think it's just not used at present? That should be fixed.

packages/esm-patient-registration-app/src/patient-registration/section/section-wrapper.component.tsx
brandones: Yeah I'm inclined to do something like `t(sectionDefinitions.name, sectionDefinitions.name)` so that translations can be used for whatever string is provided.

packages/esm-patient-registration-app/src/patient-registration/field/field.component.tsx
brandones: Yeah it kind of looked like they actually do nothing? Which is frustrating

brandones:
Shoot, forgot to mention that this needs https://github.com/openmrs/openmrs-esm-core/pull/434 and https://github.com/openmrs/openmrs-esm-core/pull/435

brandones:
Merging this. Heads up @vasharma05 and anyone else who might have an existing registration configuration.

PR #221, opened 2022-05-25
O3-283  Configurable Support for Recording Registration Observations

brandones:
## Requirements

## Summary

This makes it possible to configure the registration app to include obs.

I have also fixed up a few tests which either were being skipped or which, presumably by accident, did not test anything.

## Screenshots

Peek 2022-07-13 15-14

## Related Issue

https://issues.openmrs.org/browse/O3-283

### DIFF
--- a/packages/esm-outpatient-app/src/patient-info/appointment-details.component.tsx
+++ b/packages/esm-outpatient-app/src/patient-info/appointment-details.component.tsx
@@ -26,13 +26,13 @@ const PastAppointmentDetails: React.FC<PastAppointmentDetailsProps> = ({ pastApp
       {pastAppointments.length >= 1 ? (
         pastAppointments.map((appointment, index) => {
           return (
-            <>
+            <React.Fragment key={`past-appointment-${index}`}>
               <p className={styles.title}>{t('lastEncounter', 'Last encounter')}</p>
               <p className={styles.subtitle}>
                 {formatDatetime(parseDate(appointment.startDateTime))} · {appointment.service.name} ·{' '}
                 {appointment.location.name}{' '}
               </p>
-            </>
+            </React.Fragment>
           );
         })
       ) : (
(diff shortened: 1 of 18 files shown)


packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -146,6 +135,65 @@ export default class FormManager {
...
+  static async saveObservations(
+    obss: Record<string, string>,
+    savePatientResponse: FetchResponse,
+    currentLocation: string,
+    currentUser: Session,
+    config: RegistrationConfig,
+    abortController: AbortController,
+  ) {
+    if (Object.keys(obss).length > 0 && config.registrationObs.encounterTypeUuid) {
+      const encounterToSave = {
```
brandones, line 178: I was building this toward the REST API POST Encounter endpoint by mistake. Use the FHIR API.

packages/esm-outpatient-app/src/patient-info/appointment-details.component.tsx
```diff
@@ -26,13 +26,13 @@ const PastAppointmentDetails: React.FC<PastAppointmentDetailsProps> = ({ pastApp
       {pastAppointments.length >= 1 ? (
         pastAppointments.map((appointment, index) => {
           return (
-            <>
+            <React.Fragment key={`past-appointment-${index}`}>
```
brandones, line 29: Fixes warnings about needing unique `key`s when creating elements with `map`.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -9,14 +9,14 @@ export interface SectionDefinition {
 export interface FieldDefinition {
   id: string;
   type: string;
-  label: string;
+  label: string | null;
```
vasharma05, line 12: We can also use `label?: string` instead of using `|null`. Both representations denotes that this value can be null.

Just a suggestion.
Thanks!

vasharma05 (approved):
Great work @brandones!
Thank you!

vasharma05:
It'd be really helpful if you could share a small video, showing how is this working?
Thanks!

ibacher (approved):
LGTM

packages/esm-patient-registration-app/src/config-schema.ts
brandones: Optional and "possibly null" are distinct. Optional is more or less the same as `| undefined`. In this case, the object will always have the `label` property, but it will sometimes be null-valued.

brandones:
> It'd be really helpful if you could share a small video, showing how is this working? Thanks!

There is one in the PR description. Can you not see it? Or are there other details you're interested in?

gracepotma:
Woohoo!! Way to go Brandon! I can think of at least 3 different orgs who will want to use this right away. Thank you so much.

PR #243, opened 2022-06-14
Refactoring Patient identifiers field in the patient registration form

vasharma05:
## Requirements

## Summary

I have refactored the patient identifiers field into a different workflow, due to the repeated breaking of the patient registration page. I have completed the whole refactoring for both creating a new patient and editting the identifiers for an already existing patient.

**The following flows are covered in the new workflows:**
1. Creating a patient identifier, adding identifiers to the new patient to be created.
- [x] Managing different identifier sources for each identifierType for generation or manual entry

2. Editing an existing patient details
- [x] Updating a non-primary identifier
- [x] Deleting a non-primary identifier
- [x] Adding a new identifier to the patient.
- [x] Managing identifier sources for identifier types when editing an existing identifier and adding a new identifier.

## Screenshots

image

image

## Related Issue

*None.*

## Other

The PR #240 was made on the previous implementation of the patient identifiers, which will become obselete in this PR. I have taken into account the problem for discussed in #240 and have implemented the same in this PR for the new work, hence I'll be closing down the previous PR after this work of mine is approved for merging.
CC: @ZacButko @brandones


packages/esm-patient-registration-app/src/patient-registration/field/id/id-field.component.tsx
```diff
@@ -3,48 +3,113 @@ import { IdentifierInput } from '../../input/custom-input/identifier/identifier-
...
+  autoGeneration: boolean;
+  selectedSource: IdentifierSource;
+} {
+  const autoGeneration = identifierSource?.autoGenerationOption?.automaticGenerationEnabled;
+  return {
+    selectedSource: identifierSource,
+    autoGeneration,
+    identifierValue: autoGeneration
+      ? 'auto-generated'
+      : identifierValue !== 'auto-generated'
```
ZacButko, line 33: This is saying that if the identifier value is set to 'auto-generated' we would change it to initialValue, otherwise we set it to identifierValue ?? Are you sure this logic is correct?

packages/esm-patient-registration-app/src/patient-registration/field/id/identifier-selection-overlay.tsx
```diff
@@ -128,81 +97,67 @@ const PatientIdentifierOverlay: React.FC<PatientIdentifierOverlayProps> = ({
...
+                    name={`${identifierType?.fieldName}-identifier-sources`}
+                    defaultSelected={patientIdentifier?.selectedSource?.uuid}
+                    onChange={(sourceUuid: string) => handleSelectingIdentifierSource(identifierType, sourceUuid)}
+                    orientation="vertical">
+                    {identifierType?.identifierSources.map((source) => (
+                      <RadioButton
+                        key={source.uuid}
+                        labelText={source.name}
+                        name={source.uuid}
+                        value={source.uuid}
```
ZacButko, line 129: If we are controlling this component are we listening for user changes to this radio button? - mostly just regarding setting 'value' field

ZacButko (approved):
OK

packages/esm-patient-registration-app/src/patient-registration/field/id/id-field.component.tsx
vasharma05: Yes, if source was initially autoGeneration enabled, then the value will be 'auto-generated', but if we switch to a manual source, then we have to change back the 'auto-generated' to the intialValue, which will either be an empty string, or initial identifier value.

packages/esm-patient-registration-app/src/patient-registration/field/id/identifier-selection-overlay.tsx
vasharma05: Yes, the Radio Button Group handles the change.

PR #263, opened 2022-07-05
(feat) Offline Patient Search

manuelroemer:
## Requirements

## Summary
Makes the patient search work in offline mode. While offline, the following patients can be searched:
* Offline registered patients.
* Patients added to the offline patient list.

It searches the following attributes (if available on an offline patient object):
* Name
* Identifier value

## Other
The code can be simplified a lot if/when the patient search is migrated to use the FHIR API instead of the REST API. Then, conversions between the two data formats can be removed. The relevant code sections are marked with `TODO`s so that they can easily be spotted in the future.

### DIFF
--- a/packages/esm-patient-search-app/package.json
+++ b/packages/esm-patient-search-app/package.json
@@ -8,7 +8,7 @@
   "license": "MPL-2.0",
   "homepage": "https://github.com/openmrs/openmrs-esm-patient-management#readme",
   "scripts": {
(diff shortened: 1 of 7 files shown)


ZacButko (approved):
Thanks for this @manuelroemer . I believe @vasharma05 has just changed the patient search to use FHIR, so maybe a large chunk of translation code can be removed as you suggested.

PR #270, opened 2022-07-20
O3-1228:Coded drop-down fields should default to "no selected value"

jwnasambu:
## Requirements

## Summary

Currently, for coded attributes, the first option is always selected by default and by default no option should be selected. I therefore made the select Item value to be an empty string instead. 

## Screenshots

*None.*

## Related Issue

(https://issues.openmrs.org/browse/O3-1228)

## Other

*None.*


jwnasambu:
@brandones am still creating the screen though am having eye problem so it will take a bit long than expected.

brandones (approved):
I feel confident about this. Thanks @jwnasambu .

jwnasambu:
@brandones thanks for the review.

PR #271, opened 2022-07-20
Create README.md

gracepotma:
CC @brandones
It seemed like we should have some kind of instructions so that your work to configure the obs for the registration app doesn't get overlooked.

### DIFF
--- a/packages/esm-patient-registration-app/README.md
+++ b/packages/esm-patient-registration-app/README.md
@@ -2,6 +2,6 @@
 
 ## Configuring the Registration App to show your custom content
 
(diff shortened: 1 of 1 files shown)


packages/esm-patient-registration-app/README.md
```diff
@@ -0,0 +1,7 @@
+# esm-patient-registration-app
+
+## Configuring the Registration App to show your custom content
```
brandones, line 3:
```suggestion
## Configuring the Registration App to collect custom observations
```

brandones (approved):

PR #299, opened 2022-08-09
Lint fixes and enforcements in the patient management apps

vasharma05:
## Requirements

## Summary
Fixed all the warnings that came from the lint regarding the missing dependencies and also added lint enforcements with max limit for warnings to be 0.
This is also present in the pre-push and pre-push will fail. if there is a warning or a error found by lint.

Thanks @ZacButko!

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-active-visits-app/package.json
+++ b/packages/esm-active-visits-app/package.json
@@ -13,7 +13,7 @@
     "debug": "npm run serve",
     "build": "webpack --mode production",
     "analyze": "webpack --mode=production --env.analyze=true",
-    "lint": "eslint src --ext ts,tsx",
+    "lint": "eslint src --ext ts,tsx --fix --max-warnings=0",
     "typescript": "tsc",
     "extract-translations": "i18next 'src/**/*.component.tsx'"
   },
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -138,7 +138,7 @@ const ActiveVisitsTable = () => {
     if (currentPage !== 1) {
       goTo(1);
     }
-  }, [searchString]);
+  }, [searchString, currentPage, goTo]);
 
   if (isLoading) {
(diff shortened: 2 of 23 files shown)


dkayiwa:
> PR's like this are why I wish we had a solid testing pipeline. Updating the `useEffect`s especially are generally non-trivial and can cause undesirable behavior. Currently some of this code is tested with Jest, however I'm not sure jest would catch some of the more involved behavior from `useEffect`. To properly ward against this we should have better integrated testing with Cypress and a test environment to really understand if any feature was working before such a PR.
> 
> This is a lot to say I appreciate the work, but am not confident that all changes were thoroughly tested. If you stand by this work and are convinced we won't need to do a revert then we can merge it. Otherwise I'm more comfortable tackling this one package at a time and making sure to test changes all the way on dev3 server.

@jnsereko @gracepotma this is perfect example of why we need to seriously look into the QA testing pipeline.

jnsereko:
> PR's like this are why I wish we had a solid testing pipeline. Updating the `useEffect`s especially are generally non-trivial and can cause undesirable behavior. Currently some of this code is tested with Jest, however I'm not sure jest would catch some of the more involved behavior from `useEffect`. To properly ward against this we should have better integrated testing with Cypress and a test environment to really understand if any feature was working before such a PR.
> 
> This is a lot to say I appreciate the work, but am not confident that all changes were thoroughly tested. If you stand by this work and are convinced we won't need to do a revert then we can merge it. Otherwise I'm more comfortable tackling this one package at a time and making sure to test changes all the way on dev3 server.

i feel the pain to @ZacButko.  Updating hooks might break some parts of the system without a developer getting notified about this fault
@dkayiwa I am going to create tests for Patient Appointments ASAP 

@ZacButko @gracepotma @dkayiwa If there are critical spots that we have to create test cases for please kindly point them out through [Extending test cases for 3.x E2E testing](https://talk.openmrs.org/t/extending-test-cases-for-3-x-e2e-testing/37209)

ZacButko:
Per conversation, let's close this PR and do this upgrade one package at a time

PR #340, opened 2022-08-31
03-1490: Move the Patient Visit Header to the Patient Chart.

hadijahkyampeire:
## Requirements

## Summary

- Moved the visit header component to the patient chart instead of the outpatient app
## Dev3 screenshot with the visit header in outpatient
Screenshot 2022-08-31 at 22 49 45
## Local screenshot after moving it into the patient chart
Screenshot 2022-08-31 at 22 48 02

## Related Issue
https://issues.openmrs.org/browse/O3-1490
Related PR for removing it from the outpatient app. https://github.com/openmrs/openmrs-esm-patient-chart/pull/795


ZacButko:
Hi @donaldkibet . The motivation here are the tickets [O3-1161](https://issues.openmrs.org/browse/O3-1161) and [O3-1251](https://issues.openmrs.org/browse/O3-1251).

Currently the way that this is written is that it will render onto any page matching `/^patient\/.+\/chart/`, which I believe is the definition of the patient chart page? Would it not be appropriate to have this header which is only on patient chart be in the patient-chart-app ? 

If you are saying that this is being used in service queues for a different purpose I would agree that it would be bad to remove until we can make a solution there too. Currently on dev3 I am not able to see service queues working. Do you have a screenshot or video of when and how this is being used?

Again the motivation for this is to reduce complexity and load time by making direct components where they are used rather than the extension system. 

@hadijahkyampeire , Since this is removing an element that renders according to single-spa pages we need to be extra careful here that we are not breaking any existing workflows. Passing tests in this repo alone are not enough, since really any microfrontend could be depending on it. For @jnsereko , @jayasanka-sack this type of PR is another prime example of when it would be amazing to have full-scale integration tests including visual diffs . Right now if we don't manually check everywhere we think this could make a change across multiple repos we won't know if it is working right or not, and even then we might not know about other places that are depending on this existing.

hadijahkyampeire:
@ZacButko @donaldkibet I don't the visit header is used anywhere else and according to its design, it looks like it is a header specifically for the patient chart page. I agree it will have some components of the service queue but that's an endpoint that we can just call on the patient chart and get the service queue data.

ZacButko (approved):
OK thanks @donaldkibet . @hadijahkyampeire let's move forward with this then. Merge https://github.com/openmrs/openmrs-esm-patient-chart/pull/795 first, then this one

hadijahkyampeire:
Thanks @ZacButko , I still have one small thing to do on the other PR now that dev3 has been fixed, I will let you know when it is ready for approval.

PR #342, opened 2022-09-01
(fix) Fix tests for the Date of Birth component

denniskigen:
## Requirements

## Summary

This PR provides a bunch of anticipatory fixes necessary to support a tweak to the registration form. More specifically, @makombe wishes to change the type of `datePickerType` prop from `simple` to `single`. The big benefit from that would be the provision of a calendar UI from which users can more conveniently select the date of the birth value. Applying that change with the current setup yields a test failure. This failure is caused by the react module mock in `dob.test.tsx`. More specifically, that mock went on to provide a partial `useContext` mock. This wound up breaking some of the Carbon date picker component functionality that depends on `useContext`.

This PR gets rid of module mocks for React and Formik in the `DobComponent` test suite and refactors the `render` function to return a component wrapped in a `PatientRegistrationContext`. This refactor enables the test to pass both with the current setup as well as with @makombe's anticipated change.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.test.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.test.tsx
@@ -1,54 +1,62 @@
-import React, { useContext } from 'react';
-import { render, screen, fireEvent } from '@testing-library/react';
+import React from 'react';
+import { Formik, Form } from 'formik';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import '@testing-library/jest-dom/extend-expect';
 import '@testing-library/jest-dom';
-import { Formik, Form, useField } from 'formik';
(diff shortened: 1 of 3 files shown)


packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.test.tsx
```diff
@@ -1,54 +1,62 @@
...
       <Form>
-        <DobField />
+        <PatientRegistrationContext.Provider
+          value={{
+            identifierTypes: [],
+            values: formValues,
+            validationSchema: null,
+            setValidationSchema: (value) => {},
+            inEditMode: false,
+            setFieldValue: () => {},
```
denniskigen, line 52: This test setup isn't the most ideal because toggling tabs using the content switcher does not propagate changes down via `setFieldValue`. However, it will suffice for now.

makombe (approved):
LGTM

PR #378, opened 2022-09-19
03 1527: Update appointment status.

hadijahkyampeire:
## Requirements

## Summary

- Implemented an actionable notification component to use for toasts.
- Handle all status changes `CheckIn`, `Complete`, `Missed`,  and `Cancel`.
- Also use real data from the API

## Screenshots
Screenshot 2022-09-19 at 13 55 16
Screenshot 2022-09-19 at 13 55 36
Screenshot 2022-09-19 at 13 56 32

## Related Issue

https://issues.openmrs.org/browse/O3-1527

## Other
- Still have issues with instant updating the status in the table after a successful change, it seems the mutate takes long or I need some kind of refresh


packages/esm-appointments-app/src/helpers/functions.ts
```diff
@@ -63,7 +63,9 @@ export const getTodaysAppointment = (appointment: Appointment) => {
     appointmentNumber: appointment.appointmentNumber,
     color: appointment.service.color,
     identifier: appointment.patient?.identifier,
-    duration: getAppointmentDuration(appointment.startDateTime, appointment.endDateTime),
+    duration: appointment.service?.durationMins
+      ? appointment?.service?.durationMins + 'min'
```
ZacButko, line 67: It would be best if this was translated

packages/esm-appointments-app/src/home-appointments/actionable-notifications/notification.component.scss
```diff
@@ -0,0 +1,6 @@
+.actionableNotification {
+  z-index: 100000;
+  position: fixed;
+  width: 100%;
+  bottom: 2rem;
```
ZacButko, line 5: spacing-07

packages/esm-appointments-app/src/home-appointments/actionable-notifications/notification.component.scss
```diff
@@ -0,0 +1,6 @@
+.actionableNotification {
+  z-index: 100000;
```
ZacButko, line 2: z-index is outrageous... probably ok with something less than 1000

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
```diff
@@ -9,35 +9,37 @@ import {
...
 import { Add, CheckmarkOutline } from '@carbon/react/icons';
-import { isDesktop, useLayoutType, ConfigurableLink, useConfig, navigate } from '@openmrs/esm-framework';
+import {
+  isDesktop,
+  useLayoutType,
+  ConfigurableLink,
+  useConfig,
+  navigate,
+  showToast,
+  showNotification,
```
ZacButko, line 28: This seems like an unused import. Use or delete?

packages/esm-appointments-app/src/home-appointments/appointments-list.scss
```diff
@@ -95,6 +95,7 @@
   color: $support-02;
   display: flex;
   align-items: center;
+  height: 3rem;
```
ZacButko, line 98: spacing-09

packages/esm-appointments-app/translations/en.json
```diff
@@ -10,36 +10,43 @@
...
+  "appointmentMissed": "Appointment Missed",
+  "appointmentMissedError": "Error marking appointment as Missed",
   "appointmentNoteLabel": "Write an additional note",
   "appointmentNotePlaceholder": "Write any additional points here",
   "appointmentNotes": "Appointment Notes",
   "appointmentNowVisible": "It is now visible on the Appointments page",
   "appointments": "Appointments",
   "appointmentScheduled": "Appointment scheduled",
   "appointmentsList": "Appointment list is empty",
+  "appointmentStatusChange": "",
```
ZacButko, line 24: empty translation?

ZacButko (approved):
I have mixed feelings about this one. Ideally a new mechanism very similar to our existing [showNotification](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/docs/API.md#shownotification) should also live in core so it can be used by all microfrontends. However doing so means that the design would need to be well thought out and added to our styleguide, and if it's only being used here would possibly go against "Avoid Hasty Abstractions". So this is fine for now, but lets keep a close eye on this if we can get it added to the official styleguide and then abstracted into core.

ZacButko:
@hadijahkyampeire , After talking with Ciaran on the design call today we've decided that this is a valid variant of the `@carbon/react` [inline notification ](https://carbondesignsystem.com/components/notification/usage/) - so therefore we should extend the core `showNotification` to support the action button, rather than create an entirely new component here. Are you able to make this change?

packages/esm-appointments-app/src/home-appointments/actionable-notifications/notification.component.scss
hadijahkyampeire: oh sorry I forgot about the spacing variables, I will fix that.

ZacButko:
Re: Extensive discussion today with @hadijahkyampeire and asynchronously with Ciaran, For now the action will be to merge this, then in the near-term future implement a global snackbar-like-component with an action button that will leverage carbon's new ActionableNotification with `inline` parameter passed. This will then be available via `esm-framework` probably by a function called `showActionableNotification`

hadijahkyampeire:
@ZacButko @rbuisson I just realized that all appointments are usually returned even the `canceled` and `missed` so I also implemented a small component to render statuses differently for easy visualization. Let me what you think about this. I can do it in a follow-up PR.
Screenshot 2022-09-20 at 02 29 07

PR #383, opened 2022-09-23
03-1536: Add privileges to today's appointments widget.

hadijahkyampeire:
## Requirements

## Summary

- Check if the userHasAccess and they want to use the privileges, by default the use of privileges is `false`. If you want to use privileges you can turn it true in the config. Else you will have a full view.
```{
  "@openmrs/esm-appointments-app": {
    "useFullViewPrivilege": true
  }
}
```
If you are using privileges and have limited access, the view will be limited.

## Screenshots
Limited Access View. Can only see `Scheduled`,  add, see all or checkin appointments.
Screenshot 2022-09-23 at 22 28 37

Full View Access: Can do all actions and can see all appointments
Screenshot 2022-09-23 at 22 27 55

## Related Issue
[Add privileges to today's appointments widget.](https://issues.openmrs.org/browse/O3-1536)

### DIFF
--- a/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
+++ b/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
@@ -232,7 +232,7 @@ const AppointmentsBaseTable = () => {
           overflowMenuOnHover={isDesktop(layout)}
           rows={tableRows}
           size={isDesktop(layout) ? 'xs' : 'md'}
-          useZebraStyles={filteredAppointments?.length > 1}>
+          useZebraStyles={true}>
           {({ rows, headers, getHeaderProps, getTableProps, getRowProps }) => (
             <TableContainer className={styles.tableContainer}>
               <Table {...getTableProps()} className={styles.appointmentsTable}>
--- a/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
+++ b/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
@@ -91,8 +91,7 @@ const AppointmentsBaseTable = () => {
   const { useBahmniAppointmentsUI: useBahmniUI, useFullViewPrivilege, fullViewPrivilege } = useConfig();
   const { isLoading, appointments } = useTodayAppointments();
 
-  const userHasFullAccess = useFullViewPrivilege && userHasAccess(fullViewPrivilege, user);
-  const fullView = userHasFullAccess || !useFullViewPrivilege;
+  const fullView = userHasAccess(fullViewPrivilege, user) || !useFullViewPrivilege;
 
   const filteredAppointments = !fullView
     ? appointments.filter((appointment) => appointment.status === 'Scheduled')
--- a/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
+++ b/packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
@@ -1,11 +1,10 @@
-import React, { useMemo, useState, useEffect } from 'react';
+import React, { useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   Button,
   DataTable,
   Layer,
   DataTableSkeleton,
-  DataTableHeader,
   Table,
   TableBody,
   TableCell,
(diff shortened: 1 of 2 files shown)


packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
```diff
@@ -90,17 +87,34 @@ const RenderStatus = ({ status, t, appointmentUuid, mutate }) => {
...
+  const { useBahmniAppointmentsUI: useBahmniUI, useFullViewPrivilege, fullViewPrivilege } = useConfig();
   const { isLoading, appointments } = useTodayAppointments();
 
+  const [fullView, setFullView] = useState(false);
+
+  const filteredAppointments = !fullView
+    ? appointments.filter((appointment) => appointment.status === 'Scheduled')
+    : appointments;
+
+  const userHasFullAccess = useFullViewPrivilege && userHasAccess(fullViewPrivilege, user);
```
ZacButko, line 101: Whether a user has full access or not should not be dependent on if the system is currently using `useFullViewPrivilege`. Is it safe to remove the `useFullViewPrivilege` in this statement?

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
```diff
@@ -90,17 +87,34 @@ const RenderStatus = ({ status, t, appointmentUuid, mutate }) => {
...
+  useEffect(() => {
+    if (!useFullViewPrivilege || userHasFullAccess) {
+      setFullView(true);
+    } else {
+      setFullView(false);
+    }
+  }, []);
```
ZacButko, line 108: I don't think that this needs to be a useEffect. `const fullView = userHasFullAccess || !useFullViewPrivilege` should be sufficient

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
```diff
@@ -227,7 +241,7 @@ const AppointmentsBaseTable = () => {
           overflowMenuOnHover={isDesktop(layout)}
           rows={tableRows}
           size={isDesktop(layout) ? 'xs' : 'md'}
-          useZebraStyles={appointments?.length > 1}>
+          useZebraStyles={filteredAppointments?.length > 1}>
```
ZacButko, line 244: Why isn't this just `{true}` ?

ZacButko (changes requested):
Thanks Hadijah. I think it would be good to remove the useState and useEffect if possible. It doesn't seem required for this case -> https://beta.reactjs.org/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

ZacButko:
Also, as a side note, the useEffect should have thrown linting warnings for the missing dependencies, which tells that linting is not enforced on this package. @hadijahkyampeire would you be able to turn on lint enforcement for this package? See [this PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/299) for how to do this, but before @vasharma05  turned it on for the whole repo at once. Instead let's just do it one package at a time, starting with this one, if possible.

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
hadijahkyampeire: From what I discussed with Romain we want to give a user the choice to decide whether they want to apply the use of privileges, that is where the boolean comes in. So He says we might not want to apply privileges for everyone so that is why he suggests we make it configurable.

So if someone doesn't want to use any privileges they will see everything, from what I understood, he meant forexample for OpenMRS we might not want to have any limitations to the table, so that's why we need that boolean.

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
hadijahkyampeire: I think it is because when it is one row, it resembles the headers but I think this is minor, I will just make it true.

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
ZacButko: Yes, understood. However that logic takes place later. Here you are just defining whether they have access or not, *if* access was to be considered. Your filter logic happens next. Basically if you combine the two statements I think you would see useFullViewPrivilige twice, meaning it can be removed here

hadijahkyampeire: hmmm, I think I am now confused, maybe I understood it wrongly.

ZacButko: 1. Your original statement is
`const userHasFullAccess = useFullViewPrivilege && userHasAccess(priv, user);
const fullView = userHasFullAccess || !useFullViewPrivilege;`

2. Substitute for userHasFullAccess
`const fullView = (useFullViewPrivilige && userHasAccess(priv, user)) || !useFullViewPrivilige`

3. Distributive law
`const fullView = (!useFullViewPrivilige || useFullViewPrivilige) && (!useFullViewPrivilige || userHasAccess(priv, user)`

4. Since `!a || a` is always true, we can eliminate it
`const fullView = !useFullViewPrivilige || userHasAccess(priv, user)`

ZacButko (approved):

packages/esm-appointments-app/src/home-appointments/appointments-list.component.tsx
hadijahkyampeire: Yah you are right I now understand I will push a change tomorrow. Thanks

PR #404, opened 2022-10-05
Added configuration for switching between Address Hierarchy and simple address fields

vasharma05:
## Requirements

## Summary
Adding a configuration for switching between using the address hierarchy and simple address fields.

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


packages/esm-patient-registration-app/src/patient-registration/field/address/address-hierarchy.component.tsx
```diff
@@ -78,17 +85,27 @@ export const AddressHierarchy: React.FC = () => {
...
+              <ComboInput
+                key={`combo_input_${index}`}
+                name={attributes.name}
+                labelText={t(attributes.name)}
+                id={attributes.name}
+                setSelectedValue={setSelectedValue}
+                selected={selected}
+              />
+            ))
+          : addressLayout.map((attributes, index) => (
```
ZacButko, line 99: I'm really weirded out that this is using the same variable `addressLayout` for the dynamic one as the static version... Is this really the case?

packages/esm-patient-registration-app/src/patient-registration/field/address/address-hierarchy.component.tsx
```diff
@@ -78,17 +85,27 @@ export const AddressHierarchy: React.FC = () => {
...
-            setSelectedValue={setSelectedValue}
-            selected={selected}
-          />
-        ))}
+        {useAddressHeirarchy
+          ? addressLayout.map((attributes, index) => (
+              <ComboInput
+                key={`combo_input_${index}`}
+                name={attributes.name}
+                labelText={t(attributes.name)}
```
ZacButko, line 93: This works, but is missing the fallback text for when `eact-i18next` fails. Do you think its reasonable to add this here and for the static version as well?

ZacButko (approved):
Thanks Vineet. I'm just curious about some things.

packages/esm-patient-registration-app/src/patient-registration/field/address/address-hierarchy.component.tsx
vasharma05: The address layout provides the fields to be shown on the patient registration page. Hence we are using the same address layout variable/

packages/esm-patient-registration-app/src/patient-registration/field/address/address-hierarchy.component.tsx
vasharma05: We have defined the translation keys a few lines above this one, hence the keys and values for the address fields are stored in the translation file and is taken for the specific key.

ZacButko: Right, but we've seen that translation file be ignored if react-i18next isn't loaded properly. Our site-wide syntax for this as far as i am aware is `t("variableName", "Fallback Text")`. What do you think?

PR #407, opened 2022-10-06
Address Hierarchy in the Patient Registration

vasharma05:
## Requirements

## Summary

Address hierarchy is added in the patient registration page, where the user can search through the addresses saved in the backend and selecting an address will fill the address on the form.

## Screenshots

https://user-images.githubusercontent.com/51502471/194228273-161cd239-31cf-49ab-86e1-4d4cd83f04e8.mov

## Related Issue

*None.*

## Other

*None.*


packages/esm-patient-registration-app/src/patient-registration/field/address/address-search.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [searchString, setSearchString] = useState<string>('');
+  const { addresses, isLoading, error } = useAddressHierarchy(searchString);
+  const { setFieldValue } = useFormikContext();
+
+  const handleInputChange = (str) => {
+    setSearchString(str);
+  };
+
+  const handleChange = ({ selectedItem }) => {
+    console.log(selectedItem);
```
ZacButko, line 22: Please turn on linting for patient registration

packages/esm-patient-registration-app/src/patient-registration/field/address/address-search.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+    const [country, stateProvince, countyDistrict, address1, cityVillage] = selectedItem.split(', ');
+    console.log({
```
ZacButko, line 24: Please turn on linting to catch these console.logs

packages/esm-patient-registration-app/src/patient-registration/field/address/address-search.component.tsx
```diff
@@ -0,0 +1,68 @@
...
+              ? isLoading
+                ? t('loadingResults', 'Loading results')
+                : error
+                ? error.message
+                : addresses?.length === 0
+                ? t('noResultsFound', 'No results found')
+                : null
+              : null
+          }
+          light
```
ZacButko, line 61: shouldn't need the `light` parameter since you're using `Layer`

ZacButko (changes requested):
Please remove console.logs

ZacButko (approved):

vasharma05:
Thanks @ZacButko for the review!

vasharma05:
@donaldkibet, I am working to restore the previous built version along with this new change, so that this won't be blocked for now.
Thanks!

vasharma05:
@donaldkibet, with the version of the addresses implemented, we were facing this issue to fill in the address registration forms, due to lack of data from the backend.

Since this data isn't mandatory, this totally blocks us out to fill in the address


https://user-images.githubusercontent.com/51502471/194269803-cd3f9146-8355-414c-b7fa-6c7fdf27b2a7.mov

vasharma05:
Hi @donaldkibet !
I have restored your request, and you can use your address implementation by changing the following configuration.

```
fieldConfigurations:{
     address: {
           useAddressHierarchy: true -> false
     }
}
```

hadijahkyampeire (commented):
LGTM, About Donald's issue, can't we find a way to keep both implementations, the config will work but users might miss out if they are not aware of changing the config. We could make the search input also allow user inputs if the address doesn't match any in the DB, and while saving, save all the address info and each API will pick what they want. @vasharma05 we should talk more about this later today, I think there is a better way to address this.

vasharma05:
Yes, the PR #404 was removed the implementation and now I'm trying to restore your implementation.
What I am worried about is which particular solution should we go with, taking in consideration the community implementation and the Paladium implementation.

vasharma05:
@donaldkibet, I've been meaning to connect to you and I messaged you on Slack. Do you have some time tomorrow for a meeting on this topic?

vasharma05:
Hi @donaldkibet!
Please see the updated PR description on the changes.
Thanks!

hadijahkyampeire (approved):
Let's merge it and test it

makombe:
> Yes, the PR #404 was meant to remove the Paladium implementation and now I'm trying to restore Paladium's implementation. What I am worried about is which particular solution should we go with, taking in consideration the community implementation and the Paladium implementation.

The address hierarchy was not needed by Palladium alone, it was ohri Ethiopia who implemented the feature and we also needed the same. So it came to our advantage. My thought is that we should be as accommodating as possible lest people are likely to branch out and do their own things which is not the spirit of the community.

makombe:
@vasharma05 Please resolve the conflicts then I will be able to test and give feedback.

vasharma05:
Hey @donaldkibet and @makombe , please feel free to test the changes.

Also, @makombe , I have also pushed the fix for the failing API request for fetching field values due to the incorrect field name sent to the endpoint.
Thanks!

makombe:
> 

Thanks @vasharma05  I have tested and it works. The search feature awesome, it makes the work of populating the fields much easy. Maybe if we can be able to make the labels configurable instead of them being address4,stateProvince etc. Is there away we can override the labels?
Screenshot from 2022-10-11 23-03-07

makombe (approved):

PR #412, opened 2022-10-06
Revert "Added configuration for switching between Address Hierarchy and simple address fields"

donaldkibet:
Proposing to revert this because its breaking the `Address Hierarchy` feature already existing. @ZacButko @denniskigen @rbuisson


ibacher:
We still need this feature for distributions that do not use the address hierarchy at all (and there are several of those).

makombe:
@ibacher I am also of the opinion that we need to make things as configurable as possible instead of removing and not minding of other implementors . I had made it possible even for other distributions who do not need address hierarchy to be able to configure what they need #374

PR #422, opened 2022-10-13
O3-1561: (Address Hierarchy) Show search results in a manner that only the above levels of the search string are shown and are selected

vasharma05:
## Requirements

## Summary
This PR brings changes to the Address Hierarchy in the Patient registration page.
### Currently
When we search an address, it shows hundreds of results from the highest hierarchical field to the lowest ones.
E.g. When I search for Battambang state in Cambodia, it returns all the search results including every district, address and village in Battambang.
image
When we select an option, it fills all the fields as shown below:
image

### Changes from PR
When we search Battambang, it will only show the higher hierarchical positions to the search string, as shown in the image:
image
image

When we select an option, it only fills the appropriate fields, except filling all the fields.
image

## Screenshots

*Attached above*

## Related Issue

https://issues.openmrs.org/browse/O3-1561

## Other

*None.*

### DIFF
--- a/packages/esm-patient-registration-app/src/config-schema.ts
+++ b/packages/esm-patient-registration-app/src/config-schema.ts
@@ -238,7 +238,7 @@ export const esmPatientRegistrationSchema = {
       useAddressHierarchy: {
         enabled: {
           _type: Type.Boolean,
-          _description: 'Whether to use the Address heirarchy in the registration form or not',
+          _description: 'Whether to use the Address hierarchy in the registration form or not',
           _default: true,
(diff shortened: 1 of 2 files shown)


rbuisson (approved):
Can't review the technicalities but it does seem to match the business requirement. Thanks @vasharma05 .

hadijahkyampeire (approved):
Code looks good as well, thanks @vasharma05

vasharma05:
Thanks @rbuisson and @hadijahkyampeire !!!

PR #424, opened 2022-10-13
(chore) Fix husky by using postinstall hook

vasharma05:
## Requirements

## Summary
According to the statement on the Husky website [here](https://typicode.github.io/husky/#/?id=install), it states that
```
Yarn 2+ doesn't support prepare lifecycle script, so husky needs to be installed differently (this doesn't apply to Yarn 1 though). See [Yarn 2+ install](https://typicode.github.io/husky/#/?id=yarn-2).
```
Hence I have incorporated the changes according to the instructions [here](https://typicode.github.io/husky/#/?id=yarn-2).

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


packages/esm-appointments-app/src/appointment-forms/appointments-form.component.tsx
```diff
@@ -53,7 +53,7 @@ interface AppointmentFormProps {
   patientUuid?: string;
   context: string;
 }
-const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment = {}, patientUuid, context }) => {
+const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment, patientUuid, context }) => {
```
ibacher, line 56: This change isn't going to break anything is it?

ibacher (approved):
Nice catch @vasharma05!

packages/esm-appointments-app/src/appointment-forms/appointments-form.component.tsx
hadijahkyampeire: It is not I confirmed it.

package.json
```diff
@@ -58,6 +58,7 @@
     "jest-environment-jsdom": "^28.1.0",
     "lerna": "^3.20.2",
     "openmrs": "latest",
+    "pinst": "^3.0.0",
```
ibacher, line 61: Do we need this dependency? My guess is no for this setup (since the monorepo package is not published.

hadijahkyampeire (approved):
Thanks @vasharma05

package.json
vasharma05: Thank you @ibacher for the catch, I removed it.

PR #477, opened 2022-11-21
(fix): Fix translations for sex options

makombe:
## Requirements

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


ibacher (approved):
Seems like "unknown" didn't get added to the translations? Otherwise, this is great!

makombe:
@ibacher perused through translation file and unknown is already available. I think that is why it has been omitted.

ibacher:
It's probably ok for now... We probably need to start some kind of concept-tagging for these single-word translations, since translation is usually a matter of conceptual translation rather than word translation, i.e., to be able to distinguish between "unknownGender" and "unknownCauseOfDeath" which [seems to be where the string is coming from](https://github.com/openmrs/openmrs-esm-patient-management/blob/68758e0eb92511d20314cb0ef7ad8d863f6d3829/packages/esm-patient-registration-app/src/patient-registration/section/death-info/death-info-section.component.tsx#L21).

PR #481, opened 2022-12-01
Adds breadcrumbs to appointments calendar

pirupius:
## Requirements

## Summary
Currently the calendar screen in appointments has no way of navigating backwards to the main appointments screen. The only options would be pressing the back button on the browser. 

This PR adds breadcrumbs for easier navigation.

## Screenshots
Screenshot 2022-12-01 at 15 59 00


denniskigen (approved):
LGTM. Thanks, @pirupius!

pirupius:
@denniskigen That seems like a nice alternative but i think the current design was based off the zeplin/sketch. 
I on the other hand feel like we should remove that sidenav all together since it has only one menu item but that is something that might need to be tabled first to the other stakeholders + designers. 

Am going to go ahead and merge this.

PR #495, opened 2022-12-28
Possibility to display desired user keys by configuration

icrc-agomes:
## Requirements

## Summary
In Active visits, we now have the possibility to display multiple user keys through the configuration. 

The configuration contemplates the structure to add the desired header and respective key of the field.
The keys are programmed to be displayed between the visit time and the user's name, if no configuration is provided the widget will behave normally.

## Screenshots
image

Where the black boxes are there should be the respective keys.

### DIFF
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -73,7 +73,7 @@ const ActiveVisitsTable = () => {
 
   const computeHeaderData = useCallback(
     (t, config) => {
-      let headers = [
+      const headers = [
         {
           id: 0,
           header: t('visitStartTime', 'Visit Time'),
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -71,7 +71,7 @@ const ActiveVisitsTable = () => {
   const currentPathName = window.location.pathname;
   const fromPage = getOriginFromPathName(currentPathName);
 
-  const computeExpensiveValue = useCallback(
+  const computeHeaderData = useCallback(
     (t, config) => {
       let headers = [
         {
@@ -120,53 +120,7 @@ const ActiveVisitsTable = () => {
     [activeVisits],
   );
 
-  const headerData = useMemo(() => computeExpensiveValue(t, config), [computeExpensiveValue, config, t]);
-  //   let headers = [
-  //     {
-  //       id: 0,
-  //       header: t('visitStartTime', 'Visit Time'),
-  //       key: 'visitStartTime',
-  //     },
-  //     {
-  //       id: 1,
-  //       header: t('idNumber', 'ID Number'),
-  //       key: 'idNumber',
-  //     },
-  //     {
-  //       id: 2,
-  //       header: t('name', 'Name'),
-  //       key: 'name',
-  //     },
-  //     {
-  //       id: 3,
-  //       header: t('gender', 'Gender'),
-  //       key: 'gender',
-  //     },
-  //     {
-  //       id: 4,
-  //       header: t('age', 'Age'),
-  //       key: 'age',
-  //     },
-  //     {
-  //       id: 5,
-  //       header: t('visitType', 'Visit Type'),
-  //       key: 'visitType',
-  //     },
-  //   ];
-
-  //   console.log('activeVisits', activeVisits);
-
-  //   config.identifiers.map((identifier) => {
-  //     headers.push({
-  //       id: identifier.id,
-  //       header: t(identifier.header.key, identifier.header.default),
-  //       key: identifier.header.key,
-  //     });
-  //   });
-
-  //   console.log('headers', headers);
-  //   return headers;
-  // };
+  const headerData = useMemo(() => computeHeaderData(t, config), [computeHeaderData, config, t]);
 
   const rowData = activeVisits.map((visit) => ({
     ...visit,
--- a/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
+++ b/packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
@@ -71,41 +71,102 @@ const ActiveVisitsTable = () => {
   const currentPathName = window.location.pathname;
   const fromPage = getOriginFromPathName(currentPathName);
 
-  const headerData = useMemo(
-    () => [
-      {
-        id: 0,
-        header: t('visitStartTime', 'Visit Time'),
-        key: 'visitStartTime',
-      },
-      {
-        id: 1,
-        header: t('idNumber', 'ID Number'),
-        key: 'idNumber',
-      },
-      {
-        id: 2,
-        header: t('name', 'Name'),
-        key: 'name',
-      },
-      {
-        id: 3,
-        header: t('gender', 'Gender'),
-        key: 'gender',
-      },
-      {
-        id: 4,
-        header: t('age', 'Age'),
-        key: 'age',
-      },
-      {
-        id: 5,
-        header: t('visitType', 'Visit Type'),
-        key: 'visitType',
-      },
-    ],
-    [t],
+  const computeExpensiveValue = useCallback(
+    (t, config) => {
+      let headers = [
+        {
+          id: 0,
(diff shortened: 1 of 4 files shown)


packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -71,42 +71,62 @@ const ActiveVisitsTable = () => {
...
+          id: 0,
+          header: t('visitStartTime', 'Visit Time'),
+          key: 'visitStartTime',
+        },
+      ];
+
+      config?.activeVisits?.identifiers?.map((identifier, index) => {
+        headers.push({
+          id: index + 1,
+          header: t(identifier?.header?.key, identifier?.header?.default),
```
hadijahkyampeire, line 87: I don't think these translations will actually work, we should probably find strings for them.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
```diff
@@ -37,18 +39,89 @@ export function useActiveVisits() {
...
+      visitType: visit?.visitType?.display,
+      visitUuid: visit.uuid,
+    };
+
+    //in case no configuration is given the previsous behavior remanes the same
+    if (!config?.activeVisits?.identifiers) {
+      config.activeVisits.identifiers = [
+        {
+          header: {
+            key: 'idNumber',
```
hadijahkyampeire, line 61: Shouldn't this be translated?

hadijahkyampeire (commented):
Thanks @icrc-agomes for this, I realized you have used too much logic to compute headers but I think the logic can be simplified, if we are sure of the columns that we can possibly have then we just make sure the ids are not the same, I don't think they necessarily have to follow each other in order, I think we can have 0, 3, 4, 5 and they can still work. I will try it out locally and share with you. Also I have left small comments regarding translating headers.

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
icrc-agomes: Hello @hadijahkyampeire, thanks for the reply.

Regarding the headers, I developed having in count that I don't know how many columns the client desires (although if no configuration is provided the previous behavior will work). 
Also, the client wanted these headers to be placed between the **_Visit Time_** and **_Name_**, thus the complexity of the headers and indexation.

I tried today the translations one more time with the Spanish file and it was working.

You can try with this config and respective keys in translation files:


```
{
  "@openmrs/esm-home-app": {
    "extensionSlots": {
      "homepage-widgets-slot": {
        "order": [
          "active-visits-widget"
        ],
        "configure": {
          "active-visits-widget": {
            "activeVisits": {
              "identifiers": [
                {
                  "header": {
                    "key": "idNumber",
                    "default": "ID Number"
                  },
                  "identifierName": "HSU ID"
                },
                {
                  "header": {
                    "key": "idLegacy",
                    "default": "ID Legacy"
                  },
                  "identifierName": "Legacy ID"
                }
              ],
              "attributes": [
                {
                  "header": {
                    "key": "telephoneNumber",
                    "default": "Telephone Number"
                  },
                  "display": "Telephone Number"
                }
              ]
            }
          }
        }
      }
    }
  }
}
```

hadijahkyampeire: okay, thanks @icrc-agomes I will test it with the configurations, thanks.

icrc-agomes: Hello @hadijahkyampeire, just a small reminder when you are able to review this PR please do it :)
Thanks a lot!

delphinepas: @hadijahkyampeire do you have any update regarding this PR? thank you cc @icrc-agomes @icrc-fdeniger @denniskigen

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -71,42 +71,62 @@ const ActiveVisitsTable = () => {
...
+          key: identifier?.header?.key,
+        });
+      });
+
+      config?.activeVisits?.attributes?.map((attribute, index) => {
+        headers.push({
+          id: index + config?.activeVisits?.identifiers.length + 1,
```
ibacher, line 94: Can we change the logic here so that we confirm that `config.activeVisits.identifiers` exists and has a value or set a constant to 0? If this value is unset, i.e., if the expression is `undefined`, the result of this will always be `NaN`.

Basically, this should be:

```
const indexPadding = config?.activeVisits?.identifiers.length ?? 0;
// map using `indexPadding`
```

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -71,42 +71,62 @@ const ActiveVisitsTable = () => {
...
+          header: t(attribute?.header?.key, attribute?.header?.default),
+          key: attribute?.header?.key,
+        });
+      });
+
+      headers.push(
+        {
+          id: headers[headers.length - 1].id + 2,
```
ibacher, line 102: Can we just have one variable tracking the current index for all of these instead of three different index calculations?

packages/esm-active-visits-app/src/active-visits-widget/active-visits.component.tsx
```diff
@@ -71,42 +71,62 @@ const ActiveVisitsTable = () => {
...
-        key: 'age',
-      },
-      {
-        id: 5,
-        header: t('visitType', 'Visit Type'),
-        key: 'visitType',
-      },
-    ],
-    [t],
+  const computeHeaderData = useCallback(
```
ibacher, line 74: Can we just lift this function out of the component altogether?

packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
```diff
@@ -37,18 +39,89 @@ export function useActiveVisits() {
...
+
+    //map identifires on config
+    config?.activeVisits?.identifiers?.map((configIdentifier) => {
+      //check if in the current visit the patient has in his identifiers the current identifierType name
+      const visitIdentifier = visit?.patient?.identifiers.find((visitIdentifier) => {
+        if (visitIdentifier?.identifierType?.name === configIdentifier?.identifierName) {
+          return true;
+        }
+        return false;
+      });
```
ibacher, line 77:
```suggestion
      const visitIdentifier = visit?.patient?.identifiers.find((visitIdentifier) =>
        visitIdentifier?.identifierType?.name === configIdentifier?.identifierName);
```

packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
```diff
@@ -37,18 +39,89 @@ export function useActiveVisits() {
...
+
+      if (personAttributes) {
+        //add the new attribute or rewrite existing one to activeVisit object
+        //the parameter will corresponde to the name of the key value of the configuration
+        //and the respective value is the persons value
+        activeVisits = {
+          ...activeVisits,
+          [header?.key]: personAttributes?.value,
+        };
```
ibacher, line 113:
```suggestion
        activeVisits[header?.key]: personAttributes.value;
```

packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx
```diff
@@ -37,18 +39,89 @@ export function useActiveVisits() {
...
+
+      if (visitIdentifier) {
+        //add the new identifier or rewrite existing one to activeVisit object
+        //the parameter will corresponde to the name of the key value of the configuration
+        //and the respective value is the visit identifier
+
+        activeVisits = {
+          ...activeVisits,
+          [configIdentifier?.header?.key]: visitIdentifier?.identifier,
+        };
```
ibacher, line 87:
```suggestion
        activeVisits[configIdentifier.header?.key]: visitIdentifier.identifier;
```

ibacher: Heck, this whole `if` could be:

```suggestion
activeVisits[configIdentifier.header?.key]: visitIdentifier?.identifier ?? '--';
```

PR #496, opened 2023-01-06
(fix) Fix spelling mistake in registration name input field

gracepotma:
Fixes a spelling mistake in the `warnText` string displayed when filling in name fields in the patient registration form. Changes the text from "Number is name is dubious" to "Number in name is dubious".

## Screenshot

> Showing updated warning text
image

### DIFF
--- a/packages/esm-patient-registration-app/translations/en.json
+++ b/packages/esm-patient-registration-app/translations/en.json
@@ -58,7 +58,7 @@
   "negativeYears": "Negative years",
   "no": "No",
   "noResultsFound": "No results found",
-  "numberInNameDubious": "Number is name is dubious",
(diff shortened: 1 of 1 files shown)


gracepotma:
I have to admit "dubious" isn't a very clear word. @denniskigen what would you suggest here? I'm blanking. But such a small detail probably not worth too much time/thought. RefApp 2.x doesn't even have a number checker for the name field.

denniskigen:
I'd have gone with something like: `Names typically don't contain numbers`. 

Here's a rubric for why this kind of check might be misguided https://shinesolutions.com/2018/01/08/falsehoods-programmers-believe-about-names-with-examples/.

denniskigen (approved):

PR #498, opened 2023-01-12
(test) Fix tests failing due to inconsistent whitespace

denniskigen:
## Requirements

## Summary

Following on from https://github.com/openmrs/openmrs-esm-patient-chart/pull/934, this PR fixes an unusual regression on my system involving the formatDateTime helper function. More specifically, an assertion that checked for the existence of whitespace between a timestamp and the meridian indicator (i.e. AM / PM) fails to recognize literal whitespace and expects non-breaking whitespace instead. 

It's entirely likely that this regression is due to some weird quirk on my setup but nonetheless, the fix I've provided is compatible with setups experiencing this issue as well as those that are not.

## Screenshots

> Test failure

Screenshot 2023-01-12 at 14 03 10

> A row with the expected content clearly exists:

Screenshot 2023-01-12 at 14 03 37

### DIFF
--- a/packages/esm-outpatient-app/src/queue-patient-linelists/queue-linelist-base-table.test.tsx
+++ b/packages/esm-outpatient-app/src/queue-patient-linelists/queue-linelist-base-table.test.tsx
@@ -87,10 +87,10 @@ describe('QueuePatientBaseTable: ', () => {
     });
 
     const expectedTableRows = [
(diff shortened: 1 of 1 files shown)


denniskigen:
@CynthiaKamau @vasharma05 can you replicate this regression on your setups?

hadijahkyampeire (approved):
Thanks @denniskigen

PR #511, opened 2023-01-24
(feat) Integrate support for a Client Registry(MPI) within Registration and Search workflows

samuelmale:
## Requirements

- [X] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [X] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [X] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary
Integrates support of Master Patient Index synchronisation within:

#### Registration 
     - Register a new patient record from an external MPI record.
     - New registrations are synchronised with the configured MPI. (The backend CR module does most of the heavy lifting)
#### Search
     - Searches against the MPI. As an MVP, the search is only identifier based.

## Attachments
ezgif-1-97a01d531f

*In this demo, the underlying backend is configured with an early implementation of the CR module and it uses this MPI: https://namibia-sante-mpi.globalhealthapp.net/#!/dashboard*

## Related Issue

*TBD*

## Other

*None.*


packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -31,4 +31,30 @@ export const configSchema = {
...
+      _type: Type.String,
+      _default: 'ac79a22f-4783-498d-80d8-d0523f6f91d2',
+    },
+    primaryOmrsIdentifierType: {
+      _type: Type.String,
+      _default: '05a29f94-c0ed-11e2-94be-8c13b969e334',
+    },
+    disableSearch: {
+      _type: Type.Boolean,
+      _default: true,
```
samuelmale, line 57: Disabled by default because the backend support isn't ready.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -306,6 +312,24 @@ export const esmPatientRegistrationSchema = {
       _default: '736e8771-e501-4615-bfa7-570c03f4bef5',
     },
   },
+  MPI: {
+    title: {
```
ibacher, line 316: If possible, any display strings should **not** be part of the configuration. Just use a reasonable default and use the translation pipeline to translate things. This allows us to support multiple languages while still retaining the ability for implementations to change this to whatever they want.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -306,6 +312,24 @@ export const esmPatientRegistrationSchema = {
...
+      _type: Type.String,
+      _default: 'MPI',
+    },
+    baseAPIPath: {
+      _type: Type.String,
+      _default: '/ws/fhir2/R4/MPIPatient',
```
ibacher, line 322: This is a big red-flag for me...

1. I'm not sure we really need this as a configuration setting. This app should just work with a defined API endpoint. I don't see why an implementation would change this right now.
2. Adding a custom domain like this to the FHIR module is a non-starter. We should discuss what actual queries are necessary here and how we can properly represent those.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -306,6 +312,24 @@ export const esmPatientRegistrationSchema = {
...
+    },
+    preferredPatientIdentifierTitle: {
+      _type: Type.String,
+      _default: 'Health ID',
+    },
+    preferredPatientIdentifierType: {
```
ibacher, line 328: I think that this type of configuration probably belongs as part of the backend settings, not as part of the frontend configuration.

ibacher (commented):
Some first-pass comments on this. It would probably be best to split this into multiple PRs, e.g., the search stuff and the registration stuff. In all honesty, I'm happier with the search part right now than the registration part.

packages/esm-patient-registration-app/src/config-schema.ts
samuelmale: The inspiration of having this config came from the fact that we have mixed terminologies in-regards to the title of the registry. Some folks prefer calling it a "Client Registry"; others simply an MPI. I believe it's imperative to keep this option configurable. cc: @pmanko @ebambo

packages/esm-patient-registration-app/src/config-schema.ts
samuelmale: This config was necessary before we agreed on a generic backend module. I'm happy to drop it. @ibacher did you have a look at: https://issues.openmrs.org/browse/HIE-1?

packages/esm-patient-registration-app/src/config-schema.ts
samuelmale: Nice one. I believe there could be an existing GP; guess I was too dumb to think about it.

packages/esm-patient-registration-app/src/config-schema.ts
ibacher: @samuelmale I'm not saying it shouldn't be configurable. I'm just saying it needs to be localisable. If we make this a configuration property, implementors can only ever supply a single string regardless of the number of languages they support. If we use the standard i18n mechanism, we can support both localising it into other languages and having implementation-specific verbiage. This is because translation keys can be provided as part of the application configuration so if I want it to say "Client Registry", I can setup the configuration for this app using [this feature](https://github.com/openmrs/openmrs-esm-core/pull/546).

You're right that this needs to be configurable, but making it part of the config schema like this actually makes it less customisable than it ought to be.

samuelmale: Gotcha!! That was detailed @ibacher Thanks!

packages/esm-patient-registration-app/src/config-schema.ts
smallgod: @pmanko fyi. Kindly add Pawel to this thread. thanks

packages/esm-patient-registration-app/src/config-schema.ts
samuelmale: Question: How can we best represent these in the backend? 

Notes:
We basically want to be able to map identifiers basing on 2 or 3 orientations: 'internal-preferred' | 'internal' | 'external'. 

1. 'internal-preferred': represents the primary identifier: Currently configurable through a metadata mapping: `metadatamapping/termmapping?code=emr.primaryIdentifierType` (I'm guessing it's controlled by some GP) 
2. 'internal': represents extra or secondary OpenMRS identifiers
3. 'external': represents the identifier assigned to the patient by the MPI

@ibacher Would it be ideal to configure the 'external' identifier-type through a Global Property declared by the CR module?

ibacher: > I'm guessing it's controlled by some GP

It's actually controlled by a metadata mapping.

I don't love using a GP for this, because it's fundamentally not global, i.e., the external identifier (presumably) varies with the MPI, and using a GP for this kind of restricts us to a MPI-per-system. I'd tend to prefer just having a table to store that configuration data...

samuelmale: This means we would need an extra endpoint to fetch this config? Should we setup a technical call with the folks working on the CR module? Btw, it would be great if you added some technical notes to the epic => https://issues.openmrs.org/browse/HIE-1

cc: @smallgod, @pmanko

ibacher: Might be good to have a call about this. I'm not 100% sure that we actually need the frontend to know about much of this, i.e., it's not directly querying the MPI, but rather a backend operation that queries the MPI, so we can mark the MPI-provided identifier in that response, if it's necessary to differentiate them. Basically, I'm not quite sure I understand why we need this distinction in the frontend, I'm sure I'm missing something.

samuelmale: The front-end needs to be able to classify records into:
1. Internal (Not synchronised with the MPI)
2. External (Record from the MPI and not synchronised with the EMR)
3. Cross cutting (Records synchronised with both systems)

It achieves the above requirements through patient identifications

ibacher: But it should be the backend directly communicating with the MPI, right?

samuelmale: Yes, its the backend in communication with the MPI

PR #524, opened 2023-01-31
(fix) O3-1832: Fix issue where a new patient wouldn't get registered

vasharma05:
## Requirements

## Summary

This PR fixes a weird edge case with the registration form where the form would not register a user successfully. More specifically, if a user:
- Launches the registration form and then fills in the required fields. 
- Configures an identifier by clicking `Configure identifier` and then selects an identifier (say `ID Card`, for example).
- Types in a value for the identifier in the ID Card field.
- Clears the identifier value that got typed in.
- Attempts to save the form by clicking `Register Patient`.

They would end up getting the following error:

identifiers-error

The reason for this error is the way the identifiers property of the request payload gets constructed:

patient-payload

Note how the second object in the identifiers array is `null` when it should be an object with the following shape:

```ts
{
  identifier: string;
  identifierType: string;
  location: string;
  preferred: boolean;
}
```

This PR fixes this issue by ensuring that empty identifiers don't get included in the patient creation payload.

## Related Issue
https://issues.openmrs.org/browse/O3-1832

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
@@ -224,25 +224,29 @@ export class FormManager {
     location: string,
     abortController: AbortController,
   ): Promise<Array<PatientIdentifier>> {
-    let identifierTypeRequests = Object.values(patientIdentifiers).map(async (patientIdentifier) => {
-      const {
-        identifierTypeUuid,
-        identifierValue,
-        identifierUuid,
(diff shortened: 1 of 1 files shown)


rbuisson:
Thanks @vasharma05  for looking into this.

Let's wait for @denniskigen 's review 👍🏻

denniskigen (approved):
Very handy fix. 

Thanks for reporting this, @hadijahkyampeire. And thank you, @vasharma05, for the fix!

PR #525, opened 2023-01-31
(enhancement): Add ability to use configurable labels on address hierarchy inputs

donaldkibet:
## Requirements

## Summary

- Add the ability to use config provided label on the Address hierarchy as shown in the screenshot

## Screenshots
Screenshot 2023-01-31 at 4 04 11 PM

## Related Issue

*None.*

## Other

*None.*


vasharma05:
Hey @donaldkibet , these labels are coming from the address templates itself, instead of being hardcoded. I believe we can change the labels in the address templates for the required changes.

The "optional" text is according to the designs to clearly mark which fields are optional and which are not. I believe this doesn't need to be done on the frontend.

PR #541, opened 2023-02-15
(fix) O3-1839: Address field resetting values on patient registration form

vasharma05:
## Requirements

## Summary
This PR fixes the resetting of patient registration form when address fields are loaded. 

The fix is to remove setting the address field's initial values when address fields are loaded. Address fields are fetched from the backend and hence it takes time to come on the form.

Setting the initial value is taken care by the address component itself, and there's no requirement of setting the fields in the patient registration component.

## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/O3-1839

## Other

*None.*


jayasanka-sack (approved):
Thanks, @vasharma05 !

PR #544, opened 2023-02-16
(test) O3-1873: Configure JUnit reporting for E2E tests

jayasanka-sack:
## Requirements

## Summary
This PR addresses the Jira ticket [O3-1873](https://issues.openmrs.org/browse/O3-1873) by adding JUnit reporting configuration to the e2e tests in order to display the results on Bamboo and providing more comprehensive information about the tests. Previously, the e2e tests were only configured to produce HTML reports, which limited our ability to analyze and track the test results on bamboo.

With this change, the e2e tests will now generate both JUnit and HTML reports on CI, which will allow us to display the results on Bamboo and download the HTML reports as attachments for further analysis. This will enable us to better understand the test results, identify any issues, and take appropriate action to improve our testing process.

Apart form that, I updated environment variables as well. I've noticed that some people are confused about needing to specify different hosts for WS and UI URLs, when in fact they should be the same. To simplify things, I updated the code to use a single environment variable, to specify the base URL for both UI and API requests. This helps to avoid any confusion and makes it easier for everyone to use.

## Screenshots
A preview of the outcome (Generated using playwright CLI). 
Screenshot 2023-02-16 at 12 01 15
Screenshot 2023-02-16 at 12 01 45

## Related Issue

https://issues.openmrs.org/browse/O3-1873

## Other

Learn about playwright reporters: https://playwright.dev/docs/test-reporters

### DIFF
--- a/e2e/core/global-setup.ts
+++ b/e2e/core/global-setup.ts
@@ -15,7 +15,7 @@ async function globalSetup() {
   const token = Buffer.from(`${process.env.E2E_USER_ADMIN_USERNAME}:${process.env.E2E_USER_ADMIN_PASSWORD}`).toString(
     'base64',
   );
-  await requestContext.post(`${process.env.E2E_WS_BASE_URL}rest/v1/session`, {
+  await requestContext.post(`${process.env.E2E_BASE_URL}/ws/rest/v1/session`, {
     data: {
       sessionLocation: process.env.E2E_LOGIN_DEFAULT_LOCATION_UUID,
       locale: 'en',
--- a/.env
+++ b/.env
@@ -1,8 +1,7 @@
 # This is an example environment file for configuring dynamic values.
 # This can be removed or changed with better solutions in the future.
 # It should *only* be kept as long as there are no sensitive values in here.
-E2E_UI_BASE_URL=http://localhost:8080/openmrs/spa/
-E2E_WS_BASE_URL=http://localhost:8080/openmrs/ws/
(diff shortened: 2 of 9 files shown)


e2e/core/global-setup.ts
```diff
@@ -15,7 +15,7 @@ async function globalSetup() {
   const token = Buffer.from(`${process.env.E2E_USER_ADMIN_USERNAME}:${process.env.E2E_USER_ADMIN_PASSWORD}`).toString(
     'base64',
   );
-  await requestContext.post(`${process.env.E2E_WS_BASE_URL}/rest/v1/session`, {
+  await requestContext.post(`${process.env.E2E_WS_BASE_URL}rest/v1/session`, {
```
anjula-sack, line 18: Can't we just remove the `/` from the env variable itself? I kind a prefer it the way it is cuz it shows that /rest is the path of the api. WDYT @jayasanka-sack?

jayasanka-sack: Yes, that would be clean. The reason to configure the env variable was the base URL of the UI and the API fixture was expecting the paths with a forward slash. However, I've noticed that some people are confused about needing to specify different hosts for WS and UI URLs, when in fact they should be the same. To simplify things, I will update the code to use a single environment variable, such as HOST, to specify the base URL for both UI and API requests. This should help to avoid any confusion and make it easier for everyone to use.

anjula-sack (approved):
LGTM!

denniskigen (approved):
Thanks, @jayasanka-sack!

PR #545, opened 2023-02-16
(fix) O3-1860: Add appropriate permissions to view and edit/ configure identifiers

jona42-ui:
## Requirements

## Summary
Currently, only the system administrator is allowed to configure identifiers on the patient registration. Want to
ensure that users have appropriate permissions to view, edit, or configure patient identifiers.
Added a `UserHasAccess` component with privileges;

"Add Patient Identifiers", "Edit Patient Identifiers", and "Get Identifier Types" (removing an identifier should result in the identifier being voided, not deleted).

## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/O3-1860

## Other

*None.*


packages/esm-patient-registration-app/src/patient-registration/field/id/id-field.component.tsx
```diff
@@ -113,7 +113,13 @@ export const Identifiers: React.FC = () => {
 
   return (
     <div className={styles.halfWidthInDesktopView}>
-      <UserHasAccess privilege="coreapps.systemAdministration">
+      <UserHasAccess
```
vasharma05, line 116: Hi @jona42-ui !
Actually the User has access needs to wrap the button below only, and not the whole header. The previous wrap is not correct.

If there is not sufficient privelege, it'll hide the `Identifiers` heading as well from the form, but we only need to hide the `Configure` button.

vasharma05 (changes requested):

vasharma05:
This is the place to put the access for editing an identifier: https://github.com/openmrs/openmrs-esm-patient-management/blob/9ae1708a69aae2e506f3a2e88103ffd5d6382c16/packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx#L114

vasharma05:
Here is the place to update the permission for deleting an identifier

https://github.com/openmrs/openmrs-esm-patient-management/blob/9ae1708a69aae2e506f3a2e88103ffd5d6382c16/packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx#L134

packages/esm-patient-registration-app/src/patient-registration/field/id/id-field.component.tsx
```diff
@@ -113,7 +113,13 @@ export const Identifiers: React.FC = () => {
...
+        privilege={[
+          'App: coreapps.addPatientIdentifierType',
+          'App: coreapps.deletePatientIdentifierType',
+          'App: coreapps.editPatientIdentifierType',
```
vasharma05, line 120: I have marked the right place for these permissions, since the overlay only allows fetching all the identifier types and adding identifier types.

Also, I believe the correct priveleges should end with `identifier` and not `identifierType`, since we are not managing the identifier-types, but the saved identifiers instead.

vasharma05: CC: @ibacher @reagan-meant @rbuisson

vasharma05: @jona42-ui , please look into this asap.

jona42-ui: thanks @vasharma05 so that means i need to also conditionallly edit the handleEdit, handleReset functions to include the 
permission levels of users. it that right?

I believe what we are achieving here is allowing users have other privileges apart from deleting and then the admin should access all other privileges(did i get this right?)

jona42-ui: what i know is that;
**privilege** prop and only renders its child components if the user has the specified privilege.
corectme here if am not right pliz

vasharma05: If you go to the patient registration form, you see a `configure` button. Clicking the Configure buttons opens an overlay.
image

As you can see, this allows adding identifiers to the patient's form. Hence the appropriate privileges for opening the configuring overlay are `coreapps.addPatientIdentifier`.

When you come to edit a patient, you can see the edit buttons and delete buttons:
image

Hence we need to place the appropriate privelege checks at appropriate places.

ibacher: So, see the preloaded privileges [here](https://dev3.openmrs.org/openmrs/admin/users/privilege.list). These just should be "Add Patient Identifiers" and "Edit Patient Identifiers" (removing an identifier should result in the identifier being voided, not deleted). I'm not sure where these other strings came from, but the idea is not to invent privileges but use the ones that are already being enforced on the backend.

jona42-ui:
updated  commit;
https://github.com/openmrs/openmrs-esm-patient-management/pull/545/commits/61a6db37cfb069d096a5cfcf62133e05c8a19639

vasharma05:
Hi @jona42-ui!
The rebasing of this ticket is not proper and hence pulled in multiple changes. Please look into this urgently.

jona42-ui:
*okay let me work on it*

jona42-ui:
hi @vasharma05 how do i reverse the rebase or i create a completely new branch?

vasharma05 (changes requested):
These are the appropriate priveleges required

vasharma05:
In the `packages/esm-patient-registration-app/src/patient-registration/field/id` the following change is required:

```
          <UserHasAccess privilege={['Get Identifier Types', 'Add Patient Identifiers']}>
            <Button
              kind="ghost"
              className={styles.setIDNumberButton}
              onClick={() => setShowIdentifierOverlay(true)}
              size={isDesktop(layout) ? 'sm' : 'md'}>
              {t('configure', 'Configure')} <ArrowRight size={16} />
            </Button>
          </UserHasAccess>
```

vasharma05 (approved):
Looks Good to me!
I hope working on this ticket must have helped you in having a better understanding of the code.
Thanks!

jona42-ui:
yah sure, thanks so much @vasharma05 for taking me by hand

gracepotma:
@jona42-ui is this ready for re-review now? This is URGENT to complete this week for the client. Please let us know **ASAP** if you need to hand this over to Vineet or Hadijah.

PR #571, opened 2023-03-01
Fixing patient list bugs ideas and suggestions

Jexsie:
## Requirements

## Summary

As per now, we are experience a couple of bugs in the Patient List app. I have created this PR for help from anyone who might have a clue on what should be done.  I have made a few inspections and checks, and got something out---in my understanding. 

- Patient list tabs doesn't switch the content:
      This is the main issue, I made changes to update the `TabType` in regard to the clicked tab, which I thought might be
      useful. 
      Making this shorter, the content in the table doesn't change probably because of [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-list-app/src/api/api-remote.ts#LL40-L85C2). The
    `getAllPatientList()`, which I think lacks some functionality to handle data querying, we are also lacking some attritubes like `isStarred`, `type` in the `custom`.   Plus the  `updatePatientList()` that is essential for the toggling of `isStarred` 
image

- Editing patient list details throws an error, yet, the details are actually edited:
      So, i think the code is fine here.  

image

 There is actually an error with code **400** after `editPatientList()` and for some reason I can't really tell,
   details are updated. 

Screenshot from 2023-02-28 23-13-07
Screenshot from 2023-02-28 23-13-35

   I went on and checked what's happening in the `editPatientList()` up to  the `postData()` and, here I can't tell what's
    wrong.

image
image
   Those are the two issues that I think are still blocking me. The rest I think I can handle
## Screenshots

*None.*

## Related Issue

https://issues.openmrs.org/browse/O3-1925
https://issues.openmrs.org/browse/O3-1927

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-list-app/src/patient-list-list/patient-list-list.component.tsx
+++ b/packages/esm-patient-list-app/src/patient-list-list/patient-list-list.component.tsx
@@ -24,12 +24,13 @@ const headersWithoutType: Array<DataTableHeader<keyof PatientList>> = [
   { key: 'isStarred', header: '' },
 ];
 
-function createLabels() {
+function createLabels(setSelectedTab: React.Dispatch<React.SetStateAction<TabTypes>>) {
   const res: Array<ReactNode> = [];
+  const handleSelectedTab = (tabType) => setSelectedTab(tabType);
 
   for (let index = 0; index < Object.keys(TabTypes).length / 2; index++) {
     res.push(
-      <Tab key={index} id={'tab-' + index}>
+      <Tab key={index} id={'tab-' + index} onClick={() => handleSelectedTab(TabTypes[Object.keys(TabTypes)[index]])}>
         {labelMap[index]}
       </Tab>,
     );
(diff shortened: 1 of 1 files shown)


Jexsie:
The bug for popping up of error message after editing a list seems to be fixed.
[Screencast from 2023-03-02 15-33-37.webm](https://user-images.githubusercontent.com/104269786/222432313-33e3b9de-2283-4691-95b5-8e41215717ee.webm)

Piumal1999:
Hi @Jexsie Thank you for taking the time to look into these issues and showing interest in helping to resolve it. The "edit patient list" issue is already fixed and It seems like most of these issues are causing due to errors in the [backend](https://github.com/openmrs/openmrs-module-cohort).

Btw, as a best practice, It would be better to create separate pull requests for each ticket/issue. Also, feel free to use [this talk thread](https://talk.openmrs.org/t/bugs-related-to-3-x-patient-lists/38688/9) to discuss your findings.

Thank you

PR #589, opened 2023-03-10
(Fix) 03-1927: Patient list tabs don't switch the content accordingly.

hadijahkyampeire:
## Requirements

## Summary

- This PR fixes the issue of patient list tabs not switching the content view accordingly, this was coming up because we were not fetching the cohortType correctly from the backend and we were not filtering on the content by `cohortType` or `isStarred`

## Screenshots

- Loom https://www.loom.com/share/16bd5419fdc9447eb23e07a84cdbe9ea

## Related Issue

- https://issues.openmrs.org/browse/O3-1927

## Other
- There is more refactoring we can do in these files but we can do that after fixing the functionality.
- The starred functionality does not work because it misses an API for it, however if there is no API we could store it in a user session or localStorage. (Just a suggestion)


denniskigen (approved):
LGTM. Thanks, @hadijahkyampeire!

PR #629, opened 2023-03-30
(fix) O3-1994 set the workflow started (base) port to port 8180 

chirandevinda:
## Requirements

## Summary
When running the e2e tests, we used to start all the esm modules in parallel. When running using yarn, it uses ports 8081-8086 for those esm modules. However, in Ubuntu 22.04 (ubuntu-latest) runner, port 8084 is used by a service called mono in default. Therefore the esm-patient-lists app fails to run on port 8084.
So as solution for this we set the workflow started (base) port to port 8180 and we run the e2e tests with yarn start --sources <whatever> --port 8180.

## Screenshots
image

*None.*

## Related Issue
https://issues.openmrs.org/browse/O3-1994
*None.*

## Other

*None.*


.github/workflows/e2e.yml
```diff
@@ -39,7 +41,7 @@ jobs:
         run: npx playwright install chromium --with-deps
 
       - name: Run dev server
-        run: yarn start --sources 'packages/esm-*-app/'&
+        run: yarn start --sources 'packages/esm-*-app/' --port 8180 &
```
Piumal1999, line 44:
```suggestion
        run: yarn start --sources 'packages/esm-*-app/' --port 8180 & # Refer to O3-1994
```

It would be better to keep a comment mentioning the reason behind this change. 

cc: @jayasanka-sack

PR #684, opened 2023-05-02
fix: Broken pagination on the advanced patient search page

jnsereko:
## Requirements

## Summary
Fixed broken pagination on the advanced patient search page 
cc @vasharma05  

## Screenshots

https://user-images.githubusercontent.com/58003327/235641120-0727f594-eb48-40b6-9681-564d011f8559.mov

## Related Issue

https://issues.openmrs.org/browse/O3-2068

## Other

*None.*

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-page/patient-search-lg.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/patient-search-lg.component.tsx
@@ -43,10 +43,6 @@ const PatientSearchComponent: React.FC<PatientSearchComponentProps> = ({
     resultsToShow,
   );
 
-  useEffect(() => {
-    goTo(1);
-  }, [query, goTo]);
-
   const handlePatientSelection = useCallback(
     (evt, patientUuid: string) => {
       evt.preventDefault();


vasharma05 (approved):
Thanks for the fix @jnsereko !

Please confirm the fix after the merge and try running all the required use cases.

vasharma05:
Hi @jnsereko !
There's a fallback in the pagination.

Use case: Search 10 and go into advanced patient search. Then go to page > 2. After going to that page, try searching for "Joshua". You'll see that it needs to revert back to 1st page. In the image below, there are 2 search results, but since the page number is the prev page page, it shows no results.

image

vasharma05:
Reverting the PR for now.

jnsereko:
> Hi @jnsereko ! There's a fallback in the pagination.
> 
> Use case: Search 10 and go into advanced patient search. Then go to page > 2. After going to that page, try searching for "Joshua". You'll see that it needs to revert back to 1st page. In the image below, there are 2 search results, but since the page number is the prev page page, it shows no results.
> 
> image

this happens because when you change to page 2, the lower and upper boundary gets pushed to 5 and 10 respectively and if the search term returns results below 10, ie 2, pagination logic will try to return the second page of the 2 results hence returning an empty list 

We just need to push the user to the first page each time the search query changes.
Unfortunately, `useEffect` hook pushes the user to the first page regardless of whether the search query changes or not.
Possible solution is to use `useRef` to tack each time the query changes and push the user to the first page.

something like 
```
  const prevQuery = useRef<string>();

  if (prevQuery.current !== query) {
    prevQuery.current = query;
    goto(1);
  }
```

PR #697, opened 2023-05-15
(feat) O3-2100: Unknown patient name and estimated D.O.B config

Jexsie:
## Requirements

## Summary
This PR introduces the allowing/ restricting of registering unidentified patients and patients with estimated date of birth using the configuration.

The new configuration points added are:
```
{
  "@openmrs/esm-patient-registration-app": {
    "fieldConfigurations": {
      "name": {
        "allowUnidentifiedPatients": true,
        ...
      },
      "dateOfBirth": {
        "allowEstimatedDateOfBirth": true
        ...
      }
    }
  }
}

```

## Screenshots
image
image

*None.*

## Related Issue
 - https://issues.openmrs.org/browse/O3-2100

## Other

*None.*


packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
```diff
@@ -68,13 +69,17 @@ export const NameField = () => {
...
         <div className={styles.nameField}>
-          <div className={styles.dobContentSwitcherLabel}>
-            <span className={styles.label01}>{t('patientNameKnown', "Patient's Name is Known?")}</span>
-          </div>
-          <ContentSwitcher className={styles.contentSwitcher} onChange={toggleNameKnown}>
-            <Switch name="known" text={t('yes', 'Yes')} />
-            <Switch name="unknown" text={t('no', 'No')} />
-          </ContentSwitcher>
+          {allowUnknownName && (
+            <>
```
mseaton, line 73: What about the edge case where "nameKnown" = false?  For example, maybe for a while the system is configured to allow uknown patients, and then the implementation wants to remove this feature?  In this case, we'd likely want to show the "Unknown/Known" section for cases where you are editing a patient that is already configured as "unknown", even if normally this is not displayed.  Maybe have this be (allowUnknownName || !nameKnown)?

Jexsie: Hey @mseaton. Taking a deep thought into this, I think this is sufficient. I created a patient with an Unknown name and an estimated date of birth then tried editing the patient details. See the screencast below. 

The constant `nameKnown` is responsible for toggling **yes** or **no**, that is when the `unidentifiedPatient` config is enabled (true). It is not part of the configuration.
[Screencast from 2023-05-15 18-59-39.webm](https://github.com/openmrs/openmrs-esm-patient-management/assets/104269786/527c3edc-63fd-417b-ac55-f3c797c1dbd5)

Jexsie: So briefly, if a patient was created as unknown before, we should show the toggle while editing.

mseaton: Yes, correct.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
```diff
@@ -19,17 +19,18 @@ function checkNumber(value: string) {
...
+  const { setCapturePhotoProps, currentPhoto, setFieldValue, initialFormValues } =
+    useContext(PatientRegistrationContext);
   const fieldConfigs = fieldConfigurations?.name;
   const [{ value: unidentified }] = useField('unidentifiedPatient');
   const nameKnown = !unidentified;
+  const showToggleUnknownName =
+    fieldConfigs?.unidentifiedPatient ||
+    (initialFormValues?.familyName === 'UNKNOWN' && initialFormValues?.givenName === 'UNKNOWN')
+      ? true
+      : false;
```
Jexsie, line 33: Since the `nameKnown` constant is used for toggling yes or no, it couldn't be used to show the toggle when editing an UNKNOWN patient. So, I had to directly use the initial form values to track if a patient is unknown or not.

This works fine for the name field since we have a way to track if a patient is known or not. But for the date of birth, I think there is hardly any way to identify whether this D.O.B was estimated or not.

mseaton: @Jexsie - both the "unknown name" and "estimated date of birth" data points are available in the patient record.  

For "unknown name", you can get an idea of how this is stored if you look at "form-manager.ts", where you can see this block of code:

```
    if (values.unidentifiedPatient) {
      attributes.push({
        // The UUID of the 'Unknown Patient' attribute-type will always be static across all implementations of OpenMRS
        attributeType: '8b56eac7-5c76-4b9c-8c6f-1deab8d3fc47',
        value: 'true',
      });
    }
```

So, you will be able to determine if an existing patient has "unknown name" = true if they have an attribute with attributeType.uuid = `8b56eac7-5c76-4b9c-8c6f-1deab8d3fc47` and value = `true`.  This is what you need to look for, not the fact that the name fields are set as UNKNOWN, UNKNOWN.

Estimated date of birth is even more explicit in the data model.  The `patient` and `person` resources themselves have a `birthdateEstimated` property that you can check.  See the `Patient` interface defined in `patientOperations.ts` for example.

Please use these fields, which is what is actually set in the data in the back-end when these are entered in the registration form.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: Hi @mseaton!
I am working on taking this PR forward.

I have a query. According to the above scenario:
An unknown patient is registered. Then the implementation disables allowing the unknown patient. What should happen in this case if the above patient's details are open to edit. In this case, which case should we follow more precisely?

1. We show the toggle for the `Patient is Known` toggle? In this case 2 scenarios arise:

1.a. If the user (nurse/ provider) **doesn't** toggle the Patient is Known -> Yes, should we allow saving the updates to that patient?
1.b. If the user toggles to Patient is known -> Yes, should we hide the toggle so that the user cannot set the toggle to No again, or should we allow the same?

2. We don't show the "Patient is Known" toggle and automatically set the names to empty input boxes and set the attribute value of "Unknown patient" to false ?

I will be handling the "Unknown patient" attribute in both the cases, but which user scenario to move ahead with, please let me know.
Thanks!

vasharma05: CC: @donaldkibet @rbuisson @ibacher

mseaton: @vasharma05 - if the patient being edited has a person attribute indicating that they are an unknown patient, then when we edit this patient we need to show this.  Both for completeness and accuracy of viewing the data, but also to ensure that this can be edited if desired.  Otherwise, for any patient that might have been marked as unknown historically (or maybe imported or migrated in this way) will never be able to have this attribute changed.

The same is true for the estimated birthdate field, which came up in this thread as a parallel issue.

So we just need to toggle this section as visible if a) it is configured as visible in the app OR b) if the patient is already an unknown patient based on their existing data.

ibacher: I agree with @mseaton on this.

Basically, as I understand it, the ask here is that we display the toggle:

1. If configured to do so
2. If not configured to do so, but editing a patient with an unknown name / estimated birthdate

Which seems reasonable.

A further note on this: the patient attribute is intended to be the "source of truth" for whether the patient's name is known or not. The string "UNKNOWN UNKNOWN" is a default. We actually let the implementer configure this value. The intention is to support i18n, since even patient's with an unknown name have whatever we store in the name displayed in a few places. For instance, for an implementation primarily used in a French-speaking country, "INCONNU INCONNU" is probably preferable.

Similarly, with birthdate, we have a `birthdate_estimated` column in the backend which should be used to determine whether a patient's birthdate is estimated. If you're using the REST API, this is just another boolean variable returned with the patient; if you're using the FHIR API, the birthdate will not have a day, but just a year or year and month).

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -37,7 +37,7 @@ export interface RegistrationConfig {
   fieldConfigurations: {
     name: {
       displayMiddleName: boolean;
-      unidentifiedPatient: boolean;
+      allowUnidentifiedPatients: boolean;
```
ibacher, line 40: We shouldn't rename configuration properties in `fix`-type PRs. The idea of the (fix), (feat), etc. is to allow us to judge when releasing how much of a SemVer bump we need. Changing existing configuration property names should at least be a `feat`, since it will break existing working configurations.

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
```diff
@@ -25,7 +25,8 @@ export const DobField: React.FC = () => {
   const { t } = useTranslation();
   const {
     fieldConfigurations: { dateOfBirth },
-  } = useConfig() as RegistrationConfig;
+  } = useConfig<RegistrationConfig>();
+  const allowEstimatedBirthDate = dateOfBirth?.useEstimatedDateOfBirth?.enabled;
```
ibacher, line 29:
```suggestion
  const allowEstimatedBirthDate = dateOfBirth?.useEstimatedDateOfBirth?.enabled ?? false;
```

ibacher: We should also have a check in here for if the patient's birthdate is already estimated, e.g., 

```suggestion
  const allowEstimatedBirthDate = (dateOfBirth?.useEstimatedDateOfBirth?.enabled ?? false) || (Boolean(birthdate) && dobknown);
```

Which requires moving this a bit.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
```diff
@@ -19,17 +20,25 @@ function checkNumber(value: string) {
...
-      name: { displayCapturePhoto },
+      name: {
+        displayCapturePhoto,
+        allowUnidentifiedPatients,
+        defaultUnknownGivenName,
+        defaultUnknownFamilyName,
+        displayMiddleName,
+      },
     },
   } = useConfig() as RegistrationConfig;
```
ibacher, line 35:
```suggestion
  } = useConfig<RegistrationConfig>();
```

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
```diff
@@ -19,17 +20,25 @@ function checkNumber(value: string) {
...
-  const { t } = useTranslation();
-  const { setCapturePhotoProps, currentPhoto, setFieldValue } = useContext(PatientRegistrationContext);
-  const { fieldConfigurations } = useConfig();
-  const fieldConfigs = fieldConfigurations?.name;
-  const [{ value: unidentified }] = useField('unidentifiedPatient');
-  const nameKnown = !unidentified;
```
ibacher, line 32: Don't we need some, if not all of these?

packages/esm-patient-registration-app/src/patient-registration/form-manager.test.ts
```diff
@@ -8,7 +8,6 @@ const formValues: FormValues = {
   givenName: '',
   middleName: '',
   familyName: '',
-  unidentifiedPatient: false,
```
ibacher, line 11: But this should be the default right? At least for a blank patient.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
```diff
@@ -6,6 +6,7 @@ import { Patient, Relationship, PatientIdentifier, Encounter } from './patient-r
 
 export const uuidIdentifier = '05a29f94-c0ed-11e2-94be-8c13b969e334';
 export const uuidTelephoneNumber = '14d4f066-15f5-102d-96e4-000c29c2a5d7';
+export const unidentifiedPatientAttributeTypeUuid = '8b56eac7-5c76-4b9c-8c6f-1deab8d3fc47';
```
ibacher, line 9: Seems weird to put a constant here that isn't used in this file. Just put it in the file we actually need it.

ibacher (changes requested):
A few small things I noticed reading through this.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: Thanks @mseaton for the help.

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: Thanks @ibacher !

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: I actually re-did the imports from the useConfig.

packages/esm-patient-registration-app/src/patient-registration/form-manager.test.ts
vasharma05: Actually, I'm using the attribute value as the source of truth to determine whether the patient is known or not.
I found this field redundant, as it increased the overhead of keeping the track of this above field and the attribute field

packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
vasharma05: Since this is the resource file, and this variable is being used in multiple places, hence I felt this okay to export from the resource file.

packages/esm-patient-registration-app/src/config-schema.ts
vasharma05: I feel the new name is more relevant to what it stands for. May I push this PR as a feat then?

vasharma05:
Hi @ibacher !
For the estimated birthdates, the FHIR APIs only return the estimated year of birth and not the month of birth, even though we send the estimated dob to the backend.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -200,7 +200,7 @@ export const esmPatientRegistrationSchema = {
   fieldConfigurations: {
     name: {
       displayMiddleName: { _type: Type.Boolean, _default: true },
-      unidentifiedPatient: {
+      allowUnidentifiedPatient: {
```
vasharma05, line 203:
```suggestion
      allowUnidentifiedPatients: {
```

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
vasharma05: Added the || condition in the render

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
ibacher: Ah , I see! What about the `useTranslation` hook?

ibacher:
> For the estimated birthdates, the FHIR APIs only return the estimated year of birth and not the month of birth, even though we send the estimated dob to the backend.

It’s slightly more complicated than that. For patients 5 and under, it will return the year + month. But anyone six or older it just sends the year. This is because the WHO guidelines for children are broken down by month(s) for children up to 5 years of age. Past that, it’s an estimated birthdate, so including other date parts gives false precision (technically, we always store a full date).


From a user perspective, I wouldn’t expect anyone to estimate an age as “30 years and 2 months”. The “2 months” are very unlikely to be clinically significant. However, for toddlers, infants, and newborns, 2 months can very clinically significant, so we return that data.

packages/esm-patient-registration-app/src/config-schema.ts
ibacher: Yes! I agree it’s a much better name!

packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
ibacher: This export is only used in the name-field component (unless I’m missing something). If it was used in multiple places, I could see moving it here. PS “resource” largely just means “api calls”; including constants used in the api calls but needed elsewhere as exports makes some sense.

vasharma05: Got it. I remembered the UUID was being in multiple files, but forgot I cleaned it up as well.
I'll get to it.
Thanks!

packages/esm-patient-registration-app/src/patient-registration/field/name/name-field.component.tsx
vasharma05: It's shifted to the top of the component.

ibacher:
I'm actually surprised no one had asked about that before...

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -1,4 +1,5 @@
 import { Type, validator, validators } from '@openmrs/esm-framework';
+import _default from 'yup/lib/locale';
```
ibacher, line 2: Is this used here?

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
```diff
@@ -25,7 +25,8 @@ export const DobField: React.FC = () => {
   const { t } = useTranslation();
   const {
     fieldConfigurations: { dateOfBirth },
-  } = useConfig() as RegistrationConfig;
+  } = useConfig<RegistrationConfig>();
+  const allowEstimatedBirthDate = dateOfBirth?.allowEstimatedDateOfBirth;
   const [dobUnknown] = useField('birthdateEstimated');
   const dobKnown = !dobUnknown.value;
```
ibacher, line 31: It's a little weird to create a variable which is `!dobUnknown.value` and then use it as `!dobKnown`. It might read more clearly just to have `dobUnknown.value` in the OR, no?

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -290,6 +292,11 @@ export const esmPatientRegistrationSchema = {
       },
     },
     dateOfBirth: {
+      allowEstimatedDateOfBirth: {
```
ibacher, line 295: I don't understand the difference between `allowEstimatedDateOfBirth` and `useEstimatedDateOfBirth.enabled`. Why do we have both of these?

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -363,14 +363,6 @@ export class FormManager {
       }
     }
 
-    if (values.unidentifiedPatient) {
```
ibacher, line 366: This attribute should still be set if the patient's name is unknown... That's still the case, right?

ibacher: To be clear, this attribute is supposed to be the source of truth for whether we have an anonymous patient or not. That way we're not treating the string "UNKNOWN" as a special value because we would like to be able to localise that string.

ibacher (commented):
Looks like this is in good shape to me. A few comments / questions to make sure we're adhering to requirements.

denniskigen:
@Jexsie, could you respond to the comments raised in the PR review?

packages/esm-patient-registration-app/src/config-schema.ts
vasharma05: Hi @ibacher !
The useEstimatedDateOfBirth was added by @donaldkibet with the following description

image

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
vasharma05: Yes @ibacher !
This will be set after the person attribute `Unknown patient` and will override the frontend config in case the patient is unknown.

packages/esm-patient-registration-app/src/config-schema.ts
ibacher: Right, so, maybe to rephrase, if I showed this configuration schema to an implementation team, will they be able to figure this out on their own? One quick fix would be to actually change the description of `useEstimatedDateOfBirth` so that it says "Whether to use a fixed day and month for estimated dates of birth" and related changes to the other two.

The important thing to remember is that these configuration schemas aren't primarily for developers but for implementers, i.e., this is a user interface, and we should make an effort to ensure that our user interfaces are clear and easy-to-use.

vasharma05: I agree @ibacher !
Thank you!

ibacher:
Can we resolve the e2e tests?

vasharma05:
Hi @ibacher !
I tried to fix the tests for the past 2 hours, but due to lack of knowledge, I am not able to fix it.
I request you to please let the PR be merged with the test, and I'll cover this test along with the tests related to patient management.
Thanks!

vasharma05:
This is required for Cambodia's latest server.
Can you please take the PR ahead for now?
Thanks!

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
ibacher: Can this be addressed?

ibacher:
>  Can you please take the PR ahead for now?

So, one very small comment I had initially left; once that's addressed, I'm happy to merge this in and follow-up on the e2e tests later.

vasharma05:
I've made the requested changes @ibacher !

ibacher (approved):

ibacher:
Looks like a syntax err

vasharma05:
Fixed the error @ibacher

PR #764, opened 2023-07-24
(refactor) Refactor test setup to leverage caching by `turbo`

denniskigen:
## Requirements

## Summary

This PR refactors the test setup in this monorepo to leverage the caching capabilities of `turbo`. The current setup does properly leverage `turbo` for the following reasons:

- It uses a single Jest configuration, which gets used by the `test` scripts in the root-level `package.json` to run tests across the entire monorepo. As such, all the tests get re-run regardless of whether the associated files changed.

Ideally, we would like to leverage caching when executing tests. This PR introduces exactly that by moving testing-related scripts from the root-level `package.json` to each `package.json` file of each package. Additionally, this PR also:

- Moving mocks from the root-level `__mocks__` directory to package-specific `__mocks__` directories. This step also involves moving shared mocks to the `test-helpers` file.
- Cleaning up a bunch of test warnings and adding some useful refactors.

With these changes, we can now fully leverage caching when running tests.

## Video

https://github.com/openmrs/openmrs-esm-patient-management/assets/8509731/c2da7145-7e54-4cbe-9cf3-d03a97a7115a

## Related Issue
*None*

## Other
*None*


jest.config.js
```diff
@@ -9,16 +12,17 @@ module.exports = {
   moduleNameMapper: {
     '\\.(s?css)$': 'identity-obj-proxy',
     '@openmrs/esm-framework': '@openmrs/esm-framework/mock',
-    '^lodash-es/(.*)$': 'lodash/$1',
-    '^uuid$': '<rootDir>/node_modules/uuid/dist/index.js',
     '^dexie$': require.resolve('dexie'),
+    '^lodash-es/(.*)$': 'lodash/$1',
+    '^react-i18next$': path.resolve(__dirname, 'react-i18next.js'),
+    '^uuid$': path.resolve(__dirname, 'node_modules', 'uuid', 'dist', 'index.js'),
```
vasharma05, line 18: What does path.resolve do?

packages/esm-active-visits-app/package.json
```diff
@@ -13,7 +13,10 @@
     "debug": "npm run serve",
     "build": "webpack --mode production",
     "analyze": "webpack --mode=production --env.analyze=true",
-    "lint": "eslint src --ext ts,tsx",
+    "lint": "cross-env TIMING=1 eslint src --ext ts,tsx",
+    "test": "cross-env TZ=UTC jest --config jest.config.js --verbose false --passWithNoTests",
```
vasharma05, line 17: Can't we just use `../../jest.config.js`? This actually creates duplication and becomes tiring in a breaking change.
Thanks!

vasharma05: I know we are using `require('../../jest.config.js')` but we can't we use it directly from the root?

packages/esm-appointments-app/src/admin/appointment-services/appointment-services.component.tsx
```diff
@@ -57,94 +57,104 @@ const AppointmentServices: React.FC<AppointmentServicesProps> = () => {
...
-              pattern="([\d]+:[\d]{2})"
-              name="startTime"
-              value={props.values.startTime}
-              onChange={props.handleChange}
-              labelText={t('startTime', 'Start Time')}
-              id="start-time-picker">
-              <TimePickerSelect
-                name="startTimeTimeFormat"
+            <Layer>
+              <TextInput
```
vasharma05, line 61: Is this relevant to this PR only?

vasharma05 (commented):
Everything else looks good!
Just doubtful that adding Layer is okay to add in here.
Thanks!

packages/esm-appointments-app/src/admin/appointment-services/appointment-services.component.tsx
denniskigen: As I explained on the call, yes. This clears warnings about the soon-to-be-deprecated `light` prop for Carbon components when running tests.

turbo.json
```diff
@@ -2,26 +2,13 @@
...
-    },
-    "dev": {
-      "cache": false
-    }
+    "coverage": {},
+    "extract-translations": {},
+    "lint": {},
+    "test": {},
+    "test:watch": {},
+    "typescript": {}
```
denniskigen, line 12: Looks like I wasn't using topological dependencies in pipelines correctly. I've amended these pipelines accordingly.

jest.config.js
denniskigen: Resolves the path relative to the current working directory. As discussed on the call, we shouldn't need this, and I've reverted back to using `<rootDir>`.

packages/esm-active-visits-app/package.json
denniskigen: Thanks for the spot. Makes sense!

turbo.json
ibacher: I would've thought that toplogical dependencies would be required for typescript and _maybe_ linting (since the linter might also check types)?

ibacher (approved):

turbo.json
denniskigen: Yeah, I went a bit overboard with that. I've amended things in the latest commit.

ibacher:
This looks good to me, although, I probably look more later. I'm actually ok with having a `jest.config.js` for each project... just like we have a webpack.config.js, even though all that does is re-export a library. It may not be something we need now, but if an app develops the need for a specialised jest config, we'd already have the files.

denniskigen:
The patient chart PR revealed an issue with using a single jest config at the root of the project and referencing it in all the child packages' test script as the argument to `--config` - performance slows down to a literal crawl. So, using multiple configs as is the case with the latest commits is probably the best approach.

packages/esm-active-visits-app/package.json
denniskigen: Update: Per our discussion from earlier today, we've chosen to go with multiple jest configs because of the performance issues when using a single jest config.

PR #829, opened 2023-09-29
O3-2469: Remove "would you like reminder of this appointment" question

mogoodrich:
See: https://issues.openmrs.org/browse/O3-2469

## Requirements

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
@@ -392,7 +392,7 @@ const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment, patientU
           ))}
       </Select>
 
-      <div className={styles.inputContainer} id="radio-group">
+      {/*      <div className={styles.inputContainer} id="radio-group">
         <label className="cds--label">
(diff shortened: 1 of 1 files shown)


packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
```diff
@@ -392,7 +392,7 @@ const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment, patientU
           ))}
       </Select>
 
-      <div className={styles.inputContainer} id="radio-group">
+      {/*      <div className={styles.inputContainer} id="radio-group">
```
mogoodrich, line 395: This functionality currently does nothing (the value collected is not mapped to any backing field).

An argument (perhaps a strong one) could be made for just removing this all-together, but I stuck with just commenting out  If there's no plan at all to implement this in the near future, we should just remove this.

mogoodrich: Right, that would a be *great* feature to add, to so let's just comment out the question until we get a chance to implement the functionality behind it, so users don't think patients will be getting emails when they won't be?

mseaton (approved):

PR #839, opened 2023-10-10
O3-2459: Implement Scheduled Appointment Tab

mogoodrich:
See: https://issues.openmrs.org/browse/O3-2459

## Requirements

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*


packages/esm-appointments-app/src/hooks/useAppointmentList.tsx
```diff
@@ -13,48 +14,61 @@ interface AppointmentPatientList {
...
     errorRetryCount: 2,
   });
-  const appointments = data?.data?.map((appointment) => toAppointmentObject(appointment, identifierType));
+  const appointments = data?.data?.map((appointment) => toAppointmentObject(appointment));
   return { earlyAppointmentList: (appointments as Array<any>) ?? [], isLoading, error };
 };
 
-function toAppointmentObject(appointment: AppointmentPatientList, identifierType: string) {
-  const patientIdentifier = appointment.patient.identifiers.find(
-    (identifier) => identifier.identifierName === identifierType,
```
mogoodrich, line 52: One key change here.... the Bahmni Appointments endpoint just returns a single identifier (the preferred patient identifier) while the custom endpoint returns all the identifiers and allows the identifier to display to be configured.  I've removed this functionality, we should discuss @CynthiaKamau  what the KenyaEMR needs here and how best to support it.

packages/esm-appointments-app/src/routes.json
```diff
@@ -39,6 +39,66 @@
...
+      "slot": "scheduled-appointments-panels-slot",
+      "component": "appointmentsByStatus",
+      "meta": {
+        "title": "expected",
+        "status": "Scheduled",
+        "showForPastDate": false,
+        "showForToday": true,
+        "showForFutureDate": true
+      }
+    },
```
mogoodrich, line 54: Here's where I wired in the "default" extensions to display on the page.  We can debate what the actual defaults should be, but perhaps first discuss the way I set this up for configuration.  

We basically support the following meta parameters to configure the tabs as well as the AppointmentByStatus component:
* title - title (message code) to pass into the ApointmentsByStatus widget and to display in the tab header
* status - status (one of the valid Appointmnent Statuses) to pass into the AppointmentByStatus widget
* showForPastDate: whether this tab should be displayed for past dates
* showForToday: whether this tab should be displayed for current date
* showForFutureDate: whether this tab should be displayed for future dates

packages/esm-appointments-app/src/routes.json
```diff
@@ -39,6 +39,66 @@
...
+      "name": "cancelled-appointments-panel",
+      "slot": "scheduled-appointments-panels-slot",
+      "component": "appointmentsByStatus",
+      "meta": {
+        "title": "cancelled",
+        "status": "Cancelled",
+        "showForPastDate": true,
+        "showForToday": true,
+        "showForFutureDate": true
+      }
```
mogoodrich, line 101: We should not wire in the "Early Appointments" component by default, because it requires the currently Palladium-custom "early appointments" endpoint, but you should be able to wire it in like this:

```
   {
      "name": "came-early-appointments-panel",
      "slot": "scheduled-appointments-panels-slot",
      "component": "earlyAppointments",
      "meta": {
        "title": "cameEarly",
        "showForPastDate": true,
        "showForToday": true,
        "showForFutureDate": true
      }
    }
```
... note that we don't need to set a "meta.status" here because the custom EarlyAppointments component handles getting the specific appointments required.  We really just need these meta parameters to set up the tabs properly.

packages/esm-appointments-app/translations/am.json
```diff
@@ -104,7 +104,6 @@
   "queueAddedSuccessfully": "",
   "queueEntryError": "Error adding patient to the queue",
   "reasonForChanges": "Reason For Changes",
-  "rescheduled": "Rescheduled",
```
mogoodrich, line 107: All these changed got auto-committed when I tried to push, not sure exactly how this is supposed to work, but I can clean up on merge?

packages/esm-appointments-app/src/appointments-metrics/appointments-metrics.component.tsx
```diff
@@ -18,8 +18,9 @@ const AppointmentsMetrics: React.FC<{ serviceUuid: string }> = ({ serviceUuid })
   const { currentAppointmentDate } = useAppointmentDate();
   const formattedStartDate = formatDate(parseDate(currentAppointmentDate), { mode: 'standard', time: false });
 
-  const { appointmentList: arrivedAppointments } = useAppointmentList('Honoured');
-  const { appointmentList: pendingAppointments } = useAppointmentList('Pending');
+  // TODO we will need rework these after we discuss the logic we want to use
+  const { appointmentList: arrivedAppointments } = useAppointmentList('CheckedIn');
+  const { appointmentList: pendingAppointments } = useAppointmentList('Scheduled');
```
mogoodrich, line 23: I fixed to not throw an exception since useAppointmentList needs one of the current appointment statuses, btu we will need to fix what this actually does in the future.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -1,124 +1,120 @@
...
-          <Switch name={'Honoured'} text={t('honored', 'Honored')} />
-          <Switch name={'Pending'} text={t('notArrived', 'Not arrived')} />
-          <Switch name={'CameEarly'} text={t('cameEarly', 'Came early')} />
+      {dateType === 'today' && (
+        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
+          {scheduledAppointmentPanels.map((panel) => {
+            return panel.meta.showForToday ? (
+              <Switch name={panel.meta.title} text={t(panel.meta.title, panel.meta.title)} />
+            ) : (
+              <></>
```
ibacher, line 72: Fragments are still tracked, so it's probably better to render `null` instead of an empty fragment.

```suggestion
              null
```

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -1,124 +1,120 @@
...
-        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setScheduleType(name)}>
-          <Switch name={'Scheduled'} text={t('scheduled', 'Scheduled')} />
-          <Switch name={'Honoured'} text={t('honored', 'Honored')} />
-          <Switch name={'Pending'} text={t('notArrived', 'Not arrived')} />
-          <Switch name={'CameEarly'} text={t('cameEarly', 'Came early')} />
+      {dateType === 'today' && (
+        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
+          {scheduledAppointmentPanels.map((panel) => {
+            return panel.meta.showForToday ? (
+              <Switch name={panel.meta.title} text={t(panel.meta.title, panel.meta.title)} />
```
ibacher, line 70: The default value for `t()` is always the key. We only need to use the form with the default value if it would be different than the key.

```suggestion
              <Switch name={panel.meta.title} text={t(panel.meta.title} />
```

packages/esm-appointments-app/src/appointments/scheduled/appointments-by-status.component.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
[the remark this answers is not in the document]
mseaton: It might be nice if this could take in an array of statuses.  I could see there being a use case to combine, say, everything that is either "CheckedIn" or "Completed" in in a tab named "Honored".

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
ibacher, line 70: When calling `map()` to render an array of React components, it's best to provide a key for them:

```suggestion
          {scheduledAppointmentPanels.map((panel, i) => {
            return panel.meta.showForToday ? (
              <Switch key={i} name={panel.meta.title} text={t(panel.meta.title, panel.meta.title)} />
```

This may actually be a counter-argument to my "We don't need a fragment" claim:

```ts
          {scheduledAppointmentPanels.map((panel) => {
            return panel.meta.showForToday ? (
              <Switch key={panel.meta.title} name={panel.meta.title} text={t(panel.meta.title, panel.meta.title)} />
            ) : (
              <Fragment key={panel.meta.title} />
            )
```

Also using `panel.meta.title` doesn't assume that the `scheduleAppointmentPanels` are always returned in the same order.

packages/esm-appointments-app/translations/es.json
```diff
@@ -136,7 +135,7 @@
   "unscheduled": "No programada",
   "unscheduledAppointments": "Citas no programadas",
   "unscheduledAppointments_lower": "citas no programadas",
-  "upcoming": "Próximas",
+  "upcomming": "Upcoming",
```
mseaton, line 138: I don't know where this typo is coming from, but we should fix it.  It's a typo _and_ causing the translation to get lost.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -1,124 +1,120 @@
...
+      <ExtensionSlot name={scheduledAppointmentsPanelsSlot}>
+        {(extension) =>
+          extension.meta.title === currentTab && (
+            <div className={styles.container}>
+              <Extension
+                state={{
+                  date,
+                  appointmentServiceType,
+                  status: extension.meta.status,
+                  statusText: extension.meta.title,
```
ibacher, line 111: I'm a little confused about why we have two of these? Also `meta.title` looks like it's always written in lower case?

packages/esm-appointments-app/src/hooks/useAppointmentList.tsx
mogoodrich: Hmm... it looks like there have been some recent changes to the endpoint to return all active identifiers... we can likely use this to pull out the proper identifier. I will look into ticketing this.

packages/esm-appointments-app/translations/es.json
mogoodrich: Ugh... TBH all these changes were auto-created by the commit hooks, and I just committed them assuming they were correct.  Bad call on my part... I'll revert all these changes before merging and we can look into all this separetly.

mseaton: I had the same experience, with this exact same message code when I committed recently.  I thought I fixed it in the code, but I guess not.  Best to figure out where this misspelling is originating from and fix it.  It's someone else's bug that is affecting your commit at this point.

packages/esm-appointments-app/src/hooks/useAppointmentList.tsx
CynthiaKamau: Pulling the active identifier will work for us

mogoodrich: Great!

packages/esm-appointments-app/translations/es.json
mogoodrich: Yeah, I remember you mentioned it but guess I didn't take it to heart... will clean up it, will take a small amount of work on my part because my local branch is currently a little out  sync (perhaps due to another issue)

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
mogoodrich: Makes sense about the key @ibacher , will add it, but are you saying that even with a key we can't guarantee order?  how does adding a fragment help us out, does it help mantain order even for elements that aren't displayed?  Thanks!

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
mogoodrich: I thought I had a reason for this, but it is a little ugly, will double-check and see if I can refactor it out.

packages/esm-appointments-app/src/appointments/scheduled/appointments-by-status.component.tsx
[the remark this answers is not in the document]
mogoodrich: Yep, I suspect we are going to need that, but as the endpoint doesn't yet support it, I didn't add it yet.  If we need it, we can figure out whether to modify the endpoint or handle this on the client-side.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
mogoodrich: A null caused an issue, but, actually, realizing I wasn't think modern/functional enough... this should be a filter and then a map.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
mogoodrich: Okay so status is the actual status to search on, and statusText is the display text.  "statusText" is pretty ugly name, so I changed it to "title".  It's in lower case because it's reference to message code.

mogoodrich:
@mseaton  @ibacher  ready for re-review
@CynthiaKamau  once this gets approved, I plan to merge in... let me know if you have any concerns. I'm happy to work on getting any custom config for Palladium in.

packages/esm-appointments-app/src/routes.json
```diff
@@ -39,6 +39,66 @@
...
       "component": "homeAppointments"
+    },
+    {
+      "name": "expected-appointments-panel",
+      "slot": "scheduled-appointments-panels-slot",
+      "component": "appointmentsByStatus",
+      "meta": {
+        "title": "expected",
+        "status": "Scheduled",
+        "showForPastDate": false,
```
mseaton, line 50: I'm a bit of the mind that we should always show any tabs that have appointments in them.  Maybe we _also_ want to show some slots that have nothing in them too.  But if the backend has patients that have appointments for an earlier date that are still in the "scheduled" state, we kind of need to show them, otherwise there will be no way to clean those up manually, and there wouldn't be any visibility into that.  So we should think about this in terms of how this is set up.

ibacher: I'd probably prefer that we exposed this kind of configuration via the configuration system rather than `meta`. The configuration system is implementation-configurable, `meta` is not. There is a mechanism to allow a per-extension configuration schema, if that helps with this.

mogoodrich: @ibacher not sure I follow but I think you are saying that I can't do what I want users to be able to do, and therefore I need to rework.  I'd want someone to be able to do something like this in their own config. In this example they'd be removing one of the default extensions wired into this slot, and added two of their own.

```
"@openmrs/esm-appointments-app": {
    "extensionSlots": {
      "scheduled-appointments-panels-slot": {
        "remove": [
          "checked-in-appointments-panel"
        ],
        "add": [
          {
            "name": "my-custom-tab-using-existing-component",
            "component": "appointmentsByStatus",
            "meta": {
              "title": "My Custom Name",
              "status": "Completed",
              "showForPastDate": true,
              "showForToday": true,
              "showForFutureDate": true
            }
          },
          {
            "name": "my-custom-tab-using-my-new-component",
            "component": "myNewComponent"
          }
        ]
      }
    }
  },
```
I think you are saying this won't work, but, on top of that, how would you recommend doing this?  Thanks!

ibacher: The use-case for `meta` is somewhat different from this. In particular, its for passing information about how an extension should be rendered to the _slot_ rendering it, with the sort of canonical use-case for that being having a mechanism for saying "this extension should be rendered in one column" or something like that (which is a not-fully functional thing implemented in the [patient-chart](https://github.com/openmrs/openmrs-esm-patient-chart/blob/ee7b11ecc5ac709c74e3049a9a6b1f1759d9184e/packages/esm-patient-notes-app/src/routes.json#L12)). In other words, meta is for influencing the context in which the extension is rendered. Anything about how the extension itself functions should be part of the configuration system and fetched inside the extension via `useConfig()` in the more or less normal way.

`title` is a bit of an ambiguous property, but it's used in `meta` in cases where the slot itself will do something with that title. For example, the patient-chart dashboard mechanism uses the title as the name of the link in the left-nav, which is rendered independently of the extension itself.

Concretely, even if the above syntax worked, the limitation is that I have no way of changing `showForPastDate` (for example) in the `expected-appointments-panel`, only for newly added tabs. However, if `showForPastDate` is set via the configuration system I can set it for both newly added tabs and existing tabs declared in the routes.

Also note that the `add` property on config actually can't handle all of that. It's literally just a list of names of already-registered extensions. From the canonical docs on how this is supposed to work:

```json
{
  "@openmrs/esm-patient-chart-app": {
    "extensionSlots": {
      "patient-chart-dashboard-slot": {
        "add": [
          "nav-group#HIVAdultReturn",
          "nav-group#NCDOutpatient"
        ],
        "configure": {
          "nav-group#HIVAdultReturn": {
            "title": "HIV Adult Return",
            "slotName": "hiv-nav-group-slot"
          },
          "nav-group#NCDOutpatient": {
            "title": "NCD Outpatient",
            "slotName": "heart-nav-group-slot"
          }
        }
      }
    }
  }
}
```

There's no ability to influence `meta` here (because that's the role of the configuration system).

mogoodrich: Makes sense, thanks so much @ibacher.  Is there a pointer to an example of the pattern for the per-extension config schema?

ibacher: So, we currently don't have any production code I'm aware of using the feature, but there are some [feature tests](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-config/src/module-config/module-config.test.ts#L1104), which are hopefully informative.

Basically, you'll just import `defineExtensionConfigSchema` from `@openmrs/esm-framework` and in `startupApp()` have something like:

```ts
defineExtensionConfigSchema("expected-appointments-panel", /* the config schema for this goes here */)
```

Then whenever someone provides a configuration for an `expected-appointments-panel`, they should be able to use that configuration schema like this:

```json
"@openmrs/esm-appointments-app": {
    "extensionSlots": {
      "scheduled-appointments-panels-slot": {
        "remove": [
          "checked-in-appointments-panel"
        ],
        "add": [
          "expected-appointments-panel#custom-tab"
         ],
         "configure": {
           "expected-appointments-panel#custom-tab": {
             "showForPastDate": true,
           }
         }
      }
    }
  },
```

mogoodrich: Thanks @ibacher ! I've got totally backlogged on stuff but hope to get back to this this week and will try this out!

packages/esm-appointments-app/src/index.ts
```diff
@@ -12,11 +19,42 @@ const options = {
...
+    _default: true,
+  },
+};
+
 export function startupApp() {
   const appointmentsBasePath = `${window.spaBase}/home/appointments`;
 
   defineConfigSchema(moduleName, configSchema);
 
+  defineExtensionConfigSchema('expected-appointments-panel', expectedAppointmentsPanelConfigSchema);
```
mogoodrich, line 56: So @ibacher  I added this extension config...

packages/esm-appointments-app/src/routes.json
```diff
@@ -39,6 +39,11 @@
       "name": "todays-appointments-dashboard",
       "slot": "todays-appointment-slot",
       "component": "homeAppointments"
+    },
+    {
+      "name": "expected-appointments-panel",
+      "slot": "scheduled-appointments-panels-slot",
+      "component": "appointmentsByStatus"
```
mogoodrich, line 46: here's where the extension is defined

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -1,124 +1,114 @@
...
-      {isToday && (
-        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setScheduleType(name)}>
-          <Switch name={'Scheduled'} text={t('scheduled', 'Scheduled')} />
-          <Switch name={'Honoured'} text={t('honored', 'Honored')} />
-          <Switch name={'Pending'} text={t('notArrived', 'Not arrived')} />
-          <Switch name={'CameEarly'} text={t('cameEarly', 'Came early')} />
+      {dateType === 'today' && (
+        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
+          {scheduledAppointmentPanels
+            .filter((panel) => panel.config.showForToday)
```
mogoodrich, line 69: I would expect the "default" parameters for the extension to be ability here, but the config is null.  Do I need to look up the config somehow.
I also confirmed that if I *don't* register an extension config, the config is still null... while the docs suggest it should be getting the full module config in this case?

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -1,124 +1,114 @@
...
+            })}
         </ContentSwitcher>
       )}
-      <div className={styles.container}>
-        <AppointmentsTable {...currentConfig} />
-      </div>
+
+      <ExtensionSlot name={scheduledAppointmentsPanelsSlot}>
+        {(extension) =>
+          extension.config.title === currentTab && (
```
mogoodrich, line 98: Config is null here as well...

mogoodrich (commented):
@ibacher so I tried your suggestions and it didn't work, though I suspect I'm missing something, can you take a look at the updated code (see my new comments).  Thanks!

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
ibacher: `config` isn't a property of the extension like that. Can you point me to the docs so I can fix it? This is why we normally have `useConfig()` to get the actual configuration. It looks like there's also an API-gap here if we need to use it like this (so outside of the extension itself).

mogoodrich: There was no documentation, but "config" was a property of an extension, so I figured it might work and I couldn't find any documentation of how it should work?

Is there any example of how to use useConfig with an extension?

ibacher: You know, I just didn't realise that was a property! Learn something new all the time.

Usually `useConfig()` is done inside the component rendered in the extension; I don't think there's a way to use it outside of that context, since it relies on a context provider to actually calculate the configuration.

ibacher: @mogoodrich Is there only the one extension available? I realised this needed a slightly different approach so I implemented [basically this strategy](https://github.com/openmrs/openmrs-esm-patient-chart/pull/1286#discussion_r1275392450).

Screenshot of Expected appointments tab

The issue here is that the documentation [here](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-extensions/src/store.ts#L94) is inaccurate. Specifically it says that the config will be null "until the slot is mounted", when it should say "until the _extension is rendered_", which is quite different.

ibacher: At a technical level, what seems to happen is this:

1. The `Extension` object gets rendered
2. As part of that, [it calls the internal `renderExtension()` function](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-react-utils/src/Extension.tsx#L79C15-L79C15), which [updates the internal extension store](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-extensions/src/render.ts#L40C8-L40C8) add the new instance for the extension.
3. Once the internal extension store is updated, [the `updateConfigExtensionStore()` internal function is called](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-extensions/src/store.ts#L147). This updates the `configExtensionStore` with the records of all extension _instances_.
4. The update to the `configExtensionStore` fires off [the `computeExtensionConfigs()` handler](https://github.com/openmrs/openmrs-esm-core/blob/3fb6a3ac0ce437148f59da083ee266d34fae5e53/packages/framework/esm-config/src/module-config/module-config.ts#L152) which actually computes each extensions configuration for each mounted instance (since different instances of extensions can have different configurations).

The up-shot of this is that an extension's `config` is `null` until the extension has been mounted and an additional tick or two has been run. And, thinking about it, I don't know that there's anyway to do the logic above in an improved way, since each mounted extension has basically a separate config.

ibacher:
@mogoodrich I've pushed a commit to this branch with my changes. I'm going to go through and annotate them a bit.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
   const { currentAppointmentDate: date } = useAppointmentDate();
 
   const [currentTab, setCurrentTab] = useState<string>(null);
   const [dateType, setDateType] = useState<DateType>('today');
   const scheduledAppointmentPanels = useConnectedExtensions(scheduledAppointmentsPanelsSlot);
-
-  const isDateInPast = dayjs(date).isBefore(dayjs(), 'date');
-  const isDateInFuture = dayjs(date).isAfter(dayjs(), 'date');
-  const isToday = dayjs(date).isSame(dayjs(), 'date');
+  const { allowedExtensions, showExtension, hideExtension } = useAllowedExtensions();
```
ibacher, line 35: So, first, I've added a custom hook here that basically provides a very simple map of extension names to a boolean for whether the tab associated with that extension should be rendered. This map defaults to being empty and is handled in the `ExtensionWrapper` component below.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+  const shouldShowPanel = useCallback(
```
ibacher, line 36: This is just a filtering function that's used to control which tabs get rendered.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
     } else {
       setDateType('today');
     }
   }, [date]);
 
-  // set the default tab on first render
   useEffect(() => {
-    if (!currentTab) {
-      setCurrentTab(getDefaultTab(scheduledAppointmentPanels));
+    if (currentTab === null && allowedExtensions) {
```
ibacher, line 54: This is now the logic for determining which tab should be displayed if none are currently selected. Basically, the displayed tab is whatever the first extension that's allowed for the given `dateType`.

ibacher: On reflection, I think there should be some slightly more complicated logic when the `dateType` changes to ensure that the `currentTab` value still works.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
-      {dateType === 'futureDate' && (
-        <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
-          {scheduledAppointmentPanels
-            .filter((panel) => panel.config.showForFutureDate)
-            .map((panel, i) => {
-              return <Switch key={i} name={panel.config.title} text={t(panel.config.title)} />;
-            })}
-        </ContentSwitcher>
-      )}
+      <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
```
ibacher, line 66: I collapsed the various `ContentSwitcher`s into a single `ContentSwitcher` filtered using the `shouldShowPanel()` callback defined above. I also used a key that's derived from the extension name, as that's a more stable identifier and it's always best to use the data over a simple counter if possible because it helps reduce React re-renders.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
-                  status: extension.config.status,
-                  title: extension.config.title,
-                }}
-              />
-            </div>
-          )
-        }
+        {(extension) => {
+          return (
+            <ExtensionWrapper
```
ibacher, line 75: To make this work, I had to wrap the extension in a new component, which is just responsible for:

1. Rendering the extension (visibly or invisibly) so that we get the configuration
2. Updating the `allowedExtensions` with a computation of whether the `Switcher` for this extension should be rendered.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+              hideExtensionTab={hideExtension}
+            />
+          );
+        }}
       </ExtensionSlot>
     </>
   );
 };
 
+function useAllowedExtensions() {
```
ibacher, line 91: The `useAllowedExtensions()` hook. The only thing challenging here is ensuring that we treat the underlying object as if it were immutable, so that changes to this object appropriately trigger re-renders.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+  );
+
+  return {
+    allowedExtensions,
+    showExtension: set.bind(null, true) as (extension: string) => void,
+    hideExtension: set.bind(null, false) as (extension: string) => void,
+  };
+}
+
+function ExtensionWrapper({
```
ibacher, line 108: The `ExtensionWrapper` is the heart of making this work. Again it just:

1. Renders the extension
2. Updates the `allowedExtensions` object.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+}: {
+  extension: ConnectedExtension;
+  currentTab: string;
+  appointmentServiceType: string;
+  date: string;
+  dateType: DateType;
+  showExtensionTab: (extension: string) => void;
+  hideExtensionTab: (extension: string) => void;
+}) {
+  const currentConfig = useRef(null);
```
ibacher, line 125: We track the extension config, so that we can display or hide the tab as the configuration gets updated, but we use a Ref so that we can compare the value at last render to the value at the next render and avoid unnecessary renders.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+
+  // This use effect hook controls whether the tab for this extension should render
+  useEffect(() => {
+    if (
+      currentConfig.current === null ||
+      (currentConfig.current !== null && !shallowEqual(currentConfig.current, extension.config))
+    ) {
+      currentConfig.current = extension.config;
+      shouldDisplayExtensionTab(extension?.config, dateType)
```
ibacher, line 134: This is where we determine whether the tab should be displayed. We call the function that implements that logic and then set the value in `allowedExtensions` to either `true` or `false`.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+        ? showExtensionTab(extension.name)
+        : hideExtensionTab(extension.name);
+    }
+  }, [extension, date, dateType, currentTab, showExtensionTab, hideExtensionTab]);
+
+  return (
+    <div
```
ibacher, line 141: And here, we just render the extension. Note that we check whether the current tab is the extension and that controls whether we show or hide the element.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+          appointmentServiceType,
+          status: extension.config?.status,
+          title: extension.config?.title,
+        }}
+      />
+    </div>
+  );
+}
+
+function shouldDisplayExtensionTab(config: ConfigObject | undefined, dateType: DateType): boolean {
```
ibacher, line 157: This is the function used to determine, based on the configuration and the `dateType` whether the extension tab should be rendered or not.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -16,101 +23,165 @@ interface ScheduledAppointmentsProps {
...
+    case 'futureDate':
+      return config.showForFutureDate ?? false;
+    case 'pastDate':
+      return config.showForPastDate ?? false;
+    case 'today':
+      return config.showForToday ?? false;
+  }
+}
+
+function shallowEqual(objA: object, objB: object) {
```
ibacher, line 172: I'm going to add this to `esm-utils`, but it's here for now. This just checks if two objects are "shallowly equal" to one another. This is basically the algorithm React uses internally to determine whether to re-render a component based on the component's props or state.

ibacher (commented):
Adding some detailed remarks on my commit here

mogoodrich:
@ibacher awesome, thanks for doing all this, and so quickly!  I should be able to review and test tomorrow...

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -93,19 +93,27 @@ const ScheduledAppointments: React.FC<ScheduledAppointmentsProps> = ({ visits, a
...
-      setAllowedExtensions({ ...allowedExtensions, [key]: value });
+      return { ...state, ...addedState };
     },
-    [allowedExtensions, setAllowedExtensions],
+    {},
   );
 
   return {
-    allowedExtensions,
-    showExtension: set.bind(null, true) as (extension: string) => void,
```
ibacher: I actually think the `.bind()` here on a function returned from `useCallback()` was probably the issue. However, `useReducer()` exists exactly for these cases where you have some simple state a a handful of actions that can modify the state.

packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
```diff
@@ -69,7 +69,7 @@ const ScheduledAppointments: React.FC<ScheduledAppointmentsProps> = ({ visits, a
     <>
       <ContentSwitcher className={styles.switcher} size="sm" onChange={({ name }) => setCurrentTab(name)}>
         {scheduledAppointmentPanels.filter(shouldShowPanel).map((panel) => {
-          return <Switch key={`panel-${panel.name}`} name={panel.config.title} text={t(panel.config.title)} />;
+          return <Switch key={`panel-${panel.name}`} name={panel.name} text={t(panel.config.title)} />;
```
ibacher, line 72: Using `name` (extension name) over `title` as everything else is based on extension name.

CynthiaKamau (approved):

CynthiaKamau:
Please resolve the conflicts @mogoodrich

mogoodrich (commented):
Thanks @mseaton  see my two responses... overall, I do feel this is becoming a bit complicated, but I think the best thing to do now is to get this merged in and improve/iterate on follow-on commits, as the PR comments have become rather bloated.  I've created a new ticket in our MCOE epics to look into some of the issues you raise.

mogoodrich:
Thanks @CynthiaKamau ... yes, the title is what is used for the text in the tab and can be a message code or raw text.

PR #842, opened 2023-10-12
O3-2171: Use a non-root user to run e2e tests in Bamboo

jayasanka-sack:
## Requirements

## Summary
In Bamboo CI setup for running all E2E, we've been facing issues with Playwright E2E tests execution. These tests run within a Playwright container, and the generated reports couldn't be cleaned up by the CI agent due to permission limitations. Playwright runs its tests with root user privileges, which causes problems when trying to delete files. This issue has been hindering our ability to run all E2E tests in the pipeline.

To address this problem, this pull request introduces a solution that extends the Playwright Docker image, creating a user within the container that shares the same UID and GID as the CI agent user. By doing this, we enable the CI agent to have the necessary permissions to clean up the test reports generated by Playwright.

### Changes Made
1. Created a custom Playwright Docker image that extends the original image.
1. Added a user to the custom image with the same UID and GID as the CI agent user to ensure proper file access permissions.
1. Updated backend and DB services to use prefilled images

The above changes are tested in a dummy bamboo plan: https://ci.openmrs.org/browse/REFAPP-D3E-275

## Screenshots

*None.*

## Related Issue
https://issues.openmrs.org/browse/O3-2171
*None.*

## Other

*None.*


denniskigen (approved):
Docker is hardly my forte, but it LGTM.

jayasanka-sack:
Thanks, @denniskigen !

e2e/support/bamboo/e2e-test-runner.sh
ibacher, line 1: Can we restore the `yarn install —immutable` bit? This is often necessary because the e2e tests don’t always run in a preinstalled context.

jayasanka-sack: Hey Ian,

I removed the 'yarn install --immutable' bit from the bamboo docker setup because we encountered some persistent permission issues when trying to install dependencies and run commands using yarn in a non-root user context. I explored a few different approaches to resolve this issue:

1. Copying Files and Install Dependencies During Docker Build: I attempted to copy the required files to the image and install dependencies during the Docker build process. However, I found that it couldn't install dependencies with the 'playwright' user.

2. Installing Dependencies with Root User: Another attempt involved installing dependencies with the root user during the build. Unfortunately, this caused problems because the 'playwright' user couldn't run yarn scripts later on.

Even after trying to adjust permissions and ownership, the permission errors persisted. You can easily reproduce it. 

these errors look something like:

```
Internal Error: EACCES: permission denied, open '.yarn/install-state.gz'
```
Considering that the `e2e/support/bamboo` directory is specifically dedicated to bamboo and not used elsewhere, I thought about preinstalling the dependencies within the bamboo plan and then mounting this as a volume. This approach should help us avoid the permission issues we encountered earlier.

Let's discuss this approach further and see if it's a suitable solution. Thanks for your understanding.

PR #856, opened 2023-10-23
(feat) Allow person searching by patient identifier when defining relationships

icrc-psousa:
## Summary
This PR introduces the capability to search for individuals based on patient attributes, such as the patient ID, when defining relationships within the patient registration page.
This is accomplished by first querying the patient's endpoint and then merging the results with the persons endpoint, instead of solely querying the persons endpoint. This enhancement enables the search function for patient attributes that are not originally available in the persons' endpoint.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
@@ -1,7 +1,5 @@
-import { useMemo } from 'react';
 import useSWR from 'swr';
-import useSWRImmutable from 'swr/immutable';
-import { FetchResponse, openmrsFetch, useConfig } from '@openmrs/esm-framework';
+import { openmrsFetch, useConfig } from '@openmrs/esm-framework';
 import { Patient, Relationship, PatientIdentifier, Encounter } from './patient-registration.types';
 
 export const uuidIdentifier = '05a29f94-c0ed-11e2-94be-8c13b969e334';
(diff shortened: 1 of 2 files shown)


packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.tsx
```diff
@@ -184,9 +182,23 @@ export function usePatientPhoto(patientUuid: string): UsePatientPhotoResult {
...
   const abortController = new AbortController();
 
-  return openmrsFetch(`/ws/rest/v1/person?q=${query}`, {
+  const patientsRes = await openmrsFetch(`/ws/rest/v1/patient?q=${query}`, {
     signal: abortController.signal,
   });
+
+  const personsRes = await openmrsFetch(`/ws/rest/v1/person?q=${query}`, {
+    signal: abortController.signal,
+  });
```
ibacher, line 191: Could we do this with a ` const [patientsRes, personsRes] = await Promise.all([fetch1, fetch2])`? That way we can run both fetches in actual parallel.

```
await fn();
await fn2();
```

Is semantically equivalent to:

```
Promise.resolve(fn()).then(Promise.resolve(fn2())).then(/* remainder of function */)
```

packages/esm-patient-registration-app/src/patient-registration/section/patient-relationships/relationships-section.component.tsx
```diff
@@ -64,9 +64,7 @@ const RelationshipView: React.FC<RelationshipViewProps> = ({
   );
 
   const searchPerson = async (query: string) => {
-    const abortController = new AbortController();
```
ibacher, line 67: I'd keep the abortController here and pass it as an argument to fetch person (it shouldn't be here, but that's because it should be defined further up in the component).

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.resource.tsx
```diff
@@ -21,12 +21,19 @@ export function useActiveVisits() {
...
+    const patientUUID = visit.patient?.uuid;
+    const isToday = dayjs(visit.startDatetime).isToday();
+    if (patientUUID && isToday) {
+      uniquePatientUUIDs.add(patientUUID);
+    }
+  });
 
   return {
-    activeVisitsCount: activeVisitsCount,
+    activeVisitsCount: uniquePatientUUIDs.size,
```
ibacher, line 36: Could we move this stuff into a separate PR? Doesn't seem related to the searching stuff.

packages/esm-patient-registration-app/src/routes.json
```diff
@@ -45,6 +45,13 @@
       "online": true,
       "offline": true
     },
+    {
```
ibacher, line 48: Ditto... I think this actually is a different PR?

icrc-psousa: This is already in main branch. Something went wrong when merging it here. It's fixed now!

PR #880, opened 2023-11-14
(feat) Don't remove other user properties when adding recently viewed patient

icrc-loliveira:
## Requirements

## Summary
This PR goal is to fix the unwanted behavior, when selecting a patient in the home page search box all user properties were deleted except for the `patientsVisited`.

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search.resource.tsx
+++ b/packages/esm-patient-search-app/src/patient-search.resource.tsx
@@ -57,7 +57,7 @@ export function useInfinitePatientSearch(
   return results;
 }
 
-export function useUserProperties() {
+export function useRecentlyViewedPatients() {
   const { t } = useTranslation();
   const { user } = useSession();
   const userUuid = user?.uuid;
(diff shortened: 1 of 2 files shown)


packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -57,7 +57,7 @@ export function useInfinitePatientSearch(
   return results;
 }
 
-export function useRecentlyViewedPatients() {
+export function useUserProperties() {
```
ibacher, line 60: I'm not sure why you renamed this hook? This name is somewhat less descriptive.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -138,20 +139,18 @@ export function useRESTPatients(
   return results;
 }
 
-export function updateRecentlyViewedPatients(patientUuid: string, user: LoggedInUser) {
+export function updateRecentlyViewedPatients(patientUuid: string, user: LoggedInUser, userProperties) {
```
ibacher, line 142: It would be better to make this function an export of the `useRecentlyViewedPatients()` hook, e.g., `addPatient()` or something. Then it can use the same session data and we don't need the three-step thing of ensuring that the consumer of this hook also grabs the `userProperties` export and passes it as an argument to `updateRecentlyViewedPatients()`

ibacher (changes requested):
Thanks @icrc-loliveira! A couple of suggestions for how to make this work better.

ibacher (approved):

PR #902, opened 2023-12-11
(chore) Extend ESLint configuration

denniskigen:
## Requirements

## Summary

In this PR, I've extended the ESLint configuration used in this project as follows:

- Include recommended rules from ESLint itself, Prettier, and TypeScript ESLint.
- Enable the `@typescript-eslint/consistent-type-imports` rule to enforce consistent usage of the `import type` syntax for type imports.
- Configure the `no-console` rule to throw an error for any console statement except those using warn and error.
- Add some useful TypeScript rules which are initially turned off to keep the diff in this PR manageable.

I've made some other tangential changes, including:

- Renaming extensions for resource files that don't contain JSX from `.tsx` to `ts`.
- Replacing `pretty-quick` with `lint-staged` as our pre-commit hooks library because `pretty-quick` is not compatible with the latest versions of prettier.

## Screenshots

*None*

## Related Issue

*None*

## Other

*None*


jayasanka-sack (commented):
Thanks for the PR, @denniskigen! It's so nice to see efforts toward making our codebase cleaner.

I'm curious about why we're excluding test files from these rules. Personally, I lean towards having consistent rules across both application and test files.

denniskigen:
Should be good to go, @jayasanka-sack.

ibacher (approved):

ibacher:
I'd tend to be more on board with updating the linting rules and then applying them to tests in another PR, but since it's done, it's done (as a rule, its better to have many smaller commits than a few large commits and it's better to have commits that make as small a change as possible at a time).

Thanks @denniskigen!

PR #904, opened 2023-12-14
 (feat) In Appointments App, filter out location dropdowns to only show appointment locations

chibongho:
## Requirements

## Summary

When scheduling a new appointment, we want to have the location selection dropdown be restricted to "Appointment Locations". There is already existing backend endpoint to query locations by tag name, so we just need to use that wherever we do location selection dropdown.

## Screenshots

*None.*

## Related Issue

*None.*

## Other

I did `yarn up @openmrs/esm-framework@next` to get the latest version of esm-framework, but it also upgraded a bunch of other stuff in yarn.lock. I hope that's ok.

### DIFF
--- a/packages/esm-appointments-app/src/admin/appointment-services/appointment-services.component.tsx
+++ b/packages/esm-appointments-app/src/admin/appointment-services/appointment-services.component.tsx
@@ -8,6 +8,7 @@ import { showNotification, showToast, useLocations } from '@openmrs/esm-framewor
 import type { AppointmentService } from '../../types';
 import { closeOverlay } from '../../hooks/useOverlay';
 import styles from './appointment-services.scss';
+import { appointmentLocationTagName } from '../../constants';
 
 interface AppointmentServicesProps {}
 
@@ -15,7 +16,7 @@ const AppointmentServices: React.FC<AppointmentServicesProps> = () => {
   const { t } = useTranslation();
   const { appointmentServiceInitialValue, addNewAppointmentService } = useAppointmentServices();
 
-  const locations = useLocations();
+  const locations = useLocations(appointmentLocationTagName);
   const handleSubmit = async (values: AppointmentService, helpers: FormikHelpers<AppointmentService>) => {
     const payload = {
       name: values.name,
--- a/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
@@ -28,7 +28,6 @@ import {
   Toggle,
 } from '@carbon/react';
 import {
-  type ConfigObject,
   ExtensionSlot,
   showNotification,
   showToast,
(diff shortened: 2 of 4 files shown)


yarn.lock
```diff
@@ -1321,6 +1321,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@babel/runtime@npm:^7.14.8":
```
mogoodrich, line 1324: @ibacher  @denniskigen  what is the protocol for updating yarn.lock files?  I assume this is okay?

mogoodrich (commented):
Defer to Ian and Dennis on the yarn.lock file stuff, but generally LGTM...

@CynthiaKamau  this is breaking  change, in that after this commit it will only show locations in the location dropdown that have the "Appointment Location" tag (which is set up by Bahmni Appointments).  This seems like a reasonable change, but wanted to make sure it works for your team.

yarn.lock
chibongho: Looks like the `yarn.lock` changes are causing some of the CI checks to fail. I only really need to change `esm-framework` to grab the most recent `next` version. Any advice?

CynthiaKamau:
> Defer to Ian and Dennis on the yarn.lock file stuff, but generally LGTM...
> 
> @CynthiaKamau this is breaking change, in that after this commit it will only show locations in the location dropdown that have the "Appointment Location" tag (which is set up by Bahmni Appointments). This seems like a reasonable change, but wanted to make sure it works for your team.

@donaldkibet @makombe

CynthiaKamau:
Thanks @chibongho , kindly fix the conflicts

ibacher:
@chibongho My best guess for how we got here is you ran: `yarn up @openmrs/esm-framework`, but then discarded the changes to package.json. This creates a situation where the package.json file has `"@openmrs/esm-framework": "next"` as a declared dependency, but the Yarn lock file only has a resolution like this:

```
"@openmrs/esm-framework@npm:5.3.1-pre.1182":
  version: 5.3.1-pre.1182
  resolution: "@openmrs/esm-framework@npm:5.3.1-pre.1182"
```

I.e., it's resolving for a rule in the package.json that says `"@openmrs/esm-framework": "5.3.1-pre.1182"`. In this scenario, you need to run `yarn` or `yarn install` again to get things right. So the ideal command to update here would be something like `yarn up @openmrs/esm-framework@next openmrs@next && git restore package.json && yarn install` (note when updating `@openmrs/esm-framework` it is a good idea to update `openmrs` at the same time so that any updates that have been made to the default Webpack configuration are also taken into account).

To fix the current state you're in, I would run `yarn up openmrs@next && git restore package.json && yarn install` and commit the results (this should _only_ update the `yarn.lock` file).

denniskigen:
That's right, @ibacher. And it's likely my advice that led Chi astray.

chibongho:
Thanks @ibacher,  that fixed the yarn.lock issues.

mseaton (approved):
Looks good to me - pretty straightforward - thanks @chibongho !

PR #912, opened 2023-12-24
(fix) Phone field should be defined as a built-in field

brandones:
## Requirements

## Summary

The phone field was defined as a default in the fieldDefinitions config. This is wrong; this makes it so that an implementer who provides a custom field definition will wipe out the phone field.

Also it seems that the `phone & email` field is no more, so I got rid of that.

## Related Issue

Discovered while reproing https://openmrs.atlassian.net/issues/O3-1425

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/field/field.test.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/field/field.test.tsx
@@ -288,9 +288,7 @@ describe('Field', () => {
       error = err;
     }
 
-    expect(error).toBe(
-      "Invalid field name 'invalidField'. Valid options are 'weight', 'name', 'gender', 'dob', 'address', 'id', 'phone & email'.",
-    );
+    expect(error).toMatch(/Invalid field name 'invalidField'. Valid options are /);
     expect(screen.queryByTestId('invalid-field')).not.toBeInTheDocument();
 
     consoleError.mockRestore();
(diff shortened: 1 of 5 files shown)


brandones:
Ohhhh I see why this happened now. https://github.com/openmrs/openmrs-esm-patient-management/pull/612 Probably the right thing to do is to add a `fieldConfiguration`

brandones:
Also is there supposed to be a built-in email field?

packages/esm-patient-registration-app/src/patient-registration/field/field.test.tsx
```diff
@@ -288,9 +288,7 @@ describe('Field', () => {
       error = err;
     }
 
-    expect(error).toBe(
-      "Invalid field name 'invalidField'. Valid options are 'weight', 'name', 'gender', 'dob', 'address', 'id', 'phone & email'.",
-    );
+    expect(error).toMatch(/Invalid field name 'invalidField'. Valid options are /);
```
vasharma05, line 291: The error message is incomplete.

vasharma05:
Just a simple drawback, this will make the application configuration bit stiff, if such more specific fields are added in the future, otherwise all good!

packages/esm-patient-registration-app/src/patient-registration/field/field.test.tsx
brandones: It's intentional. With the list, then every time the built-ins change the test needs updating. Without, it fails to test against the possibility that something goes wrong with listing the built-ins. It's a minor preference, but I think the risk of the latter is worth the convenience of avoiding the former.

brandones:
Ok, thanks @vasharma05 !

PR #288, opened 2022-08-02
(enhancement):Add ability to custom concept answers for coded registr…

makombe:
…ation obs

## Requirements

- [ ] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [ ] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [ ] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

PR #914, opened 2023-12-24
(fix) Coded person attribute field should report error if not configured correctly

brandones:
## Requirements

## Summary

I set out to fix [O3-1425: Coded patient attribute fields don't submit data](https://openmrs.atlassian.net/browse/O3-1425), but found it had been duplicated and maybe fixed by @vasharma05 in [O3-1918](https://openmrs.atlassian.net/browse/O3-1918)? Though I actually still can't get coded patient attributes to save locally. So I might file the same bug yet again once I document it well.

What I have done here is just
- Ensure consistent behavior with customConceptAnswers
- Make the coded person attribute field complain with an error if it is misconfigured
- Not allow the coded person attribute field to render as a textbox when misconfigured
- Clean up tests (including some tests that were written to verify that things were behaving incorrectly...) and write a few more

## Screenshots

Old behavior
reg-ptatt-bad

New behavior
reg-platt-good
reg-platt-err1
reg-platt-err2
reg-platt-err3

## Other

I built this branch off of https://github.com/openmrs/openmrs-esm-patient-management/pull/912 and then rebased to main; it might not work until that PR is in.


packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -57,12 +57,13 @@ export function ObsField({ fieldDefinition }: ObsFieldProps) {
...
           label={fieldDefinition.label}
           required={fieldDefinition.validation.required}
+          customConceptAnswers={fieldDefinition.customConceptAnswers}
         />
       );
     default:
       return (
         <InlineNotification kind="error" title="Error">
-          Concept has unknown datatype "{concept.datatype.display}"
+          Concept for obs field "{fieldDefinition.id}" has unknown datatype "{concept.datatype.display}"
```
vasharma05, line 66: Hi @brandones !
This text needs to be translated. Can you please add the translation?
Thanks!

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -146,40 +145,32 @@ interface CodedObsFieldProps {
...
 
-  const fieldName = `obs.${concept.uuid}`;
-  const fieldDefinition = config?.fieldDefinitions?.filter((def) => def.type === 'obs' && def.uuid === concept.uuid)[0];
+  const answers = useMemo(
+    () =>
+      customConceptAnswers.length
+        ? customConceptAnswers
+        : isLoadingConceptAnswers
+          ? []
+          : conceptAnswers.map((answer) => ({ ...answer, label: answer.display })),
```
vasharma05, line 165: Now that I think of it, providing a custom-concept answer shouldn't be allowed.
If someone wants to make any change to the answers, that should be the backend, having a single source of truth, but I think this might become a breaking change.
Thanks!
CC: @ibacher

packages/esm-patient-registration-app/src/patient-registration/field/person-attributes/coded-person-attribute-field.component.tsx
```diff
@@ -1,33 +1,77 @@
...
+    customConceptAnswers.length ? '' : answerConceptSetUuid,
+  );
   const { t } = useTranslation();
   const fieldName = `attributes.${personAttributeType.uuid}`;
+  const [error, setError] = useState(false);
+
+  useEffect(() => {
+    if (!answerConceptSetUuid) {
+      reportError(
+        `The person attribute field '${id}' is of type 'coded' but has been defined without an answer concept set UUID. The 'answerConceptSetUuid' key is required.`,
```
vasharma05, line 36: Translation required

packages/esm-patient-registration-app/src/patient-registration/field/person-attributes/coded-person-attribute-field.component.tsx
```diff
@@ -1,33 +1,77 @@
...
+      );
+      setError(true);
+    }
+  }, [answerConceptSetUuid]);
+
+  useEffect(() => {
+    if (!isLoadingConceptAnswers) {
+      if (!conceptAnswers) {
+        reportError(
+          `The coded person attribute field '${id}' has been defined with an invalid answer concept set UUID '${answerConceptSetUuid}'.`,
```
vasharma05, line 46: Translation required

packages/esm-patient-registration-app/src/patient-registration/field/person-attributes/coded-person-attribute-field.component.tsx
```diff
@@ -1,33 +1,77 @@
...
+        );
+        setError(true);
+      }
+      if (conceptAnswers?.length == 0) {
+        reportError(
+          `The coded person attribute field '${id}' has been defined with an answer concept set UUID '${answerConceptSetUuid}' that does not have any concept answers.`,
```
vasharma05, line 52: Translation required

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
brandones: This was added by @makombe in https://github.com/openmrs/openmrs-esm-patient-management/pull/288, reviewed by @donaldkibet . I agree that it is probably bad, but PIH country implementers probably also will need it, as they have great difficulty changing the concepts in their servers (at least as of 2022), CC @mseaton 

Nevertheless this debate should not hold up this PR. But thank you for flagging it @vasharma05

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
brandones: Do we translate error messages? @ibacher @denniskigen

ibacher: We don't translate error messages on the console, but if it's displayed to the end-user, then yes, yes we do. (Since this is an inline notification I assume it's displayed). Granted, since the text here was just tweaked we can delay that to another PR.

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
ibacher: I actually broadly think that having a single set of answers for a concept is not always the right thing. What I would prefer is that `conceptConceptAnswers` refers to a custom concept set, rather than having the series of concepts as answers driven purely by frontend configuration. But this isn't really the forum for that discussion.

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
brandones: Every error that gets thrown shows up to the end-user. We don't have the norm of translating those. These are like that: hopefully, only an admin should ever see it, for it indicates that something is very badly wrong. The only difference with a thrown error is that a thrown error generally crashes down to the error boundary, while this I have considered "recoverable."

brandones: (obviously my point is not to whinge about having to add translations, but to ensure we're clear on our translation policies :stuck_out_tongue_closed_eyes: )

ibacher: You're right. The rule here is somewhat narrower, since not every error, ends up translated (though, frankly, backend error messages are also translatable, so those that aren't can be treated like bugs if they are displayed to users). 

I _do_ think that, while we haven't necessarily been consistent about it, it is good to translate these sort of error messages even if they are intended for administrators rather than end-users. We're trying to move towards a world were don't assume all administrators of the system are Anglophones (or have an English-speaking support team), which is necessary for global reach.

ibacher (approved):

PR #919, opened 2024-01-04
(fix) O3-2693 Appointment Creation Form DatePicker uses incorrect minDate format

chibongho:
## Requirements

## Summary

As of today (2024-01-03) the Appointment creation DatePicker only allows for dates after 2024-04-20
image

With this change, the minDate defaults correctly to today's date (2024-01-03)
image

The Carbon documentation is lacking on what format should be accepted for its minDate and maxDate params. From cursory trial-and-error, it seems like the format should follow the `dateFormat` passed in to the DatePicker.

https://react.carbondesignsystem.com/?path=/docs/components-datepicker--overview#datepicker-mindate

## Related Issue
This is a quick fix for this particular issue, but there are likely other similar problems elsewhere in the code base. A more complete fix should probably involve a standardized DatePicker. See:
https://openmrs.atlassian.net/browse/O3-998
https://openmrs.atlassian.net/browse/O3-1991

### DIFF
--- a/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
@@ -90,7 +90,7 @@ const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment, patientU
   );
 
   const appointmentService = services?.find(({ uuid }) => uuid === patientAppointment.serviceUuid);
-  const today = dayjs().startOf('day').format();
+  const today = dayjs().startOf('day').format('DD/MM/YYYY');
 
   useEffect(() => {
     if (locations?.length && sessionUser) {


packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
```diff
@@ -90,7 +90,7 @@ const AppointmentForm: React.FC<AppointmentFormProps> = ({ appointment, patientU
   );
 
   const appointmentService = services?.find(({ uuid }) => uuid === patientAppointment.serviceUuid);
-  const today = dayjs().startOf('day').format();
+  const today = dayjs().startOf('day').format('DD/MM/YYYY');
```
chibongho, line 93: See line 245 for actual usage of this variable.

denniskigen (approved):
LGTM

packages/esm-appointments-app/src/appointments/forms/create-edit-form/appointments-form.component.tsx
mogoodrich, line 93: Adding a quick comment without knowing much about this ticket or the broader tickets, which I think are meant to deal with this broadly. but the *display* format of a date should be customizable by implementation.  For our default in O2, we tend to use a format "1-Jan-2024" , which the month specifically implemented as text so that "DD/MM/YYYY" is not misintrepreted as "MM/DD/YYYY" format by US users.  But you  may already be on top of that... :)  Again, I realize this is a quick fix just to get this working again.

vasharma05: We can directly pass the date object in the `maxDate` prop, instead of the format, which is more reliable.
CC: @ibacher @denniskigen

vasharma05: I used something similar here: https://github.com/openmrs/openmrs-esm-patient-chart/blob/f0328f86c16be660c3b06f542cd4d2a2c4fb9ae5/packages/esm-patient-chart-app/src/visit/visit-form/visit-date-time.component.tsx#L38

ibacher: Oh... yes, if the use of this variable is to populate the `maxDate` prop, we shouldn't be doing any formatting with it at all, as that will be pretty inherently non-portable.

PR #920, opened 2024-01-04
(feat) O3-2684: Remove hard coded queue entry statuses

cioan:
I've just removed the hard coded service queue entry statuses.

### DIFF
--- a/packages/esm-service-queues-app/src/active-visits/active-visits-table.test.tsx
+++ b/packages/esm-service-queues-app/src/active-visits/active-visits-table.test.tsx
@@ -114,8 +114,8 @@ describe('ActiveVisitsTable: ', () => {
     });
 
     const expectedTableRows = [
-      /Eric Test Ric Not Urgent Waiting For Triage 206 hours and 2 minutes/,
-      /John Smith Emergency Attending Clinical Consultation 206 hours and 2 minutes/,
+      /Eric Test Ric Not Urgent waiting 206 hours and 2 minutes/,
+      /John Smith Emergency in service 206 hours and 2 minutes/,
     ];
 
     expectedTableRows.forEach((row) => {
--- a/packages/esm-service-queues-app/src/active-visits/active-visits-table.test.tsx
+++ b/packages/esm-service-queues-app/src/active-visits/active-visits-table.test.tsx
@@ -114,8 +114,8 @@ describe('ActiveVisitsTable: ', () => {
     });
 
     const expectedTableRows = [
-      /Eric Test Ric Not Urgent waiting Triage 206 hours and 2 minutes/,
-      /John Smith Emergency in service Clinical Consultation 206 hours and 2 minutes/,
+      /Eric Test Ric Not Urgent Waiting - Triage 206 hours and 2 minutes/,
+      /John Smith Emergency In Service - Clinical consultation 206 hours and 2 minutes/,
     ];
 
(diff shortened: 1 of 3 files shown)


mseaton (commented):
I'm not really sure about this.  It seems more consistent with the original intention @cioan is if you were to eliminate the if/then as you have, but then return `${status} ${service}` instead of just `status`, then this would allow statuses like "Waiting For", "Attending", and "Finished" to result in text like "Waiting For Triage", "Attending Clinical Consultation" and "Finished Triage", which is more aligned with original behavior.

cioan:
Thanks @mseaton ! I have updated the PR to reflect your suggestion.

packages/esm-service-queues-app/src/active-visits/active-visits-table.test.tsx
```diff
@@ -114,8 +114,8 @@ describe('ActiveVisitsTable: ', () => {
     });
 
     const expectedTableRows = [
-      /Eric Test Ric Not Urgent Waiting For Triage 206 hours and 2 minutes/,
-      /John Smith Emergency Attending Clinical Consultation 206 hours and 2 minutes/,
```
mseaton, line 118: Still not really sure about this as a complete solution.  I agree about removing the hard-coded text, but I'd be inclined to then just split the service and status into 2 separate columns, rather than try to merge them together into a pseudo-status.

Regardless, the function that calls the buildStatusString method should no longer pass in a status that is converted to lower case, but just preserve whatever case is in the status display.

cioan: @mseaton , I am not sure what you are suggesting here. As you could see below, the tests are passing fine now after I made this change to the buildStatusString. 
Are you suggesting we change the queue display table to add an extra column to display the service? This won't be helpful at all for the MCOE EMR.

mseaton: Probably easier to discuss @cioan if it's unclear what I'm suggesting

cioan: Thanks @mseaton for clarifying all those issues. Based on our discussion, I have updated the PR with adding the "dash" between the Status and Service Name to differentiate between the two different columns. I also added two additional tickets to document the other 2 issues we discovered:
- https://openmrs.atlassian.net/browse/O3-2698
- https://openmrs.atlassian.net/browse/O3-2699

mseaton (approved):
So this generally makes sense to me, though I think the other tickets you created should be reviewed soon for what the desired behavior is.  I would like to hear from @CynthiaKamau or another from Palladium as to this change, as this will impact their system the most from what I currently understand.

cioan:
Thanks @mseaton !

packages/esm-service-queues-app/src/helpers/functions.ts
```diff
@@ -17,13 +17,7 @@ export const buildStatusString = (status: string, service: string) => {
...
   }
 
-  if (status === 'waiting') {
-    return `${status} for ${service}`;
-  } else if (status === 'in service') {
-    return `Attending ${service}`;
-  } else if (status === 'finished service') {
-    return `Finished ${service}`;
-  }
+  return `${status} - ${service}`;
```
mogoodrich, line 20: This makes sense to me, thanks all!

PR #922, opened 2024-01-08
(feat) O3-2697: Add ability to configure icons on side links (service queues)

CynthiaKamau:
## Requirements

## Summary

-Add support for links that have icons on the homepage

## Screenshots

Screenshot 2024-01-05 at 14 24 37


packages/esm-service-queues-app/src/createDashboardLink.component.tsx
```diff
@@ -22,13 +25,25 @@ function DashboardExtension({ dashboardLinkConfig }: { dashboardLinkConfig: Dash
...
 
   return (
-    <ConfigurableLink
-      className={classNames('cds--side-nav__link', {
-        'active-left-nav-link': navLink.match(name),
-      })}
-      to={`${spaBasePath}/${name}`}>
-      {t('serviceQueues', 'Service queues')}
-    </ConfigurableLink>
+    <div>
```
ibacher, line 28: Is the `div` wrapper necessary here?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -140,6 +140,11 @@ export const configSchema = {
     _default: '',
     _description: 'Custom label for patient chart button',
   },
+  showDashboardLinkIcon: {
```
ibacher, line 143: Honestly, I don't think this needs to be a configuration option. I'd just check if the link configuration includes an icon.

packages/esm-service-queues-app/src/createDashboardLink.component.tsx
```diff
@@ -2,18 +2,21 @@ import React, { useMemo } from 'react';
...
+import { ConfigurableLink, useConfig } from '@openmrs/esm-framework';
+import { Button } from '@carbon/react';
+import styles from './root.scss';
 export interface DashboardLinkConfig {
   name: string;
   title: string;
+  renderIcon: React.ComponentType<any>;
 }
 
 function DashboardExtension({ dashboardLinkConfig }: { dashboardLinkConfig: DashboardLinkConfig }) {
```
ibacher, line 14: Maybe we should turn this into a component? Almost nothing here (except the text `serviceQueues` is app-specific.

packages/esm-service-queues-app/src/createDashboardLink.component.tsx
pirupius, line 28: This wrapper isn't necessary

packages/esm-service-queues-app/src/createDashboardLink.component.tsx
```diff
@@ -22,13 +25,25 @@ function DashboardExtension({ dashboardLinkConfig }: { dashboardLinkConfig: Dash
...
+        {config.showDashboardLinkIcon ? (
+          <Button
+            className={styles.navlinkIcon}
+            hasIconOnly
+            size="sm"
+            kind="ghost"
+            renderIcon={renderIcon}
+            iconDescription="Add"
+          />
+        ) : null}{' '}
```
pirupius, line 43: This might be a stretch but i was of the view that we build this into `ConfigurableLink` component and we can pass the Icon name or component instead of adding a button component every time we need to add an icon. 

Carbon has provision for `renderIcon` within the `SideNavMenu` component ([Docs Link](https://react.carbondesignsystem.com/?path=/story/components-ui-shell-sidenav--fixed-side-nav-w-icons)) but this doesn't seem like the path we took to add the links so we could have provision for something similar like so 

```
renderIcon={Fade}
```

@ibacher @denniskigen what's your view on this?

ibacher: I think that this would somewhat confuse the purpose of `ConfigurableLink`. It is not really a component for rendering links in the side navigation menu. It is a component for rendering links in any context. I'd possibly be on-board with a component that combined a `ConfigurableLink` with a button, but that would basically be the `DashboardExtension` that I suggested above might make sense to be its own component.

CynthiaKamau: I have created this component to sort this out : https://github.com/openmrs/openmrs-esm-core/pull/885

brandones:
@CynthiaKamau @ibacher should we adapt this to use the new icon system?

brandones:
The question of a more generic extension-like thing with a link and an icon would be something we would do with the link system. Maybe one day :)

ibacher:
I _think_ this should work as-is with the new icon system... They are just React.Components at the end of the day.

packages/esm-service-queues-app/src/config-schema.ts
brandones: Please remove this config element

packages/esm-service-queues-app/src/createDashboardLink.component.tsx
brandones: @CynthiaKamau Please remove this div unless it is necessary for some reason.

brandones (changes requested):
@CynthiaKamau Please address the feedback from the previous reviewers.

brandones:
@CynthiaKamau Why is this in draft? Is it ready?

packages/esm-service-queues-app/src/createDashboardLink.component.tsx
brandones: https://openmrs.atlassian.net/browse/O3-3506

brandones:
Unless @ibacher do you think we should just close this and do this in core straight away

ibacher:
> do you think we should just close this and do this in core straight away

Might as well. I think a component like this is a sensible thing to have in core.

CynthiaKamau:
@brandones the proposal was to do this on core

PR #932, opened 2024-01-12
(fix) Fix regression in patient lists search and pagination 

denniskigen:
## Requirements

## Summary

This PR fixes regressions in the Patient Lists search and pagination functionality. Previously, the search only returned results from the first page of the Datatable, leading to an empty state when the search term was on a different page. Pagination doesn't work as well, as can be observed in the linked video.

The fix involves enabling the search to scan the entire collection of lists, ensuring that results are comprehensive and accurate. Additionally, I've updated the related test suite with new assertions that take care of various uses cases related to searching.

## Screenshots

> Before

https://github.com/openmrs/openmrs-esm-patient-management/assets/8509731/43d295ea-9282-46a4-ab9e-c5852e6c4cc1

> After

https://github.com/openmrs/openmrs-esm-patient-management/assets/8509731/27385fc4-864f-426a-b879-c88abf9c6fb5

## Related Issue
*None*

## Other
*None*


packages/esm-patient-list-management-app/src/lists-dashboard/lists-dashboard.component.tsx
```diff
@@ -61,14 +59,11 @@ const ListsDashboard: React.FC = () => {
 
   const tableHeaders = [
     { id: 1, key: 'display', header: t('listName', 'List name') },
+    { id: 2, key: 'type', header: t('listType', 'List type') },
```
denniskigen, line 62: Beyond this one [screen](https://zpl.io/VDNgdJv), I'm not sure that we should explicitly hide the list type column for any of the list tabs. Happy to change this back once get more clarity.

packages/esm-patient-list-management-app/src/lists-table/custom-pagination.component.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
```
denniskigen, line 1: This wrapper component around the Carbon [Pagination](https://react.carbondesignsystem.com/?path=/docs/components-pagination--overview) component is drawn from the Patient Chart and should probably be made more widely available elsewhere. It matches the related [designs](https://zpl.io/brGXRqX) better than the default Pagination component out of the box.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -268,17 +253,20 @@ const PatientListStarIcon: React.FC<PatientListStarIconProps> = ({ cohortUuid, i
   return (
     <TableCell className={`cds--table-column-menu ${styles.starButton}`} key={cohortUuid} style={{ cursor: 'pointer' }}>
       <Button
+        iconDescription="Star list"
```
denniskigen, line 256: The tooltip above the Star list icon button was empty because of this `iconDescription` prop was missing.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -268,17 +253,20 @@ const PatientListStarIcon: React.FC<PatientListStarIconProps> = ({ cohortUuid, i
...
         size={isTablet ? 'lg' : 'sm'}
         kind="ghost"
         hasIconOnly
         renderIcon={isStarred ? StarFilled : Star}
+        tooltipPosition="left"
```
denniskigen, line 261: It makes more sense to position the tooltip to the left to avoid it getting truncated by the right edge of the page as is the case presently.

packages/esm-patient-list-management-app/src/config-schema.ts
```diff
@@ -14,7 +14,7 @@ export const configSchema = {
   patientListsToShow: {
     _type: Type.Number,
     _description: 'The UUID of the `My List` cohort type',
-    _default: '10',
+    _default: 10,
```
denniskigen, line 17: Don't know why this value was originally provided as a string.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -57,6 +57,13 @@ interface PatientListTableProps {
   style?: CSSProperties;
 }
 
+interface PaginationProps {
```
chibongho, line 60: Do we need this? I think `usePagination` is already typed.

packages/esm-patient-list-management-app/src/config-schema.ts
ibacher: The description here is also weird? Neither 10 nor '10' is a UUID.

packages/esm-patient-list-management-app/src/lists-table/custom-pagination.component.tsx
ibacher: Yep. We should probably ticket this to add an `OpenmrsPagination` component or something similar to the styleguide.

I think the rule here is something like: if the designs call for a component in more than one app that is not achievable without custom DOM and CSS, it should be a custom component.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -1,11 +1,13 @@
 import React, { type CSSProperties, useCallback, useId, useMemo, useState, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
+import fuzzy from 'fuzzy';
+import orderBy from 'lodash-es/orderBy';
```
ibacher, line 4: The linter rule that requires this strikes me as wrong. It's correct that, for tree-shaking, you should import the function from `lodash`, but the idea of `lodash-es` is that its standard ES module, so `import { orderBy } from 'lodash-es'` should be fully tree-shakeable. Maybe I've misunderstood `lodash-es` though?

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
ibacher: User-facing strings should always be run through the translation system.

packages/esm-patient-list-management-app/src/lists-table/use-pagination-info.ts
```diff
@@ -0,0 +1,32 @@
...
+    } else if (pageSize * pageNumber > totalItems) {
+      return `${pageSize * (pageNumber - 1) + currentItems} / ${totalItems}`;
+    } else {
+      return `${pageSize * pageNumber} / ${totalItems}`;
+    }
+  }, [pageSize, totalItems, pageNumber, currentItems]);
+
+  return {
+    pageSizes,
+    itemsDisplayed: `${numberOfItemsDisplayed} ${t('items', ' items')}`,
```
ibacher, line 30: It would be better to do this as a templated string, I think, to accomodate RTL languages.

ibacher (changes requested):
Thanks Dennis! This looks good as always. A few minor comments that I think should be addressed.

packages/esm-patient-list-management-app/src/config-schema.ts
denniskigen: Copypasta strikes again

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -70,50 +77,52 @@ const ListsTable: React.FC<PatientListTableProps> = ({
...
-  const { paginated, goTo, results, currentPage } = usePagination(sortedData, pageSize);
-
-  const filteredLists = useMemo(() => {
+  const filteredLists: Array<PatientList> = useMemo(() => {
     if (!debouncedSearchTerm) {
-      return results;
+      return patientLists;
     }
 
     return debouncedSearchTerm
```
chibongho, line 100: The if statement above should make the ternary operator unnecessary.

packages/esm-patient-list-management-app/src/lists-table/use-pagination-info.ts
denniskigen: Sorry, @ibacher, how should that look like?

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -70,50 +77,52 @@ const ListsTable: React.FC<PatientListTableProps> = ({
...
+      debouncedSearchTerm
+        ? filteredLists
+        : results.map((list) => ({
+            id: list.id,
+            display: list.display,
+            description: list.description,
+            type: list.type,
+            size: list.size,
+          })) ?? [],
+    [filteredLists, results],
```
chibongho, line 125: Is this what fixed the search issue? I trust that you fixed it given the test you wrote, but I don't understand how.

denniskigen: The first key change is to use the entire 'patientLists' dataset instead of the paginated 'results' collection when filtering based on the search term. Additionally, 'tableRows' should be made aware of 'results' to respond to changes in the pagination logic. However, this is not a perfect solution as there is still an issue where if the number of matching search results exceeds 'pageSize', pagination breaks, and the number of columns returned is not in sync with the counts in the pagination component. I plan to address this issue in a subsequent PR after further review.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
denniskigen: Not sure, I'd have to look into that. It's entirely possible that the import syntax you've suggested is fully tree-shakeable, which would mean we would need to change the rule.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
denniskigen: It's important to get this right because this pattern is repeated in various places across the entire app.

packages/esm-patient-list-management-app/src/lists-table/use-pagination-info.ts
ibacher: This:

```suggestion
    itemsDisplayed: t('itemsDisplayed', '{{numberOfItemsDisplayed}} items', { numberOfItemsDisplayed }),
```

denniskigen: Does it need to be escaped? This is what gets rendered:

CleanShot 2024-01-16 at 12  06 56@2x

ibacher: `{{- numberOfItemsDisplayed}}` [based on this](https://www.i18next.com/translation-function/interpolation#unescape).

denniskigen: This syntax (where we're explicit about unescaping the interpolated value by setting the `escapeValue` property to false) ended up working for me:

`i18next.t('keyEscaped', { myVar: '', interpolation: { escapeValue: false } })`

Using the `{{- key }}` syntax did not.

ibacher (approved):

ibacher:
Thanks @denniskigen!

PR #951, opened 2024-01-23
(fix) Fix button sizes by matching esm-core's @carbon/react version

manuhabitela:
## Requirements

- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://zeroheight.com/23a080e38/p/880723-introduction).

## Summary

`@carbon/react` dependency was not the same as the one listed in esm-core packages, creating a difference between the carbon stylesheet and the carbon react components.

Problem was:

[esm-styleguide generates a stylesheet with `@carbon/react~1.37` CSS rules](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/package.json#L44). But esm-patient-management packages expect `@carbon/react^1.12` CSS rules.

One specific issue that this commit fixes is `<Button>` in this repo currently can't be sized via the `size` prop:

Setting a Button `size` prop applies `cds--btn--[size]` classes, with carbon ^1.12.  
While the esm-core CSS generates `cds--layout--size-[size]` classes with carbon ~1.37.

:warning: Since I'm totally new to OpenMRS I'm not really aware if this breaks anything. I noticed some small padding changes on Carbon content switchers for example. But I guess this would require someone else checking if nothing actually broke visually now everything is synced again.

This is what @ibacher helped me debug a few days ago on Slack (thanks!).

## Other

This would also be needed in other repos consuming the esm-styleguide generated stylesheet to prevent errors everywhere.

PS: didn't know if you prefer having yarn.lock updated in PRs or not, didn't include it, let me know.


ibacher (commented):
I'm not seeing a corresponding update to the Yarn lock file, which is necessary for any dependency updates.

ibacher:
cc: @denniskigen We probably need to do a better job of updating these dependencies when we update them in core.

ibacher (approved):
Thanks @denniskigen for fixing the tests!

denniskigen:
Yeah, Ian, I can confirm that the aria-label isn't getting passed through. It's an upstream issue with Carbon

ibacher:
@denniskigen I think I tried it, but it also doesn't work with the `ariaLabel` property right?

denniskigen:
Actually, no. Disregard what I said above. To modify the tooltip text shown on hover you'd need to modify the `iconDescription` prop:

https://github.com/openmrs/openmrs-esm-patient-management/assets/8509731/726b41c3-0ae6-4c37-b474-44df11ba03ab

denniskigen:
Thanks for working on this, @manuhabitela!

manuhabitela:
Hey guys, sorry for leaving this to you, went away a few days and I guess I had issues with my github notifications.

Thanks for continuing and merging!

I'll let someone with more knowledge (and actual time to spend on OpenMRS :grimacing:) than me do the same on the other repos if necessary, I think I won't be of much help now that the issue is targeted :)


> I'm not seeing a corresponding update to the Yarn lock file, which is necessary for any dependency updates.

Noted for the future, didn't know how you handled it as in some orgs, lock files are not included in PRs to prevent potential merging issues.

PR #953, opened 2024-01-26
(feat) O3-2749 Service Queues - Add Patient Banner to "Add patient to queue" form

chibongho:
## Requirements

## Summary

Make it so that the form to add a paitent to a queue has a `<PatientBanner />` so we can see who we are adding. This is consistent with the UI in the appointments app when adding new appointments.

Note: a lot of translation files are changed from the git pre-push hook. I only actually changed patient-search.component.tsx and appointments-header.component.tsx

cc: @ojwanganto

## Screenshots

image


mseaton (commented):
Before reviewing in more detail, it might be worthwhile to separate out the changes due to the pre-commit hooks from those related to this ticket.  Can we get the pre-commit hooks to execute on a checkout without changes, and review/commit that, and then this PR can build on that and show just what is meaningful?
@denniskigen  / @ibacher - what's up with these changes that come in from the git hooks?  What is your typical approach to dealing with these?
@chibongho - note, there is also a failing check below

ibacher:
> what's up with these changes that come in from the git hooks?

It's mostly to ensure that new translation strings are correctly added to all files and unused translation strings are removed. They're done in Git hooks mostly so it's automatic for people, though this may be the result of people by-passing those checks.

>  What is your typical approach to dealing with these?

In general, I just mark the translation files as "Viewed" in the PR list and focus on the code.

ibacher:
While this isn't per se a reason to stop this PR, I think we're at the point where we should refactor the patient header into a standard React component. Creating a number of slots all with the same name is not how the extension system is intended to be used...

mogoodrich:
Yeah, I bypassed the pre-commit hook to add the translations as well in my recent commit because it was adding a lot of noise... also, although it's great to clean these up automatically, I'm a little fearful that it may be a bit aggressive in deleting translations... if we remove a message code, potentially temporarily, we have to re-translate when we add it back in, but I guess that's the a trade off...

mogoodrich:
> While this isn't per se a reason to stop this PR, I think we're at the point where we should refactor the patient header into a standard React component. Creating a number of slots all with the same name is not how the extension system is intended to be used...

Tend to agree, but this *does* allow an implementation to add in their own custom header in a way that just having this be a component would not, correct @ibacher ?

mogoodrich (approved):
Approved, though we should determine if we want to merge in all these translation changes as part of this commit.

donaldkibet (approved):

chibongho:
Ok, thanks for the reviews. I'll go ahead and separate the translation stuff into a separate PR.

PR #959, opened 2024-01-30
(feat) Add configurability to Recently Searched Patients feature

denniskigen:
## Requirements

## Summary

This PR makes the Recently Searched Patients feature configurable by toggling it on only when the `showRecentlySearchedPatients` config property is truthy. This is a precursor to future work aimed at improving the performance of the feature. Presently, looking up 10 recently searched patient records by default when the search box is launched is not ideal from a performance perspective. Making this behaviour configurable means users get the default performant search out of the box, and then can choose to opt into the recently search patients functionality.

I've also removed use of the lodash `debounce` function, leveraging the `useDebounce` hook instead for debounce functionality to the `CompactPatientSearchComponent` component.


chibongho (approved):
LGTM

mogoodrich (approved):
Thanks @denniskigen .  I'm not super familiar with this code, so only gave a quick review, but looks good, and your reasoning in the PR is sound as well, thanks!

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-search.component.tsx
```diff
@@ -24,14 +23,17 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
   shouldNavigateToPatientSearchPage,
 }) => {
   const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
-  const showSearchResults = useMemo(() => !!searchTerm.trim(), [searchTerm]);
+  const debouncedSearchTerm = useDebounce(searchTerm);
+  const hasSearchTerm = useMemo(() => Boolean(debouncedSearchTerm.trim()), [debouncedSearchTerm]);
```
ibacher, line 27: Do we need the `useMemo()` here? It seems like `Boolean(debouncedSearchTerm.trim()` is a fast operation...

packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -7,6 +7,11 @@ export const configSchema = {
       _description: 'Where clicking a patient result takes the user. Accepts template parameter ${patientUuid}',
       _validators: [validators.isUrlWithTemplateParameters(['patientUuid'])],
     },
+    showRecentlySearchedPatients: {
+      _type: Type.Boolean,
+      _default: false,
```
ibacher, line 12: I think it's most likely that we should leave the default as `true`, since that's effectively the functionality we've been shipping since this feature was added.

ibacher (approved):
LGTM. Couple of nits.

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-search.component.tsx
denniskigen: No, we don't. Good catch.

packages/esm-patient-search-app/src/config-schema.ts
denniskigen: I'm inclined to keep this as `false`, at least until I've taken a stab at improving the performance of the Recent Patient search functionality (which I intend to do following this PR).

PR #960, opened 2024-01-31
(chore) Only auto-generate English translations

ibacher:
## Requirements

## Summary

We should only auto-generate translations for English. Translations for non-English languages should come from Transifex.

### DIFF
--- a/tools/i18next-parser.config.js
+++ b/tools/i18next-parser.config.js
@@ -42,7 +42,7 @@ module.exports = {
   lineEnding: 'lf',
   // Control the line ending. See options at https://github.com/ryanve/eol
 
-  locales: ['en', 'am', 'es', 'fr', 'km', 'he', 'ar', 'zh', 'zh_CN'],
+  locales: ['en'],
   // An array of the locales in your applications
 
   namespaceSeparator: ':',


denniskigen (approved):

befantasy:
But when will the non-English translations from Transifex be synchronized properly? Now, the Chinese translations （zh.json/zh_CN.json）I submitted on GitHub have been reverted to English. After a month of work, I finally saw some results on dev3.openmrs.org, and now it's back to the initial state.

ibacher:
@befantasy Where do you see these reversions? E.g., when [I look here](https://github.com/openmrs/openmrs-esm-patient-management/blob/f94de7d0a417fb06d3465c294c56b39565c88fba/packages/esm-appointments-app/translations/zh.json#L2), most of the strings have been translated, with only some strings (presumably those added since) not translated.

The goal here is to basically _not_ default to adding new English strings to non-English language files (previously, our setup would automatically add all translation keys and their English defaults to all of the JSON files), so that the actual process of translating strings is handled by Transifex.

befantasy:
Sorry, about the issue of reverting to English, it was my mistake. What I meant was: https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-list-management-app/translations/zh.json. Previously, this file was missing, so I submitted a PR （https://github.com/openmrs/openmrs-esm-patient-management/blob/4e7d0c6693b4f39a57d3be972014878a10fc0285/packages/esm-patient-list-management-app/translations/zh.json） I thought the PR had already been approved. Just now, I found that the strings in this file are still in English, and I thought they had been reverted. After verifying, I found that my PR was not approved, and someone else's PR added this file, but the original English strings were used. So I thought it had been reverted.

befantasy:
The most important thing now is that the resources translated on Transifex cannot be automatically synchronized to GitHub. I checked the GitHub action log (https://github.com/openmrs/openmrs-esm-patient-management/actions/runs/7701997417/job/20989301424), and the daily Scheduled Transifex Update always shows the message: "openmrs-esm-patient-management.esm-service-queues-app [zh_CN] - Local file is newer than remote, skipping."

ibacher:
> Just now, I found that the strings in this file are still in English, and I thought they had been reverted.

So, more or less the goal of this commit is to prevent what happened there from happening. It looks like something got lost somewhere along the way or something was force-pushed. Basically, with this change, those files will not be autopopulated with English strings, whereas prior to this change they would be.

> Scheduled Transifex Update always shows the message: "openmrs-esm-patient-management.esm-service-queues-app [zh_CN] - Local file is newer than remote, skipping."

Again, this is part of what we're trying to fix. AFAICT, what's been happening is that the files get updated automatically when new translation strings are added or removed (but only with English translations) and then Transifex is noticing that those files have been updated more recently than the copy on Transifex, so, even though there are more translations on Transifex, the Transifex client ends up thinking that the local translations are more up-to-date.

PR #968, opened 2024-02-06
(feature)  O3-2750 Service Queues - Configurable queues table to display queue entries by status

chibongho:
## Requirements

## Summary
A new queues table with the following feature:
- configurable columns in the queues table.
- For each status available on the selected queue, we show a tab with a table of queue entries of the selected status. (If no queue selected, then statuses from all possible queues are shown.)

This table is a work in progress, and is mounted on an alternate route (/queue-table-by-status). I intend for this table to replace the existing table in the main route (/) when it is feature-complete.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/509602/5370a011-d0c8-423d-a140-4efcfc630a5c


packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
```diff
@@ -173,19 +175,37 @@ export function usePriority() {
...
-export function useVisitQueueEntries(currServiceName: string, locationUuid: string): UseVisitQueueEntries {
+export function useUnmappedVisitQueueEntries(locationUuid: string): UseVisitQueueEntries<VisitQueueEntry> {
   const { queueLocations } = useQueueLocations();
   const queueLocationUuid = locationUuid ? locationUuid : queueLocations[0]?.id;
-  const config = useConfig();
-  const { visitQueueNumberAttributeUuid } = config;
 
   const apiUrl = `/ws/rest/v1/visit-queue-entry?location=${queueLocationUuid}&v=full`;
-  const { t } = useTranslation();
+  useTranslation();
```
ibacher, line 183: Is this here intentionally?

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
```diff
@@ -236,23 +256,15 @@ export function useVisitQueueEntries(currServiceName: string, locationUuid: stri
...
-  let mappedVisitQueueEntries;
-
-  if (!currServiceName || currServiceName == t('all', 'All')) {
-    mappedVisitQueueEntries = data?.data?.results?.map(mapVisitQueueEntryProperties);
-  } else {
-    mappedVisitQueueEntries = data?.data?.results
-      ?.map(mapVisitQueueEntryProperties)
-      .filter((data) => data.service == currServiceName);
+  const mappedVisitQueueEntries = visitQueueEntries?.map(useVisitQueueEntryProperties);
+  if (currServiceName && currServiceName !== t('all', 'All')) {
```
ibacher, line 260: On the other hand, it looks like we _need_ the translation hook here?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -138,6 +138,42 @@ export const configSchema = {
     _default: '',
     _description: 'Custom label for patient chart button',
   },
+  queueTableColumns: {
+    _type: Type.Array,
+    _description: 'Columns to appear in the queues table.',
+    _elements: {
+      id: {
```
ibacher, line 145: Do we need this property? In all of the examples, it ends up being the `headerI18nKey`...

packages/esm-service-queues-app/src/index.ts
```diff
@@ -94,6 +95,10 @@ export const addProviderToRoomModal = getAsyncLifecycle(
 
 export const addQueueEntry = getSyncLifecycle(addQueueEntryComponent, options);
 
+export const queueTableName = getSyncLifecycle(QueueTableName, options);
+export const queueTablePriority = getSyncLifecycle(QueueTablePriority, options);
+export const queueTableStatus = getSyncLifecycle(QueueTableStatus, options);
```
ibacher, line 100: Use `getAsyncLifecycle()`. `getSyncLifecycle()` should basically just be used for links.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,216 @@
...
+      const newQueueHasCurrentStatus = currentQueue?.allowedStatuses?.some((s) => s?.uuid == currentStatusUuid);
+      if (!newQueueHasCurrentStatus) {
+        setCurrentStatusUuid(currentQueue.allowedStatuses?.[0]?.uuid);
+      }
+    }
+  }, [currentQueue]);
+
+  const currentStatusIndex = Math.max(0, allowedStatuses?.findIndex((s) => s?.uuid == currentStatusUuid));
+
+  const visitedQueueEntriesByLocationAndQueueAndStatus = visitQueueEntriesByLocation
```
ibacher, line 69: `visited`?

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,216 @@
...
+      <Tabs
+        selectedIndex={currentStatusIndex}
+        onChange={({ selectedIndex }) => setCurrentStatusUuid(allowedStatuses[selectedIndex].uuid)}
+        className={styles.tabs}>
+        <TabList style={{ paddingLeft: '1rem' }} aria-label="Outpatient tabs" contained>
+          {allowedStatuses?.map((s) => <Tab key={s?.uuid}>{s?.display}</Tab>)}
+        </TabList>
+        <TabPanels>
+          {allowedStatuses?.map((s) => (
+            <TabPanel key={s?.uuid} style={{ padding: 0 }}>
```
ibacher, line 109: We shouldn't use something as a key if it may be `undefined`. It would be better to use the `index` argument, if this may not be set.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,216 @@
...
+          itemToString={(item) => (item ? item.display : '')}
+          onChange={handleServiceChange}
+          size="sm"
+        />
+      </div>
+      <Tabs
+        selectedIndex={currentStatusIndex}
+        onChange={({ selectedIndex }) => setCurrentStatusUuid(allowedStatuses[selectedIndex].uuid)}
+        className={styles.tabs}>
+        <TabList style={{ paddingLeft: '1rem' }} aria-label="Outpatient tabs" contained>
```
ibacher, line 104: It would be better to define a scss file and put this in a class (this should also be a standard Carbon spacing unit).

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,216 @@
...
+    results: paginatedQueueEntries,
+    currentPage,
+  } = usePagination(visitedQueueEntriesByLocationAndQueueAndStatus, currentPageSize);
+
+  const headers = queueTableColumns.map((column) => ({ header: t(column.headerI18nKey), key: column.id }));
+  const rows =
+    paginatedQueueEntries?.map((queueEntry) => {
+      const row = {};
+      queueTableColumns.forEach((conf) => {
+        row[conf.id] = <ExtensionSlot name={conf.extensionSlotName} state={{ queueEntry }} />;
```
ibacher, line 141: Yikes! This is stress-testing the extension system! I was imagining this working as something like a slot-per-column, or, better yet, a simple mapping function rather than a slot-per-cell. I guess we'll see how this works...

ibacher: For reference, a 20-person queue with just 3 columns is now 60 separate slots with 60 separate extensions.

packages/esm-service-queues-app/src/queue-table/queue-table-cells.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const { customPatientChartUrl } = useConfig<ConfigObject>();
+  return (
+    <ConfigurableLink to={customPatientChartUrl} templateParams={{ patientUuid: queueEntry.queueEntry.patient.uuid }}>
+      {queueEntry.queueEntry.display}
+    </ConfigurableLink>
+  );
+};
+
+export const QueueTablePriority = ({ queueEntry }: { queueEntry: VisitQueueEntry }) => {
+  return <>{queueEntry.queueEntry.priority.display}</>;
```
ibacher, line 16: This component really should render a pill, as shown in the [designs](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/61bb8779a7fe1d4d8997a0fc).

ibacher (changes requested):
Alright, now that I've actually seen this PR, I'm a little horrified and worried. Some more concrete things below.

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
chibongho: Good catch. Thanks. Missed this when I was refactoring the function. I'm surprised this didn't given me any errors.

packages/esm-service-queues-app/src/config-schema.ts
chibongho: Yeah, we can forgo this. This means requiring the columns headers to be unique, which should be fine.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mseaton: When would it be null?  I'd think we can just remove the `?`

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mseaton: Is that true?  If so, I agree we may need an alternative.  Ultimately, we just want to be able to have a list of columnDefinitions that can be configured with a value function or component that takes in a queueEntry as an argument/prop.  If there was a straightforward way to register functions or components with a lookup key, and then to refer to these with this key to use when rendering, that's what we need.

It isn't entirely similar, but I'd think this kind of thing could also eliminate the kind of hard-coding that one finds in places [like this](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/field/custom-field.component.tsx#L16) and [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/field/field.component.tsx#L31).  The similarity is that there is a need to support end-user configuration via a string value, and then have this string value determine what actual component is rendered.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: Sorry, that was sloppy on my part. It should never be undefined/null. I'll remove the `?`. I think using index as key is bad unless the same index always corresponds to the same object (which is not the case here).

packages/esm-service-queues-app/src/queue-table/queue-table-cells.component.tsx
chibongho: I'll make them pills. We might have to re-think the coloring if we want the flexibility of having available priorities configurable per queue (currently in [our spec](https://wiki.openmrs.org/pages/viewpage.action?pageId=351993893)). I'll worry about that in a later PR.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -119,10 +120,10 @@ export const QueueTableByStatus: React.FC = () => {
 };
 
 interface QueueTableProps {
-  visitedQueueEntriesByLocationAndQueueAndStatus: VisitQueueEntry[];
+  visitQueueEntriesByLocationAndQueueAndStatus: VisitQueueEntry[];
```
mseaton, line 123: I would expect a `QueueTable` to take in  a `queueEntries` or `visitQueueEntries` property.  Why are we calling these things `ByLocationAndQueueAndStatus`?  Ideally we'd use the same QueueTable and QueueTable props to render QueueTables in different contexts, with different columns, and with different collections of QueueEntries.  In the context above, you might set the `queueEntries` from `visitQueueEntriesByLocationAndQueueAndStatus`, but that is only this particular use case.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mseaton: This does not appear to be fixed yet.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -57,21 +57,24 @@ export const QueueTableByStatus: React.FC = () => {
...
+  const visitQueueEntriesByLocationAndQueueAndStatus = visitQueueEntriesByLocation
     ?.filter((entry) => !currentQueueUuid || entry.queueEntry.queue.uuid == currentQueueUuid)
     ?.filter((entry) => !currentStatusUuid || entry.queueEntry.status.uuid == currentStatusUuid);
 
-  if (isLoading || !allowedStatuses?.length) {
+  const noStatuses = !allowedStatuses?.length;
+  if (isLoading) {
     return <>Loading....</>;
+  } else if (noStatuses) {
+    return <>No available statuses configured for queue</>;
```
mseaton, line 77: All of these messages need to be translated, not hard-coded in English

packages/esm-service-queues-app/src/index.ts
```diff
@@ -95,9 +94,33 @@ export const addProviderToRoomModal = getAsyncLifecycle(
...
+);
+
+export const queueTableStatus = getAsyncLifecycle(
+  () =>
+    import('./queue-table/queue-table-cells.component').then(({ QueueTableStatus }) => ({ default: QueueTableStatus })),
+  {
+    featureName: 'queue table name column',
+    moduleName,
+  },
+);
```
mseaton, line 123: Can you explain why this has been modified like this, instead of just changing "getSyncLifecycle" to "get AsyncLifecycle"?

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -101,15 +104,13 @@ export const QueueTableByStatus: React.FC = () => {
...
+        <TabList style={{ paddingLeft: '1rem' }} aria-label={t('queueStatus', 'Queue Status')} contained>
           {allowedStatuses?.map((s) => <Tab key={s?.uuid}>{s?.display}</Tab>)}
         </TabList>
         <TabPanels>
           {allowedStatuses?.map((s) => (
             <TabPanel key={s?.uuid} style={{ padding: 0 }}>
-              <QueueTable
-                visitedQueueEntriesByLocationAndQueueAndStatus={visitedQueueEntriesByLocationAndQueueAndStatus}
-              />
+              <QueueTable visitQueueEntriesByLocationAndQueueAndStatus={visitQueueEntriesByLocationAndQueueAndStatus} />
```
mseaton, line 113: The entries we are passing in here have nothing to do with `ByLocation`, do they?  Filtering by `Queue` and `Status` is enough.  I don't believe we need to do anything with Location?

mseaton (changes requested):

packages/esm-service-queues-app/src/index.ts
chibongho: The two functions don't take in the same inputs. `getAsyncLifecycle` expects a `Promise` returned by dynamically calling `import("***.component")`, where `***.component.tsx` is usually a file that defines a Component that is exported with `default export`'. In this case, `queue-table-cells.component.tsx` has several exports, and therefore no `default` export, This syntax is admittedly uglier. The alternative would be to create separate files for each of those 3 components defined in `queue-table-cells.component.tsx`.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: I wanted to make it explicit that this `VisitQueueEntry[]` has been filtered. You're right that the filtering can be in different contexts and does not necessarily have to be by Location/Queue/Status. I can renamed it to `queueEntriesToDisplay`.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: Right now, I'm doing `useUnmappedVisitQueueEntries(currentLocationUuid)`, which is only getting `visitQueueEntries` from the current location (hard coded to user's session location right now). I haven't done this yet, but I plan to have the list of queues dropdown (from `useQueues()`) be limited by current location as well, and hook up the location dropdown currently defined in the queue app's home page to control this location.

packages/esm-service-queues-app/src/index.ts
mseaton: OK, defer to others like @ibacher on this.  I don't feel like I tend to see it like this in the code elsewhere, so that might imply that the right thing is to move the components into separate files with default exports?

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mseaton: The whole "mapped"  vs "unmapped" visit queue entries thing is a bit confusing to me, to be honest.  Do we need to have both of these?  I understand the value in having a nicer object to work with than the REST response, but this doesn't feel exactly right.  What is done in other esms that have user-friendly objects to work with and map to/from rest representations?

Regarding the other point, this should really depend on the nature of the overall page that is using this component.  I would not expect variable names (or assumptions) like this within a "queue-table" component.  If there is a page called "queues-for-location", then it would make sense to have variable names in that page that have names like that.  But in a generic queue table component - no.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mseaton: `queueEntriesToDisplay` is fine I guess, but I still prefer just `queueEntries` for readability and consistency.  I don't think anyone should have any expectation that this is every queue entry since the beginning of time that is expected to be passed in as a prop.  If this was always going to have the same usage, it wouldn't be a prop, you would just retrieve them within the component.  Passing it as a prop implies that the queue table takes in a list of queue entries to render, as that's the whole point.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: I intend to get rid of the `MappedVisitQueueEntry` when this new table is more feature complete and can replace the existing table. The existing table uses `MappedVisitQueueEntry`, so for now supporting both the unmapped and mapped versions is meant to be an un-invasive way to reuse old code while we try to migrate. I can add some comment about that. (@mogoodrich is also getting rid of the similar `MappedAppointment` type in his other PR)

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: @ibacher I did some rough [profiling ](https://github.com/openmrs/openmrs-esm-patient-management/commit/069738711151750b9edcdddea377b8dc7bb4a6e3) work. Using a table with 10 rows and 3 columns on my computer, I measure the time to render the cells using the components directly vs using `<ExtensionSlot>` . The direct way has 1ms of time lapse between the `useEffect(..., [])` calls of the first and the last rendered cells, while via `<ExtensionSlot>` the time lapse is 20ms. So... technically 20x worse, but still small enough to be acceptable?

packages/esm-service-queues-app/src/queue-table/queue-table-cells.component.tsx
ibacher: Yeah, my original thought was to have the color based a `ConceptAttribute`, so the queue priorities could tell you how they should appear.

packages/esm-service-queues-app/src/index.ts
ibacher: Yeah, generally, the component should be the default export, even if other things are exported (although that sounds suspiciously like code that shouldn't all be in one component file; usually a component file should just define a single React component and if things need to be shared, they go elsewhere).

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
```diff
@@ -105,8 +107,8 @@ export interface MappedVisitQueueEntry {
   queueComingFrom: string;
 }
 
-interface UseVisitQueueEntries {
-  visitQueueEntries: Array<MappedVisitQueueEntry> | null;
+interface UseVisitQueueEntries<T extends MappedVisitQueueEntry | VisitQueueEntry> {
```
mogoodrich, line 110: I haven't dove deeplying the queue module yet, so I'm just guessing about this, but if this is anything like the Appointment module, do we want to try to standardized on a single queue entry interface (probably VisitQueueEntry and retire the MappedVisitQueue entry?).  Is a lot of this code just stuff to support translating between a mapped and "normal" visit queue entry? (The answer may be yes, but we don't want to tackle it in this commit).

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
mogoodrich: seems odd that we are basing logic on string comparisions to a localized string, but I suspect this is the existing logic, not something you are introducing?

packages/esm-service-queues-app/src/index.ts
mogoodrich: Yeah, I need to refresh my memory on how all this stuff works, more or less defer to @ibacher  on this one

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
mogoodrich: Update: I see the later comments about getting rid of this, +1

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mogoodrich: If I'm following correctly :) I agree with Mike's logic here.

mogoodrich (commented):
Sorry for the late review... I added a few comments, though admittedly I'm as up to speed on the queue module to give a 100% review, but overall it looks like.  If it works for Ian I'm good to go.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
mogoodrich: Just seeing this thread now, it wasn't visible when I was looking at the PR itself... did we move away from this approach?  I agree that having a slot-per-cell seems like a lot, but don't necessarily want to push for a full refactor now if we want to just see how it works in this case.  

One goal on my end would be keep the code a simple as possible (the key being "as possible", I know there will be some complexity) from a extension and readability perspective for what I suspect will be the 80% use case for most implementions, just to use the default columns.

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
chibongho: correct. This is old code.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
chibongho: I don't love this approach., but I'm keeping it as is for now (unless others have objection). It's ok performance-wise, but developer experience isn't great as we need somewhat bloated JSON to define the columns. This approach also doesn't address how to go about having the table show different columns at different places. (For example, we might want a "current queue" column when we are looking at queue entries from all queues, but can hide the column if we're only looking at a specific queue.)

chibongho:
Ok, updated the PR.
- Separate each queue table cell component into its own file
- Separate queue-table-by-status component into its own file, remove location-related code from it
    - For now, there is no UI to filter which queue to look at, so the table renders queue entries from all queues

PR #970, opened 2024-02-07
(feat) make default onKeyUp instant seach for tablet devices configurable

donaldkibet:
## Requirements

## Summary
This PR introduces capability to disable `onKeyUp` patient search. This is informed by performance issues where network resource is limited, making every keyup hit the database. We might also want to disable the onkeyup search for desktop devices. 
@ibacher @denniskigen @vasharma05 
## Screenshots
At a facility in Homa Bay County KE
https://github.com/openmrs/openmrs-esm-patient-management/assets/28008754/bdcc02e1-a84f-4f4a-b93c-de8c002149e3

### DIFF
--- a/packages/esm-patient-search-app/src/config-schema.ts
+++ b/packages/esm-patient-search-app/src/config-schema.ts
@@ -12,6 +12,12 @@ export const configSchema = {
       _default: false,
       _description: 'Whether to show recently searched patients',
     },
+    disableTabletSearchOnKeyUp: {
+      _type: Type.Boolean,
+      _default: false,
+      _description:
+        'Disable the default "keyup search" for instant patient search as typing concludes on tablet devices',
+    },
   },
   includeDead: {
     _type: Type.Boolean,
@@ -43,4 +49,12 @@ export const configSchema = {
     _description: 'Identifier to be shown in the event defaultIdentifierTypes does is empty',
     _default: 'OpenMRS ID',
   },
+};
+
+export type PatientSearchConfig = {
+  search: { patientSearchResult: string; showRecentlySearchedPatients: string; disableTabletSearchOnKeyUp: boolean };
+  includeDead: boolean;
+  contactAttributeType: Array<string>;
+  defaultIdentifierTypes: Array<string>;
+  defaultIdentifier: string;
 };
--- a/packages/esm-patient-search-app/src/patient-search-bar/patient-search-bar.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-bar/patient-search-bar.component.tsx
@@ -15,8 +15,8 @@ interface PatientSearchBarProps {
 const PatientSearchBar = React.forwardRef<HTMLInputElement, React.PropsWithChildren<PatientSearchBarProps>>(
   ({ buttonProps, initialSearchTerm, onChange, onClear, onSubmit, small }, ref) => {
     const { t } = useTranslation();
-    const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
 
+    const [searchTerm, setSearchTerm] = useState(initialSearchTerm);
     const handleChange = useCallback(
       (val) => {
         if (typeof onChange === 'function') {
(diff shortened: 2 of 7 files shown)


makombe (approved):
LGTM!  Thanks @donaldkibet

mseaton (commented):
Why is this configuration specifically limited to tablet search?  Shouldn't this configuration apply to all patient searches that are done, regardless of device?  @makombe / @denniskigen ?

denniskigen:
@mseaton it doesn't look like this is in any way specific to tablets (except for how the config property is named). So, it should probably be named differently.

ibacher:
Yeah, this configuration property needs a different name:

1. It's not specific to tablets (and that's ok)
2. It's really got nothing to do with the `keyup` event (it's relate to React's change event, which is obliquely related to keyups, but fires for other reasons)
3. I don't think anyone else would ever figure out what this did from the name

brandones:
Reversion: https://github.com/openmrs/openmrs-esm-patient-management/pull/973

brandones:
@donaldkibet & @makombe , please refrain from making config changes without consulting anyone outside Palladium. Config changes are very high friction: once configurability is introduced, implementations start using it, and then it can't be changed without breaking those implementations. Therefore it must be designed and introduced very thoughtfully.

makombe:
@brandones  What is so far wrong with that configuration anyway? I thought through config is where implementors can extend some implementation specific features without necessarily breaking the community specific features. And in this case, this was not even Palladium specific and being that we were at the facility this was a blocker and we needed away to make our users going. We can discuss more to get  a better fix. Thanks cc @gracepotma

brandones:
@makombe Read the above comments, which explain what is wrong. If three top OpenMRS devs tell you "this doesn't make sense," then you should probably read what they have to say.

If you feel the need to push changes through in a hurry because you have deployed something broken to your users, then you need to fix your deployment process so that you don't deploy broken things to your users. Don't put that on the rest of the community by forcing through low-quality code.

I'll close the revert PR, but please fix this.

gracepotma:
@brandones : I was at this site in-person during this fix. The issue was not that the deployment had a bug: instead, through "death by a thousand cuts", there were so many performance challenges at the remote site (poor quality old hardware in-use with no new hardware to arrive in the near horizon, variable network quality, and massive amounts of previous patient data at the site from _years_ of OpenMRS v2 useage). This meant that the live-search-feature was _extremely_ slow and would cause the UI to completely hang because as you searched in the Patient Search bar for "123..." or "James...", there were so many patients with those first few ID numbers or with that first name, that the system would hang. So this was actually not a great fit for this type of site with how we had designed the live-search in the RefApp. This was _actively blocking_ all users from finding and opening patient charts in O3 during the week O3 was going live. So we all decided, and I believe @denniskigen was aware of this, that having a config that would toggle "Live Search" to instead be "Search when I click the search button when I'm done typing" certainly made sense for sites like this one. Especially since we anticipate _many_ sites like this one.

packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -12,6 +12,12 @@ export const configSchema = {
       _default: false,
       _description: 'Whether to show recently searched patients',
     },
+    disableTabletSearchOnKeyUp: {
+      _type: Type.Boolean,
+      _default: false,
+      _description:
+        'Disable the default "keyup search" for instant patient search as typing concludes on tablet devices',
```
gracepotma, line 19: If this Config is _not_ tablet specific, then the description should not be tablet-specific either ;)

gracepotma:
Thanks so much to the Palladium team for catching this important on-the-ground need and for working within the community so others may benefit. Perhaps what we need to do a better job of from the Global Support Team side is how we handle direct commits etc to mainline.

brandones:
Got it, thanks for the clarification @gracepotma . @makombe and @donaldkibet , I'm sorry if I was harsh in my response to the situation. Thank you for understanding and work.

mseaton:
> This meant that the live-search-feature was _extremely_ slow and would cause the UI to completely hang because as you searched in the Patient Search bar for "123..." or "James..."

FYI - we actually had this same issue with the O2 Refapp at many of our PIH sites, and made similar changes there to accommodate - basically search when the user wants to search (by hitting enter) and not on every keypress.  So I'm fully on board with the goals of this ticket, I'm just motivated not to make it specific only to tablet-based searching :)

PR #971, opened 2024-02-07
BREAKING: O3-2827: Appointments: Refactor (Part I) 

mogoodrich:
https://openmrs.atlassian.net/browse/O3-2827

## Requirements


packages/esm-appointments-app/src/form/appointments-form.resource.ts
```diff
@@ -0,0 +1,126 @@
+import dayjs from 'dayjs';
```
mogoodrich, line 1: Brought over from Patient Chart

packages/esm-appointments-app/src/form/appointments-form.scss
```diff
@@ -0,0 +1,105 @@
+@use '@carbon/styles/scss/colors';
```
mogoodrich, line 1: Brought over from Patient Chart

packages/esm-appointments-app/src/form/appointments-form.test.tsx
```diff
@@ -0,0 +1,205 @@
+import React from 'react';
```
mogoodrich, line 1: Brought over from Patient Chart

mogoodrich:
This admittedly a rather large PR and therefore difficult to review, best to read the explanation I have written up here:
https://openmrs.atlassian.net/browse/O3-2827

mogoodrich:
per comment on the ticket, I will likely reincorporate the distribution calendar into the form before merging.

package.json
```diff
@@ -18,14 +18,18 @@
...
     "swr": "^2.0.1",
-    "xlsx": "^0.18.5"
+    "xlsx": "^0.18.5",
+    "zod": "^3.22.2"
   },
   "devDependencies": {
     "@babel/core": "^7.11.6",
     "@carbon/react": "~1.37.0",
     "@openmrs/esm-framework": "next",
+    "@openmrs/esm-patient-common-lib": "next",
```
ibacher, line 32: Do we need this dependency? I don't see it used in any of the code marked as coming from the chart.

package.json
```diff
@@ -18,14 +18,18 @@
     "extract-translations": "turbo extract-translations -- --config ../../tools/i18next-parser.config.js"
   },
   "dependencies": {
+    "@hookform/resolvers": "^3.3.1",
```
ibacher, line 21: While this isn't terribly important, it would be preferrable, I think, to keep all dependencies in `devDependencies`. These aren't in any way different from how we handle, e.g., `dayjs`.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
```diff
@@ -70,27 +65,10 @@ const AppointmentActions: React.FC<AppointmentActionsProps> = ({ visits, appoint
...
-        const isAfterNoon = new Date().getHours() > 12;
-
-        if (scheduleType === 'Pending' && isAfterNoon) {
-          return (
-            <Button onClick={handleOpenDefaulterForm} size="sm" kind="tertiary">
-              {followUpButtonText}
-            </Button>
-          );
-        }
-
```
ibacher, line 82: I'm not in a position to say whether this is right or not, but this does seem to indicate a point where we have implementation needs for extensibility that's not currently provided.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
```diff
@@ -5,28 +5,28 @@ import utc from 'dayjs/plugin/utc';
...
 interface AppointmentActionsProps {
   visits: Array<any>;
-  appointment: MappedAppointment;
+  appointment: Appointment;
   scheduleType: string;
+  mutate: () => void;
 }
 
-const AppointmentActions: React.FC<AppointmentActionsProps> = ({ visits, appointment, scheduleType }) => {
+const AppointmentActions: React.FC<AppointmentActionsProps> = ({ visits, appointment, mutate }) => {
```
ibacher, line 25: We seem to be passing `mutate()` through three or for different levels, just so we can access it as a prop in `AppointmentForm`. This seems like a perfect place for using a React [context](https://react.dev/reference/react/createContext).

The [PatientSearchContext](https://github.com/openmrs/openmrs-esm-patient-management/blob/90ca8beba97580524d5ccfed10eb885e834589f1/packages/esm-patient-search-app/src/patient-search-context.ts#L2) is a pretty good example of how this can be used.

packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx
```diff
@@ -130,7 +133,9 @@ const AppointmentsTable: React.FC<AppointmentsTableProps> = ({
                   onClick={() =>
                     downloadAppointmentsAsExcel(
                       appointments,
-                      `${tableHeading} Appointments ${formatDate(new Date(appointments[0]?.dateTime), { year: true })}`,
+                      `${tableHeading} Appointments ${formatDate(new Date(appointments[0]?.startDateTime), {
```
ibacher, line 136: FIXME for latter, but this should all be translatable.

packages/esm-appointments-app/src/constants.ts
```diff
@@ -3,3 +3,51 @@ export const basePath = '/appointments';
 export const spaBasePath = `${window.spaBase}/home`;
 export const omrsDateFormat = 'YYYY-MM-DDTHH:mm:ss.SSSZZ';
 export const appointmentLocationTagName = 'Appointment Location';
+
+export const datePickerPlaceHolder = 'dd/mm/yyyy';
+export const dateFormat = 'DD/MM/YYYY';
+export const datePickerFormat = 'd/m/Y';
+export const weekDays = [
+  {
+    id: 'MONDAY',
```
ibacher, line 12: Is there a way we could rework this without the `id`? The label seems just as useful (and English is very unlikely to change the name of "Monday" any time soon).

packages/esm-appointments-app/src/form/appointments-form.resource.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const upcomingAppointments = appointments
+    ?.sort((a, b) => (a.startDateTime > b.startDateTime ? 1 : -1))
+    ?.filter(({ status }) => status !== 'Cancelled')
+    ?.filter(({ startDateTime }) => dayjs(new Date(startDateTime).toISOString()).isAfter(new Date()));
+
+  const todaysAppointments = appointments
+    ?.sort((a, b) => (a.startDateTime > b.startDateTime ? 1 : -1))
+    ?.filter(({ status }) => status !== 'Cancelled')
+    ?.filter(({ startDateTime }) => dayjs(new Date(startDateTime).toISOString()).isToday());
```
ibacher, line 55: This seems refactorable:

```ts
const activeAppointments = appointments
  ?.filter(({ status }) => status !== 'Cancelled') ?? [];

// no ? here since `activeAppointments` is guaranteed to be an array
const pastAppointments = activeAppointments
  .sort((a, b) => (b.startDateTime > a.startDateTime ? 1 : -1))
  .filter(({ startDateTime }) =>
      dayjs(new Date(startDateTime).toISOString()).isBefore(new Date().setHours(0, 0, 0, 0)),
    );
```

Not essential.

packages/esm-appointments-app/src/form/appointments-form.resource.ts
```diff
@@ -0,0 +1,126 @@
...
+  return openmrsFetch(`/ws/rest/v1/appointments/${appointmentUuid}`, {
+    signal: abortController.signal,
+  });
+}
+
+export function getAppointmentService(abortController: AbortController, uuid) {
+  return openmrsFetch(`/ws/rest/v1/appointmentService?uuid=` + uuid, {
+    signal: abortController.signal,
+  });
+}
```
ibacher, line 114: I haven't tracked the context in which these are used, but can they be re-written to take advantage of SWR?

packages/esm-appointments-app/src/form/appointments-form.resource.ts
```diff
@@ -0,0 +1,126 @@
...
+
+export const cancelAppointment = async (toStatus: string, appointmentUuid: string) => {
```
ibacher, line 116: Probably not this PR, but a general comment: If the only point at which you have an `await` in a function body is in the `return`, there is no point in marking it `async`.

```ts
(toStatus: string, appointmentUuid: string) => {
 // ... snip ...
 return openmrsFetch(/*... snip ...*/);
}
```

Is precisely equivalent and takes less typing.

packages/esm-appointments-app/src/index.ts
```diff
@@ -51,7 +51,7 @@ export function startupApp() {
     },
     {
       path: `${window.spaBase}/patient-list/:forDate/:serviceName`,
-      title: ([date, serviceName]) => `Patient Lists / ${decodeURI(serviceName)}`,
+      title: ([serviceName]) => `Patient Lists / ${decodeURI(serviceName)}`,
```
ibacher, line 54: Shouldn't this be:

```suggestion
      title: ([_, serviceName]) => `Patient Lists / ${decodeURI(serviceName)}`,
```

packages/esm-appointments-app/src/constants.ts
```diff
@@ -3,3 +3,51 @@ export const basePath = '/appointments';
 export const spaBasePath = `${window.spaBase}/home`;
```
ibacher, line 3: Future refactoring:

1. This should probably be called `spaHomePage`
2. The definition should be `${window.getOpenmrsSpaBase()}home` (`getOpenmrsSpaBase()` is normalised so it always returns a path ending in `/`. `window.spaBase` is a direct representation of whatever was fed into the build the process, and thus may or may not contain a trailing slash).

ibacher (approved):
A few things I noticed, but mostly comments for future work. Personal bugbear of mine: the pattern of creating an abort controller inside a function call is basically useless. The right pattern for this is _in the React component_ use our `useAbortController()` hook and pass that into any requests.

The point of abort controllers is to allow the application to cancel requests. If the abort controller's `abort()` method is never called, it doesn't do anything. `useAbortController()` calls the `abort()` method when the component is removed from the screen, so it's generally useful for aborting unnecessary GET requests.

packages/esm-appointments-app/src/home/home-appointments.resource.ts
```diff
@@ -3,22 +3,22 @@ import useSWR from 'swr';
...
   const { currentAppointmentDate } = useAppointmentDate();
 
   const apiUrl = `/ws/rest/v1/appointment/all?forDate=${currentAppointmentDate}`;
-  const { data, error, isLoading, isValidating, mutate } = useSWR<{ data: Array<Appointment> }, Error>(
+  const { data, error, isLoading, isValidating, mutate } = useSWR<AppointmentsFetchResponse, Error>(
     currentAppointmentDate ? apiUrl : null,
     openmrsFetch,
   );
 
   const results = useMemo(() => {
```
chibongho, line 20: minor: we can remove the useMemo as it's just creating an object.

packages/esm-appointments-app/src/form/appointments-form.resource.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  const { data, error, isLoading, isValidating, mutate } = useSWR<AppointmentsFetchResponse, Error>(
+    appointmentsSearchUrl,
+    fetcher,
+  );
+
+  const appointments = data?.data?.length ? data.data : null;
+
+  const pastAppointments = appointments
+    ?.sort((a, b) => (b.startDateTime > a.startDateTime ? 1 : -1))
```
chibongho, line 41: This is different from the others. is the descending sort intentional?

chibongho (approved):

package.json
mogoodrich: Thanks @ibacher !  I thought I was just keeping this consistent with the patient chart, but it looks like they are all devDependencies there, I will correct.

package.json
mogoodrich: I forget why I needed it, but it was because something wasn't working... potentially something I later removed, I will take it out and see if things still work.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
mogoodrich: Yeah, it was annoying me that I had to pass this all the way around.  I can check out out PatientSearchContext.

Another pattern I was seeing was using the useSWR call at multiple levels just to retrieve the mutate function, but that didn't seem correct, and seemed dependent on making sure you were asking for the same/correct key at multiple levels.

packages/esm-appointments-app/src/constants.ts
mogoodrich: this was stuff I just copied directly from the patient-chart when copying the appointment form over, I can look into simplifying it.

packages/esm-appointments-app/src/form/appointments-form.resource.ts
mogoodrich: I just copied this code over from the patient chart, but will test and review. I could see it making sense that you'd want descending on some views and ascending in others, maybe I need to look closer into that.  For instance, in a simple list of a patient's past appointments you might want the most recent in the past first, but for any daily view you probably want the earliest first.

packages/esm-appointments-app/src/form/appointments-form.resource.ts
mogoodrich: Will take a look, just copied this code over from patient chart so hesitant to refactor unless I can clearly test.

packages/esm-appointments-app/src/form/appointments-form.resource.ts
mogoodrich: Code I copied over, I'm not sure where they are used as well... I hope to do a more thorough cleanup of all the resource calls in the future, I'll had a TODO to the code to take a look at that then.

mogoodrich (commented):
Thanks @ibacher  @chibongho for slogging through this long PR.... I'd added comments, will go ahead and make the changes I mention.  I also want to add back in a couple features based on Antony's comments on the ticket.  will post when things are ready for re-review.

packages/esm-appointments-app/src/workload/workload-resource.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (distributionHashTable.has(date)) {
+      distributionHashTable.set(date, count);
+    }
+  });
+
+  return Array.from(distributionHashTable).flatMap(([date, value]) => ({ date: date, count: value }));
+};
+
+export const useCalendarDistribution = (
+  servieUuid: string,
```
mseaton, line 27: typo?

mogoodrich (commented):
Okay, I think I resolved and/or commented on all the PR questions...see my new comments

packages/esm-appointments-app/src/workload/workload-resource.ts
mogoodrich: fixed!

mogoodrich:
whew!  I think this is ready to go, going to merge this in, thanks everyone!

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
brandones: You can always create a new hook that does your `useSWR` call. Then you are sure to be making the same one each time. You shouldn't be passing `mutate`s around. If you don't like the aesthetic of using a fetch hook just to obtain the mutate, then you can also write a custom hook that just obtains the correct mutate function. See https://github.com/openmrs/openmrs-esm-patient-chart/blob/e52c2e74ff9a3924941acf91252dfd892a72e8bf/packages/esm-patient-orders-app/src/api/api.ts#L16 for example. Ignore the comment, it's out of date. https://github.com/openmrs/openmrs-esm-patient-chart/pull/1660

denniskigen:
Hey, @mogoodrich, there's currently a regression involving the Today's appointment widget on the dev3 home page:

CleanShot 2024-03-08 at 4  13 28@2x

From the response, it appears that the `Inpatient Ward` location doesn't have a `display` property, which we're expecting to exist [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/home/appointments-base-table.component.tsx#L190-L191) when creating the table rows to render in the AppointmentsBaseTable component. 

Is this likely indicative of an unexpected change in the locations (I'd expect locations to have a `display` property) and should we fix it there, or is it preferable to add optional chaining checks when accessing these properties on the client?

mogoodrich:
thanks @denniskigen ... I missed this this morning, but Mike just pointed it out to me, I will look at next week...

mogoodrich:
Actually, @denniskigen  do you know when this regressed? Was this something that only worked when we were running the custom Palladium build of Bahmni Appointments on dev3 (which we were doing at one point, right).  Because a lot of the Bahmni Appointments REST responses are non-standard, and, you are right, it *should* have a display property, but it doesn't and it looks like it's always been that way: https://github.com/Bahmni/openmrs-module-appointments/blob/master/omod/src/main/java/org/openmrs/module/appointments/web/mapper/AppointmentMapper.java#L266

For now, IMO, we should just build the appointments UI around this non-standard REST response (ie, just use location.name vs location.display)... fyi @mseaton  @ibacher

mseaton:
> do you know when this regressed

@mogoodrich / @denniskigen - I have a PR out to address this.  The reason for this error is that there is a particular appointment coming back that has a null location.  So if null locations are possible on appointments, then the UI just needs to account for that in the display, which it wasn't.  On a secondary note, this actually highlights a bug in which the location (if not null) would never have been displayed right due to the use of display over name.  I've fixed that in the PR too.

mogoodrich:
Makes sense, @mseaton ... so I suspect this was never working, period, but wasn't also throwing an error until we had an appointment with a null location on the demo server.

package.json
brandones: Yeah we are somewhat inconsistent on this because we changed the norm a couple times during the first couple years of O3. I don't think there is a technical difference between `devDependencies` and `dependencies` for us, with our current tooling and setup, though at least some monorepo tools install `devDependencies` at the monorepo level and `dependencies` at the package level—but I think Yarn 4 just treats them the same.

Oddly, my inclination would be the opposite of what @ibacher suggests here—I am inclined to follow the rule that a dependency that gets built/bundled goes in `dependencies`. But I am quite open to being convinced otherwise :shrug:

mogoodrich:
To clarify @denniskigen , @mseaton  did point out that it was likely a change I had made that broke this, I had forgotten I had touched that code...

denniskigen:
> Actually, @denniskigen do you know when this regressed? Was this something that only worked when we were running the custom Palladium build of Bahmni Appointments on dev3 (which we were doing at one point, right). Because a lot of the Bahmni Appointments REST responses are non-standard, and, you are right, it _should_ have a display property, but it doesn't and it looks like it's always been that way: [Bahmni/openmrs-module-appointments@`master`/omod/src/main/java/org/openmrs/module/appointments/web/mapper/AppointmentMapper.java#L266](https://github.com/Bahmni/openmrs-module-appointments/blob/master/omod/src/main/java/org/openmrs/module/appointments/web/mapper/AppointmentMapper.java?rgh-link-date=2024-03-08T23%3A54%3A35Z#L266)
> 
> For now, IMO, we should just build the appointments UI around this non-standard REST response (ie, just use location.name vs location.display)... fyi @mseaton @ibacher

It looks like this is happening again on dev3 after it got reset @mogoodrich.   Is the location a required field when scheduling an appointment?

CleanShot 2024-05-13 at 11  33 21@2x

ibacher:
Hmmm... I think the issue is that it's not required (at least in the version we're running), but the frontend assumes it's always there. It's likely this is just how the demo data is populated, which should be fixable,

mogoodrich:
Thanks @denniskigen !  Yes, location should be null-safe, not sure if this was a regression or a place where we never fixed this... I pushed up what should be a quick fix: https://github.com/openmrs/openmrs-esm-patient-management/commit/3feebb1651646b8af04e0a06d53a1bf273df5873

I have to step away now but should be able confirm that it fixed it later tonight.

mogoodrich:
Looks like it is working again, sorry and thanks for pointing it out @denniskigen

packages/esm-appointments-app/src/form/appointments-form.scss
```diff
@@ -0,0 +1,105 @@
...
+.desktop {
+  padding: 0;
+}
+
+.weekSelect {
+  max-inline-size: fit-content;
+  padding-top: spacing.$spacing-03;
+}
+
+:global(.cds--select-input):focus {
```
usamaidrsk, line 93: Hello @mogoodrich is this intended from the designs!, as this affects other fields not on the appointment form.
Thanks
cc @brandones @vasharma05

packages/esm-appointments-app/src/form/appointments-form.scss
```diff
@@ -0,0 +1,105 @@
...
+  outline: 2.5px solid colors.$red-60;
+}
+
+.errorMessage {
+  @include type.type-style('label-02');
+  color: colors.$red-60;
+  margin-top: 0.5rem;
+}
+
+:global(.cds--number input[type='number']):focus {
```
usamaidrsk, line 103: This too @mogoodrich .

brandones:
Screenshots are a requirement in PRs that make UI changes. Let's please enforce that requirement @mogoodrich @ibacher @chibongho .

Also not sure why a "refactor" included UI changes at all. A summary would have been helpful.

packages/esm-appointments-app/src/form/appointments-form.scss
mogoodrich: It's a very good point @usamaidrsk ... I just copied this scss class over as-is from the patient-chart, but I would think that *everything* in this class should be namespaces in some way to only to apply to this page... and there's likely stuff that may not be needed at all.

Maybe the best thing to do would be to comment out everything from this class and view the appointments-form from the app and then *only* add back in things that are necessary for that appointment-form, namespacing as necessary.  How does that sound @usamaidrsk ?

mogoodrich:
@brandones  summary can be found on the ticket: https://openmrs.atlassian.net/browse/O3-2827

And there were not intended to be design changes, the scss files were just copied from one ESM to another with the assumption that any styling applied in appointments-form.scss would only be applied to the appointments-form, which I was also moving over.  However, I looks like I should *not* have made that assumption, and it wasn't correct...

packages/esm-appointments-app/src/form/appointments-form.scss
usamaidrsk: Thanks for the response @mogoodrich , I think we can try with the suggestion you've made.

ibacher: Yeah... the two @usamaidrsk flagged are a bit of a weird case.

```sass
.formWrapper :global(.cds--select-input):focus
```

Would be safe, but as-written, this is overriding the behaviour of all `.cds--select-input` elements.

PR #973, opened 2024-02-10
Revert "(feat) make default onKeyUp instant seach for tablet devices configurable (#970)"

brandones:
This reverts commit ca96dc751acee9a01443b8dc1f5125bfc5ba7105.

The config is badly named and has an unhelpful description. Needs to be reworked. Until it is fixed, we should just get rid of it so we don't have to make a "breaking change" to fix it.


ojwanganto:
@brandones kindly guide on what you feel is right to have the fix in. We were at a site and we needed a fix to let the tablet users work. A commit with the desired changes is welcome so that we don't affect application use on the ground. Thank you.

PR #978, opened 2024-02-12
(test) Update provider-operations.ts

gaurav-lomte23:
Implement error handling to catch any errors that might occur during API requests and handle them appropriately, I have added the function for it if you find it useful please merge it. Thank you.


jayasanka-sack:
Hi @gaurav-lomte23 , 
Thanks a bunch for the PR! I've had a good look through it, but the proposed changes seem redundant to me. 
ex: 
```
if (!response.ok()) {
      throw new Error('Failed to fetch provider');
}
```
is pretty much the same as `expect(providerRes.ok()).toBeTruthy();.`

Also, in most cases, we don't really anticipate handling errors in tests.

PR #980, opened 2024-02-12
(feat) O3-2750  add page that can render a queue broken down by status

chibongho:
## Requirements

## Summary
The goal of this ticket is to add a new view of patients who are active in a particular queue.  Unlike existing routes/pages, this page specifically limits the view to entries in a specific queue, and renders all of the active queue entries in that queue into tables in a tabbed interface based on status.  This supports considerable configurability of whether or not a column is displayed for 'all' statuses, and what specific columns are rendered for each status view.  The intention is that this functionality provides a default view of a queue, but also provides a reusable set of configurable components to enable adding queue tables as needed throughout an O3 application.

## Related Issue

https://openmrs.atlassian.net/browse/O3-2750


packages/esm-service-queues-app/src/helpers/useQueues.ts
```diff
@@ -1,13 +0,0 @@
...
-
-export function useQueues(locationUuid?: string) {
-  const apiUrl = '/ws/rest/v1/queue' + (locationUuid ? `?location=${locationUuid}` : '');
-
-  const { data } = useSWRImmutable<{ data: { results: Array<QueueServiceInfo> } }, Error>(apiUrl, openmrsFetch);
-
-  return {
-    queues: data?.data?.results ?? [],
-  };
-}
```
mseaton, line 13: For now, we might want to keep this file in place, maybe add a comment indicating that this is deprecated in favor of the hooks/useQueues function.  This way we don't affect existing functionality while we refine these new approaches.

packages/esm-service-queues-app/src/hooks/useQueue.ts
```diff
@@ -0,0 +1,14 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+import { type QueueServiceInfo } from '../types';
```
mseaton, line 3: This QueueServiceInfo import should be removed.

packages/esm-service-queues-app/src/hooks/useQueue.ts
```diff
@@ -0,0 +1,14 @@
...
+import { useQueues } from './useQueues';
+
+export function useQueue(queueUuid?: string) {
+  const { queues, isLoading, error } = useQueues();
+
+  return {
+    queue: queues.find((q) => q.uuid == queueUuid),
```
mseaton, line 10: I'm curious about how @ibacher / @denniskigen / @brandones feel about this approach in terms of practices.  Basically, the argument that @chibongho is making is that there aren't a ton a queues in the system, and it will be common for users in the UI to toggle between different queues, so this minimizes network requests by fetching them all upfront and then finding what you need, rather than fetching them on demand.  It's an easy to change implementation detail, but would like to hear others thoughts.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -0,0 +1,25 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+import { type QueueEntry, type QueueEntrySearchCriteria } from '../types';
+import useSWR from 'swr';
+
+export function useQueueEntries(searchCriteria?: QueueEntrySearchCriteria) {
+  const searchParam = new URLSearchParams();
+  for (let [key, value] of Object.entries(searchCriteria)) {
+    searchParam.append(key, value.toString());
+  }
+  searchParam.append('v', 'full');
```
mseaton, line 10: I wonder whether it would be useful as a general practice to allow passing in a specific ref to this function (and similar), defaulting to `full` if it is undefined

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -0,0 +1,25 @@
...
+    openmrsFetch,
+  );
+
+  return {
+    queueEntries: data?.data?.results,
+    isLoading,
+    error,
+    isValidating,
+    mutate,
+  };
```
mseaton, line 24: I wonder if there is some sort of utility function that exists (or should) that handles this common paradigm of returning the results of a useSWR operation, but unpacking the results from `data?.data?.results`.  @ibacher ?

packages/esm-service-queues-app/src/hooks/useQueues.ts
```diff
@@ -0,0 +1,17 @@
...
+    'custom:(uuid,display,name,description,allowedPriorities:(uuid,display),allowedStatuses:(uuid,display))';
+  const apiUrl = `/ws/rest/v1/queue?v=${customRepresentation}` + (locationUuid ? `&location=${locationUuid}` : '');
+
+  const { data, isLoading, error } = useSWRImmutable<{ data: { results: Array<Queue> } }, Error>(apiUrl, openmrsFetch);
+
+  return {
+    queues: data?.data?.results ?? [],
+    isLoading,
+    error,
+  };
```
mseaton, line 16: Same here.  Is it worth creating a utility?

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
```diff
@@ -15,7 +15,7 @@ import { useServiceMetricsCount } from './queue-metrics.resource';
 import { useVisitQueueEntries } from '../active-visits/active-visits-table.resource';
 import styles from './clinic-metrics.scss';
 import { useQueueLocations } from '../patient-search/hooks/useQueueLocations';
-import { useQueues } from '../helpers/useQueues';
+import { useQueues } from '../hooks/useQueues';
```
mseaton, line 18: Per above comment, this should not be changed.  We should leave the original helper hook in place for now and not (yet) aim to refactor existing components.

packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
```diff
@@ -39,7 +39,7 @@ import { convertTime12to24, type amPm } from '../helpers/time-helpers';
 import dayjs from 'dayjs';
 import head from 'lodash-es/head';
 import { useQueueLocations } from './hooks/useQueueLocations';
-import { useQueues } from '../helpers/useQueues';
+import { useQueues } from '../hooks/useQueues';
```
mseaton, line 42: See above comment.

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
```diff
@@ -6,7 +6,7 @@ import { usePriority, useStatus } from '../../active-visits/active-visits-table.
 import styles from './visit-form-queue-fields.scss';
 import { type ConfigObject, useConfig, useLayoutType } from '@openmrs/esm-framework';
 import { useTranslation } from 'react-i18next';
-import { useQueues } from '../../helpers/useQueues';
+import { useQueues } from '../../hooks/useQueues';
```
mseaton, line 9: See above comment.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+import QueueTableNameCell from './cells/queue-table-name-cell.component';
+import { QueueTable } from './queue-table.component';
+import QueueTableStatusCell from './cells/queue-table-status-cell.component';
+import QueueTablePriorityCell from './cells/queue-table-priority-cell.component';
+import { useNavigate } from 'react-router-dom';
+
+interface QueueTableByStatusProps {
+  queue: Queue;
+  status: Concept;
+  configByStatus?: Map<string, QueueTableTabConfig>;
```
mseaton, line 17: Can we add some sort of comment as to what the string is here?

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+  all?: QueueTableTabConfig;
```
mseaton, line 18: We probably want to rename `all` to `allStatusTabConfig` or something more explicity.  We should also document how this is intended to be used, what happens by default, etc.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+    cellComponent: (queueEntry) => <QueueTableNameCell queueEntry={queueEntry} />,
+  },
+  {
+    headerI18nKey: 'priority',
+    cellComponent: (queueEntry) => <QueueTablePriorityCell queueEntry={queueEntry} />,
+  },
+  {
+    headerI18nKey: 'status',
+    cellComponent: (queueEntry) => <QueueTableStatusCell queueEntry={queueEntry} />,
+  },
```
mseaton, line 33: Can we add more column components in to achieve feature parity with the existing queue table rendered in the active visits table / service queues page, and then add some more of these into the default view?  We can leave off `QueueNumber`, as this is a custom column for only some implementations, but if we can add in `Coming From` and `Wait time (mins)`, those would be useful. In a later PR we can add support for actions.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+                <p className={styles.separator}>{t('or', 'or')}</p>
+                <Button kind="ghost" size="sm" renderIcon={(props) => <Add size={16} {...props} />} onClick={() => {}}>
+                  {t('addPatientToList', 'Add patient to list')}
+                </Button>
+              </Tile>
+            </div>
+          ) : null}
+          <Pagination
+            forwardText="Next page"
+            backwardText="Previous page"
```
mseaton, line 97: Can this forwardText and backwardText be translated?

packages/esm-service-queues-app/src/routes.json
```diff
@@ -70,6 +70,21 @@
       "name": "add-queue-entry-widget",
       "component": "addQueueEntry",
       "slot": "add-queue-entry-slot"
-    }
```
mseaton, line 73: These changes should be removed.  We are not adding these extensions any longer.

packages/esm-service-queues-app/src/types/index.ts
```diff
@@ -437,3 +421,125 @@ export interface WaitTime {
...
+
+export interface QueueTableColumn {
+  headerI18nKey: string;
+  cellComponent: (queueEntry: QueueEntry) => JSX.Element;
+}
+
+export interface QueueTableTabConfig {
+  columns: QueueTableColumn[];
+}
+
```
mseaton, line 433: @ibacher / @denniskigen / @brandones - I'm super surprised that the various interfaces we added below are not long-since added and available in esm-api.  Shouldn't we have common interfaces that represent our REST API that everyone shares?  We added the below here, but this really should be moved to the framework and all of the various esms can just import these common interfaces.  Is this omission intentional, or just never done?  Seems very surprising.

mseaton (commented):
@chibongho , great work pulling this together.  I added some todos, questions and comments - some for you, and others I'm hoping that others like @ibacher , @denniskigen , @brandones , and @mogoodrich might weigh in on.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
chibongho: yeah, I like that.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
chibongho: Still a bit cumbersome, but we can at least destructure the returned object by `useSWR` better by doing `{data, ...rest} = ...`, since we don't really touch anything other than `data`.

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const status = queue?.allowedStatuses.find((s) => s.uuid == statusUuid);
+  const { t } = useTranslation();
+
+  if (isLoading) {
+    return <>{t('loading', 'Loading...')}</>;
+  } else if (!queue) {
+    return <InlineNotification kind="error" title={t('invalidQueue', 'Invalid Queue')} />;
+  } else {
+    return <QueueTableByStatus queue={queue} status={status} />;
+  }
```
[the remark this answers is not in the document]
chibongho: I don't think the `loading...` thing actually confirms to Carbon design, but it is also in lots of places. They have `<SkeletonPlaceholder/>` for that (It's the blurred out tables/texts on some pages while they are loading). I just tried to play around with it but have CSS trouble with it to make it look nice.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-name-cell.component.tsx
```diff
@@ -0,0 +1,15 @@
+import { ConfigurableLink, useConfig } from '@openmrs/esm-framework';
+import React from 'react';
+import { type ConfigObject } from '../../config-schema';
+import { type QueueEntry } from '../../types';
+
+const QueueTableNameCell = ({ queueEntry }: { queueEntry: QueueEntry }) => {
+  const { customPatientChartUrl } = useConfig<ConfigObject>();
+  return (
+    <ConfigurableLink to={customPatientChartUrl} templateParams={{ patientUuid: queueEntry.patient.uuid }}>
+      {queueEntry.display}
```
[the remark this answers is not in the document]
chibongho: Doesn't completely address what you said, but I think changing it to `queueEntry.patient.person.display` will be better.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { type Concept, type Queue, type QueueTableColumn, type QueueTableTabConfig } from '../types';
+import QueueTableNameCell from './cells/queue-table-name-cell.component';
+import { QueueTable } from './queue-table.component';
+import QueueTableStatusCell from './cells/queue-table-status-cell.component';
+import QueueTablePriorityCell from './cells/queue-table-priority-cell.component';
+import { useNavigate } from 'react-router-dom';
+
+interface QueueTableByStatusProps {
+  queue: Queue;
+  status: Concept;
```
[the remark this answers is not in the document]
chibongho: It's actually not just the initial value (the state is managed by the parent component), but yeah, `selectedStatus` would be a better name.

packages/esm-service-queues-app/src/hooks/useQueue.ts
mogoodrich: Does this just return the queues themselves, or their entries?  If it just a list of queue objects, this makes sense... if this returns all the active queue entries for all the queues, I guess it depends on how big we think that could grow, seems reasonable that it would never be *that* large.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
mogoodrich: Yeah, the fact that we always have this nested "data?.data?" everywhere always seemed weird to me, a utility function would be nice (though understanding that we are unlikely to go back and refactor the *numerous* existing places where this is used.

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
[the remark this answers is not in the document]
mogoodrich: 100% agree that it would be import to distinguish between top-level pages vs components.  Not sure if there's a best practice around the best way to do this, but it definitely would have been helpful when looking through the appointments code.  

I like the idea of the "*Page" suffix on the end of page components (which is the case here, but not in many of the components, at least in the Appointments ESM), that might be preferrable to using a pages directory because it would still allow you to group components together functionally. ie the page and it's children components could all be in the same directory (for child components that it wouldn't make to live in some common components directory)

Yeah, seemed like there should be some carbon design styling around this (sounds like there is and we just have problems using it).

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  return (
+    <div className={styles.container}>
+      <div className={styles.headerContainer}>
+        <div className={!isDesktop(layout) ? styles.tabletHeading : styles.desktopHeading}>
+          <h3>{queue.display}</h3>
+        </div>
+      </div>
+      <Tabs
```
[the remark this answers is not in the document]
mogoodrich: Can we flag Paul here? :)

packages/esm-service-queues-app/src/types/index.ts
mogoodrich: Agree that these should exist in esm-api... a few things to think about:

* We will need to have a naming convention that supports interfaces for both the REST WS and FHIR representations (ie OMRSPatient vs FHIRPatient or something like that).
* We likely will need liberal use of "?" for optional fields to allow these interfaces to be used for full, default, and custom refs of an object.

mogoodrich (commented):
Still getting up to speed on the Queue's code base, but added my thoughts @chibongho  @mseaton , thanks!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
chibongho: Ok... I guess that belongs in `esm-core`

packages/esm-service-queues-app/src/hooks/useQueue.ts
chibongho: `useQueues() ` returns just the list of queue objects, not the queue entries. Given that we will have UI elements that also need to list out all possible queues (like the form to transfer patient to different queue), I think relying on `useQueues()` here will save requests.

mogoodrich: yeah, agreed 100%, thanks @chibongho

brandones: Seems like the right approach to me!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
brandones: It's 11 characters and incredibly easy to understand. Please don't write a utility function for this. It would just obfuscate and create pointless interdependency. Would be incredibly annoying to refactor, and to maintain later, and would save absolutely no work in any imaginable future scenario.

packages/esm-service-queues-app/src/types/index.ts
brandones: We intentionally take a slow and cautious approach to adding stuff to core. The right time to add it is when we start thinking it's crazy it wasn't added a long time ago. No regrets yet. The Queue stuff, idk, but the Person stuff, sure, seems like it's time.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
brandones: I wrote a small tirade and then realized I'd misunderstood what was being suggested. The thing that is being suggested, sure, fine. Is it a good use of your time? Hard to say.

packages/esm-service-queues-app/src/types/index.ts
mogoodrich: I haven't looking deeply into the code, but my thought would be if we think the the Queue REST Response provided the queue module is relatively stable, we should go ahead and create an interface for it that maps to the REST response.  This can live in the Queue module or the EMR API depending about our confidence in the stability.  Once we get into the the idea of "a mapped interface of how we want to deal with queue's on the client side" we should generally be more cautious.

packages/esm-service-queues-app/src/hooks/useQueue.ts
ibacher: I don't have a problem with the approach per se, but:

> it will be common for users in the UI to toggle between different queues

This is absolutely not how I would expect _most_ users to use the OpenMRS at all. Most users should be looking at a single queue (basically, who do I need to see next); only someone like a clinic manager is likely to ever look at more than one queue.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
ibacher: > the fact that we always have this nested "data?.data?"

Would it be clearer if this the `useSWR` was something like the below?

```ts
const { data: response, ...rest } = useSWR();

return {
  queueEntries: response?.data?.results,
  ...rest
}
```

The pattern is a result of SWR itself calling the export `data`, but what we're actually returning (in most cases) is a [`Response`](https://developer.mozilla.org/en-US/docs/Web/API/Response) object that's been enhanced with a `data` property that just stores the decoded JSON body (we have a type for this already called `FetchResponse`).

I'm not sure adding such a utility really helps with, though:

```ts
interface RestSearchResults<T> {
  results?: Array<T>;
}

interface RestSearchResultsResponse<T> extends FetchResponse<RestSearchResults<T>> {}

function unwrapSearchResults(response: RestSearchResultsResponse<T> | undefined | null): Array<T> | undefined {
  return response?.data?.results;
}

// usage

const { data: response, ...rest } = useSWR<RestSearchResultsResponse<QueueEntry>, Error>(
  apiUrl,
  openmrsFetch,
);

return {
  queueEntries: unwrapSearchResults(response),
  ...rest
};
```

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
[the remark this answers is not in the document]
ibacher: So, we already have a concept in core called a "page" (in most apps, it's just the thing called `root.component.tsx`). I wouldn't want to re-use that name, as it will add to the confusion.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return (
+      <InlineNotification
+        className={styles.inlineNotification}
+        kind={'error'}
+        lowContrast
+        subtitle={t('configureStatus', 'Please configure status to continue.')}
+        title={t('noStatusConfigured', 'No status configured')}
+      />
+    );
+  }
```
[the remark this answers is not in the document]
ibacher: Generally, I'd expect we'd use Carbon's Skeletons for loading states, but basically [this is what we should be following](https://carbondesignsystem.com/patterns/loading-pattern/).

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+import styles from '../active-visits/active-visits-table.scss';
+import { type ConfigObject } from '../config-schema';
+import { type QueueEntry, type QueueTableColumn } from '../types';
+
+interface QueueTableProps {
+  queueEntries: QueueEntry[];
+  queueTableColumns: QueueTableColumn[];
+}
+
+export function QueueTable({ queueEntries, queueTableColumns }: QueueTableProps) {
```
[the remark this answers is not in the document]
ibacher: I'm not convinced that the complexity of a generic component to do this is warranted. I can see how we could create some utilities to handle lines 37-45 and the base table but a lot of the rest of this isn't easily determined:

1. Do we have styles that apply to the container? Are these the same across different tables?
2. Should the table render a toolbar? If so, how do we handle it? (this gets complex in cases where we can't just load all of the data upfront).
3. What displays in the empty state?

Basically, I'm not yet convinced that the code saved by having a generic component to do this (which is probably at best on the order of 30 lines out of this implementation) is worth the code complexity to both implement and consume that generic component.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+      return row;
+    }) ?? [];
+
+  return (
+    <DataTable rows={rowsData} headers={headers}>
+      {({ rows, headers, getTableProps, getHeaderProps, getRowProps, getToolbarProps, onInputChange }) => (
+        <TableContainer className={styles.tableContainer}>
+          <TableToolbar
+            {...getToolbarProps()}
+            style={{ position: 'static', height: '3rem', overflow: 'visible', backgroundColor: 'color' }}>
```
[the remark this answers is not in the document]
ibacher: No. Especially if we already have SCSS styles.

brandones:
@ibacher mentioned that related to this PR there was an interesting experiment with profiling what happens when a large number of little extensions are rendered? Does someone want to elaborate on that?

chibongho:
> @ibacher mentioned that related to this PR there was an interesting experiment with profiling what happens when a large number of little extensions are rendered? Does someone want to elaborate on that?

@brandones See [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/968#discussion_r1483159505). We tried having one extension slot per cell within a 10x3 table (each slot is a relatively trivial component with no state). They take about 20ms to render on my laptop, compared to 1ms for using the components directly without extensions. So, clear performance overhead, but not terrible.

packages/esm-service-queues-app/src/types/index.ts
ibacher: I think the real test here is not how _stable_ we think something is, but how _likely it is to be needed in multiple places_.

I can buy both maintaining types for core, stable types that are used in many places (things like `OpenmrsResource`, `OpenmrsSearchResult`, maybe `Concept`, `Provider`, and `Person`). Stuff that is app specific (queues and probably queue entries) probably is best defined in the queues app.

We _shouldn't_ need to come up with our own FHIR types... the `fhir` typescript namespace should already be available, e.g. `fhir.Patient`. Also, we need to be realistic about the surface area we can cover: most of the newly-added stuff isn't even documented in our [REST docs](https://rest.openmrs.org/#openmrs-rest-api) (and the REST module's Swagger docs are often not helpful).

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
mseaton: That's fine, and I agree that in general you only want to use full when needed.  Isn't there a nice middle ground between ref and full (eg. default?).  (side-note - I have had issues getting default to always work as I expect, has anyone else seen issues with using the default rep?)

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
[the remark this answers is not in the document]
mseaton: @ibacher in the case of the queues app, the route component contains this:

```
          <Routes>
            <Route path="/" element={<Home />} />
            <Route path="/queue-table-by-status/:queueUuid" element={<QueueTableByStatusPage />} />
            <Route path="/queue-table-by-status/:queueUuid/:statusUuid" element={<QueueTableByStatusPage />} />
            <Route path="/screen" element={<QueueScreen />} />
            <Route path="/appointments-list/:value/" element={<AppointmentsTable />} />
            <Route path="/queue-list/:service/:serviceUuid/:locationUuid" element={<ServicesTable />} />
          </Routes>
```

These are the things I'm calling `pages`.  Isn't that how the end user sees them?  I mean, we are in a single-page app, so they aren't really pages, but if they show new content on the screen and they are accessible with a specific URL, then they seem like things we'd want to identify as "pages", and distinct from the various components that one might add to them.  What would you suggest?

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+        selectedIndex={currentStatusIndex}
+        onChange={({ selectedIndex }) => {
+          const newStatusUuid = allowedStatuses[selectedIndex]?.uuid;
+          const url = `/queue-table-by-status/${queue.uuid}` + (newStatusUuid ? '/' + newStatusUuid : '');
+          navigate(url);
+        }}
+        className={styles.tabs}>
+        <TabList className={styles.tabList} aria-label={t('queueStatus', 'Queue Status')} contained>
+          {allowedStatuses?.map((s) => <Tab key={s?.uuid}>{s?.display}</Tab>)}
+          {all && <Tab>{t('all', 'All')}</Tab>}
```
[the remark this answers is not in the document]
mseaton: As discussed, we can punt this for a future enhancement if desired.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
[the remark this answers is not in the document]
mseaton: Fair, I wasn't really suggesting we do that now.  It was more to get the conversation started as to whether this is a common pattern that we want to encourage to enable configurability and componentization, and to encourage standardization where possible.  Nothing we need to do for this particular PR

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+                  ))}
+                </TableRow>
+              ))}
+            </TableBody>
+          </Table>
+          {rows.length === 0 ? (
+            <div className={styles.tileContainer}>
+              <Tile className={styles.tile}>
+                <div className={styles.tileContent}>
+                  <p className={styles.content}>{t('noPatientsToDisplay', 'No patients to display')}</p>
```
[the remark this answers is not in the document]
mseaton: We discussed, and this is fine.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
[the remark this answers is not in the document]
mseaton: It's fine to use Tabs for now.  We can revisit later if needed.  @chibongho  already did ask this question in the o3 design slack channel and I think there is more to the conversation, but we can have it there and then adjust here down the line if necessary.

packages/esm-service-queues-app/src/hooks/useQueue.ts
mseaton: All great points.  Sounds like there are no major objections to the current approach, so I'm going to resolve this conversation in favor of leaving this alone.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
mseaton: So basically you are saying don't add any "v=`rep`" at all by default, and only add "v=`rep`" if a `rep` is passed in as a parameter.  Let's pick up the discussion of  the REST api throwing errors in a different thread.

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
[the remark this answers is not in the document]
ibacher: So, it's true that they aren't true pages. I'm just being picky about terminology.

In `routes.json` for example, there's an element called `extensions` and there's also an element called `pages`. These `pages` are relatively rare. For example, everything in the `/home` route is rendered in a single "page" defined [here](https://github.com/openmrs/openmrs-esm-home/blob/f4e0609219e40f9fba26aa23a91690905ea44a2c/packages/esm-home-app/src/routes.json#L6); similarly, everything in patient chart uses [a single page](https://github.com/openmrs/openmrs-esm-patient-chart/blob/4a9a90edf7500ba2d1e0ccee7d218425114bcec3/packages/esm-patient-chart-app/src/routes.json#L239). So, while from the user's perspective the elements render into routes may also appear like "pages", I don't want to muddle terminology too much. (The "pages" terminology pre-dates `routes.json`; in the migration to v5, we just refactored existing pages into JSON definitions from JS definitions).

I think the naming we have for what you're talking about (a component that takes over the entire "page") is a "dashboard". (Both the chart and home page internally call these things "dashboards"). It's a little confusing, because technically, the whole service queues app is a "dashboard" and the routes within it are just the dashboard contents.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
chibongho: `default` works for me, I'll add a `ref = 'default'` param to this funciton

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
mogoodrich: yeah, thinking this through, probably not worth the effort to fix this, but @ibacher  good (for me at least) to understood the explanation of why it works this way, so this thread provided some value.  The "data.data" always looked like it meant we were doing something wrong, but sounds likethe issue is the convention that useSWR uses to call it's response, "data".

packages/esm-service-queues-app/src/pages/queue-table-by-status-page.component.tsx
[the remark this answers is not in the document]
mogoodrich: @mseaton  you can fill me in on how everything works tomorrow... :)

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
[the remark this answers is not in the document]
mogoodrich: Agree with the above, thanks!

packages/esm-service-queues-app/src/types/index.ts
mogoodrich: @ibacher  good to know that these have already been built by others for FHIR.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
ojwanganto: I have learned something from this thread. Very helpful

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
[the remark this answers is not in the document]
chibongho: Added a new `<QueueTableByStatusSkeleton>` component for this

chibongho (commented):
Ok, made enough changes for this to not be a draft anymore. Thanks for everyone's input.

how it looks:
image

image

packages/esm-service-queues-app/src/active-visits/active-visits-table.scss
```diff
@@ -273,3 +273,7 @@ html[dir='rtl'] {
     }
   }
 }
+
+.tabList {
+  padding-left: spacing.$spacing-05;
+}
```
mseaton, line 279: Why is this added here?

packages/esm-appointments-app/translations/en.json
```diff
@@ -5,6 +5,7 @@
   "addPatientToQueue": "Add patient to queue",
   "age": "Age",
   "all": "All",
+  "allDay": "All day",
```
mseaton, line 8: Why is this coming in?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -160,6 +160,7 @@ export interface ConfigObject {
     historicalObsConceptUuid: Array<string>;
   };
   contactAttributeType: Array<string>;
+  visitQueueNumberAttributeUuid: string;
```
mseaton, line 163: Why are you adding this in?

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -0,0 +1,19 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+import { type QueueEntry, type QueueEntrySearchCriteria } from '../types';
+import useSWR from 'swr';
+
+export function useQueueEntries(searchCriteria?: QueueEntrySearchCriteria, rep: string = 'default') {
```
mseaton, line 5: I'm not 100% sure this will work in all cases.  I've seen recent cases where actually specifying "v=default" throws an error.  Which is a problem, but one we can avoid by just leaving this off altogether.  By definition, "default" will happen if no `v=...` representation is explicitly included.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-coming-from-cell.component.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import { type QueueTableColumn, type QueueTableCellComponentProps } from '../../types';
+
+const QueueTableComingFromCell = ({ queueEntry }: QueueTableCellComponentProps) => {
+  return <>{queueEntry.queueComingFrom?.display}</>;
+};
```
mseaton, line 6: I would have thought (and this goes for all of these cell components) that the cell would be exported, and then the column would be configured from this outside of this file.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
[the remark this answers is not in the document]
mseaton: OK, fine with me unless @ibacher or another has a fix suggestion for you.

mseaton (approved):
This looks great to me.  I added a few comments that might be good to address or at least consider, but happy to approve.

packages/esm-service-queues-app/src/hooks/useQueues.ts
```diff
@@ -0,0 +1,16 @@
+import { openmrsFetch } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+import { type Queue } from '../types';
+
+export function useQueues(locationUuid?: string) {
+  const customRepresentation =
+    'custom:(uuid,display,name,description,allowedPriorities:(uuid,display),allowedStatuses:(uuid,display))';
+  const apiUrl = `/ws/rest/v1/queue?v=${customRepresentation}` + (locationUuid ? `&location=${locationUuid}` : '');
+
+  const { data, ...rest } = useSWRImmutable<{ data: { results: Array<Queue> } }, Error>(apiUrl, openmrsFetch);
```
denniskigen, line 10:
```suggestion
  const { data, ...rest } = useSWRImmutable<{ data: { results: Array<Queue> } }, Error>(locationUuid ? apiUrl : null, openmrsFetch);
```

Tip: You can make data fetching [conditional](https://swr.vercel.app/docs/conditional-fetching#conditional) so that the request never happens if a `locationUuid` isn't provided.

packages/esm-service-queues-app/src/views/queue-table-by-status-view.component.tsx
```diff
@@ -0,0 +1,22 @@
+import { InlineNotification } from '@carbon/react';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useParams } from 'react-router-dom';
+import { useQueue } from '../hooks/useQueue';
+import { QueueTableByStatus } from '../queue-table/queue-table-by-status.component';
+import { QueueTableByStatusSkeleton } from '../queue-table/queue-table-by-status-skeleton.component';
```
denniskigen, line 7: The established pattern is to use default exports, so I'd expect to see:

```suggestion
import QueueTableByStatus from '../queue-table/queue-table-by-status.component';
import QueueTableByStatusSkeleton from '../queue-table/queue-table-by-status-skeleton.component';
```

instead.

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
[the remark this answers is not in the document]
denniskigen: You could try overriding its styles using:

```tsx
<TableToolbar className={styles.toolbar}>
```

And then target it using something like:

```scss
.toolbar {
	:global(.cds--table-toolbar) {
    min-height: 2rem;
  } 
}
```

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -0,0 +1,108 @@
...
+        subtitle={t('configureStatus', 'Please configure status to continue.')}
+        title={t('noStatusConfigured', 'No status configured')}
+      />
+    );
+  }
+
+  return (
+    <div className={styles.container}>
+      <div className={styles.headerContainer}>
+        <div className={!isDesktop(layout) ? styles.tabletHeading : styles.desktopHeading}>
```
denniskigen, line 78:
```suggestion
        <div className={isDesktop(layout) ? styles.desktopHeading : styles.tabletHeading}>
```

Really nitpicky: but I'd just invert the logic to avoid the negation.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
chibongho: Probably a backend bug, but in my local instance and `kgh-test`, I get different responses when I omit `v=default`, and I need the fields returned by `v=default`. I'll leave this as is for now.

mseaton:
@chibongho - this looks good to me.  Let's get this merged.

PR #997, opened 2024-02-22
 (feat) O3-2793 add actions to transition queue entries in queue table

chibongho:
## Requirements

## Summary

This PR adds "transition" actions to the queue table. A transition ends a queue entry and immediately starts a new one, albeit with (slightly) different values. This allows for us to keep a history of all queue entries over transitions in the patient's visit. The backend support for this is recently added by [this](https://openmrs.atlassian.net/browse/O3-2868). This PR contains actions to transition the queue entry status, queue entry priority and which queue a patient is on (i.e. a *transfer* to a different queue).

## Screenshots

The modal dialogs for the transitions are very similar to the dialogs already there for queue transfers in the older `<ActiveVisitsTable />` (but makes different backend calls).  

image
image
image
image


mogoodrich (approved):
LGTM, thanks @chibongho !

packages/esm-service-queues-app/src/index.ts
```diff
@@ -92,6 +92,30 @@ export const addProviderToRoomModal = getAsyncLifecycle(
...
+
+export const transitionQueueEntryStatusDialog = getAsyncLifecycle(
+  () => import('./queue-table/transitions/transition-status-dialog.component'),
+  {
+    featureName: 'transition queue entry status',
+    moduleName,
+  },
+);
+
+export const transferQueueModal = getAsyncLifecycle(
```
mseaton, line 111: Why does the naming call this a `Modal` and the others are `Dialog`?

packages/esm-service-queues-app/src/queue-table/cells/queue-table-actions-cell.component.tsx
```diff
@@ -0,0 +1,50 @@
+import { Button, OverflowMenu, OverflowMenuItem } from '@carbon/react';
+import { showModal } from '@openmrs/esm-framework';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { type QueueTableCellComponentProps, type QueueTableColumn } from '../../types';
+
+export function QueueTableActionsCell({ queueEntry }: QueueTableCellComponentProps) {
```
mseaton, line 7: Aren't there really 2 different cells here?  One is a cell containing a "Transfer" button.  The other is a cell containing an "Actions" menu.  I could see different implementations wanting one and not the other.

mseaton (commented):
@chibongho - The code for this looks well done.  My question is more around the requirements.  I generally would not expect to have different menu items to modify each property of a queue entry individually, and I would not want to transition twice if I wanted to transition both the status _and_ the priority of a queue entry.

There are also several additional properties of a Queue Entry that are not handled here, and we need to think about where these would fit in for implementations that might want them, namely `locationWaitingFor`, `providerWaitingFor`, and `priorityComments`.

The way I've been thinking of it is that we need to clearly present 2 basic options - one option is to modify the existing queue entry in place (eg. edit the status or priority, or these other fields, without creating a new queue entry) - and the other is to transition to a new queue entry.

Both of these options could likely use the same form, and just limit allowing one to change the queue itself to the "transition" case.

This likely needs a bit more analysis, but I think having 2 options (eg. Edit and Transition) that both open a form that provides the ability to enter all of the available fields, provides the most flexibility and avoid creating more queue entries than is needed/desired.

chibongho:
> @chibongho - The code for this looks well done. My question is more around the requirements. I generally would not expect to have different menu items to modify each property of a queue entry individually, and I would not want to transition twice if I wanted to transition both the status _and_ the priority of a queue entry.

@mogoodrich and I discussed this a little (albeit in a vacuum just between us). The idea is that the chance of us having to change both status and priority at the same time is slim, and that having to transition twice is not that terrible. You brought up a good point with the other possible fields to change (like priorityComments, locationWaitingFor, providerWaitingFor and sortWeight). So maybe we should have one form that's flexible enough to modify any/all fields we want, and also be able to distinguish between a transition and an edit.

> 
> There are also several additional properties of a Queue Entry that are not handled here, and we need to think about where these would fit in for implementations that might want them, namely `locationWaitingFor`, `providerWaitingFor`, and `priorityComments`.

What is locationWaitingFor and providerWaitingFor (and why do we need those in addition to which queue they are in?) Is that documented somewhere?

> 
> The way I've been thinking of it is that we need to clearly present 2 basic options - one option is to modify the existing queue entry in place (eg. edit the status or priority, or these other fields, without creating a new queue entry) - and the other is to transition to a new queue entry.
> 
> Both of these options could likely use the same form, and just limit allowing one to change the queue itself to the "transition" case.
> 
> This likely needs a bit more analysis, but I think having 2 options (eg. Edit and Transition) that both open a form that provides the ability to enter all of the available fields, provides the most flexibility and avoid creating more queue entries than is needed/desired.

Yeah, I like that. Will need more clarification from Paul/Ciaran on UI to distinguish between edits and transition.

ibacher:
> What is locationWaitingFor and providerWaitingFor (and why do we need those in addition to which queue they are in?) Is that documented somewhere?

If you like gory details, most of the discussion around how queues were designed is [on Talk](https://talk.openmrs.org/t/backend-support-for-service-delivery-queues-workflows/35247). However, the idea was basically that a queue is waiting for a resource; in the discussions, the "resources" that people might be waiting for were: a service (which is what this app defaults to assuming everything is about), a location (specifically something like "Room 123"), or a specific provider. (Technically, the queue module was supposed to be an evolution of the patientqueueing module, which always represents the resource the queue is waiting for as a location).

denniskigen (commented):
Thanks, @chibongho. Your code looks good. I've left a few comments.

packages/esm-service-queues-app/src/queue-table/transitions/transfer-queue-dialog.component.tsx
```diff
@@ -0,0 +1,186 @@
...
+          title: t('queueEntryTranferFailed', 'Error transferring queue entry'),
+          kind: 'error',
+          subtitle: error?.message,
+        });
+      });
+  };
+
+  return (
+    <div>
+      <Form onSubmit={transferQueue}>
```
denniskigen, line 103:
```suggestion
      <Form onSubmit={transferQueue}>
      	<Stack gap={4}>
```

I recommend using a [Stack](https://react.carbondesignsystem.com/?path=/docs/layout-stack--overview) to space out the form fields in all of these modals.

packages/esm-service-queues-app/src/queue-table/queue-table.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+const mockUseSession = useSession as jest.Mock;
+
+jest.mock('@openmrs/esm-framework', () => {
+  const originalModule = jest.requireActual('@openmrs/esm-framework');
+  return {
+    ...originalModule,
+    openmrsFetch: jest.fn(),
+    useConfig: jest.fn(() => ({
+      customPatientChartUrl: 'someUrl',
+    })),
```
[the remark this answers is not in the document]
brandones: Yeah I'd suggest only mocking `useConfig` if you actually are testing config options. If you just need to get your components to render with the default config, you can just add `defineConfigSchema('@openmrs/esm-service-queues-app', configSchema)` at the top of the file and the [default esm-framework mock](https://github.com/openmrs/openmrs-esm-core/blob/8a00793309f5e5b2143f98121f50075871d82761/packages/framework/esm-framework/mock.tsx#L125) will handle it for you.

packages/esm-service-queues-app/src/queue-table/transitions/transfer-queue-dialog.component.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+          <section className={styles.section}>
+            <div className={styles.sectionTitle}>{t('serviceQueue', 'Service queue')}</div>
+            <Select
+              labelText={t('selectQueue', 'Select a queue')}
+              id="queue"
+              invalidText="Required"
+              value={formState.selectedQueue}
+              onChange={(event) => setSelectedQueueUuid(event.target.value)}>
+              {queues?.map((queue) => (
```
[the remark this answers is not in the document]
chibongho: After discussing with @mseaton more, we are consolidating transitions (of queues and/or status) into a single form. I think this means we should default to the currently selected queue (in case the user only wants to transition the status).

chibongho:
PR updated. Addressed PR comments. After discussing with @mseaton, we are also consolidating the transitions into one single form. The form validates that either the status or the queue is different from current values for the submit button to be enabled.

mseaton (approved):
LGTM, thanks @chibongho !

PR #1005, opened 2024-02-29
(fix) various UI issues with transition queue entry form and queue table

chibongho:
## Requirements

## Summary

Fixes various UI issues with the transition queue entry form + queue table.

In transition queue entry form:
- the priority selection now defaults to the current value
- the form submit button now gets disabled while the form submit request is pending
- there is text to indicate the current queue entry value for queue/status/priority

In queue table by status:
- when visiting `/openmrs/spa/home/service-queues/queue-table-by-status/<queue-table-uuid>/` (without status uuid specified), we default to the first tab
- z-index is added to table toolbar content to prevent dropdown from being covered by table elements.

## Screenshots

image
"(Current value)" now placed next to current value of queue entry

image
dropdown now does not get blocked by the bell icon and "Transfer" button

### DIFF
--- a/packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
+++ b/packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
@@ -50,15 +50,8 @@ const QueueTableByStatus: React.FC<QueueTableByStatusProps> = ({
   const { queueEntries, isLoading } = useQueueEntries({ queue: selectedQueue.uuid });
   const allowedStatuses = selectedQueue.allowedStatuses;
 
-  const currentStatusUuid = selectedStatus?.uuid;
-
-  const currentStatusIndex = Math.max(0, allowedStatuses?.findIndex((s) => s.uuid == currentStatusUuid));
-
-  const queueEntriesForCurrentStatus = queueEntries?.filter(
-    (entry) => !currentStatusUuid || entry.status.uuid == currentStatusUuid,
-  );
-
-  const tableConfig = configByStatus?.get(currentStatusUuid) ?? defaultQueueTableConfig;
+  const currentStatusUuid = selectedStatus?.uuid ?? allowedStatuses?.[0]?.uuid;
+  const currentStatusIndex = allowedStatuses?.findIndex((s) => s.uuid == currentStatusUuid);
 
   const noStatuses = !allowedStatuses?.length;
   if (isLoading) {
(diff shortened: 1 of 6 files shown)


mogoodrich (approved):
LGTM, thanks @chibongho !

packages/esm-service-queues-app/src/queue-table/transitions/transition-queue-entry-modal.component.tsx
```diff
@@ -121,10 +127,16 @@ const TransitionQueueEntryModal: React.FC<TransitionQueueEntryModalProps> = ({ q
...
-                  {queues?.map((queue) => (
-                    <SelectItem key={queue.uuid} text={queue.display} value={queue.uuid}>
-                      {queue.display}
-                    </SelectItem>
+                  {queues?.map(({ uuid, display }) => (
+                    <SelectItem
+                      key={uuid}
+                      text={
+                        uuid == queueEntry.queue.uuid
+                          ? t('currentValueFormatted', '{{value}} (Current value)', { value: display })
```
mseaton, line 135: I would probably just say "Current", I don't love the use of the term "value".  Same applies to status and priority.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
```diff
@@ -50,15 +50,8 @@ const QueueTableByStatus: React.FC<QueueTableByStatusProps> = ({
...
-  const currentStatusUuid = selectedStatus?.uuid;
-
-  const currentStatusIndex = Math.max(0, allowedStatuses?.findIndex((s) => s.uuid == currentStatusUuid));
-
-  const queueEntriesForCurrentStatus = queueEntries?.filter(
-    (entry) => !currentStatusUuid || entry.status.uuid == currentStatusUuid,
-  );
-
-  const tableConfig = configByStatus?.get(currentStatusUuid) ?? defaultQueueTableConfig;
+  const currentStatusUuid = selectedStatus?.uuid ?? allowedStatuses?.[0]?.uuid;
```
mseaton, line 53: What are the implications of defaulting here to the first entry in allowed statuses?  Is that really what we want?

mseaton (approved):
Approve generally, though see my comments.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status.component.tsx
chibongho: Currently, without a status uuid specified in the URL, the tab selection defaults to the first status but shows entries from all statuses. We should default it to something. An alternative would be defaulting to "all statuses", but with how the queue table is designed right now we can't assume that the "all statuses" tab is available. Another alternative would be to only support URLs with both queue and status uuid specified.

PR #1009, opened 2024-03-01
(fix) patient list search filtering pagination when results size > pageSize

chibongho:
## Requirements

## Summary

Pagination of the patient list search table breaks when the result size > pageSize. Instead of showing the paginated results, it shows all of the results.

Current implementation is as follows: original lists => filter => paginate => transform to table rows. The last step should not require any additional search logic filtering.

Unfortunately, I need to disable the search filtering unit test as we cannot correctly test this behavior with a mocked `usePaingation` hook that comes from the `mock.tsx' file from `esm-framework`.

## Screenshots

Current behavior:
https://github.com/openmrs/openmrs-esm-patient-management/assets/509602/0eb093d3-9aab-4c30-8b78-d83eb6905930

Fix:
https://github.com/openmrs/openmrs-esm-patient-management/assets/509602/f900364a-a45f-444b-8499-99b9ee488e11

(A bit hard to see. In both cases, I'm just pressing space 3 times in the search bar)


packages/esm-patient-list-management-app/src/lists-table/lists-table.test.tsx
```diff
@@ -137,16 +124,6 @@ describe('ListsTable', () => {
...
 
-    mockedUsePagination.mockImplementation(
-      () =>
-        ({
-          currentPage: 1,
-          goTo: () => {},
-          results: patientLists.slice(0, pageSize),
-          paginated: true,
-        }) as unknown as PaginationData,
-    );
```
denniskigen, line 148: The original mistake here was always passing the `patientLists` collection to the mock function. This didn't properly mimic how the underlying function works -  usePagination receives two arguments - an array of data and a `resultsPerPage` value, which is the number of results to render on a page. I've amended the mock to match that contract, and that gets the tests passing again.

packages/esm-patient-list-management-app/src/lists-table/lists-table.component.tsx
```diff
@@ -77,7 +75,6 @@ const ListsTable: React.FC<PatientListTableProps> = ({
   const [sortParams, setSortParams] = useState({ key: '', order: 'none' });
   const [searchTerm, setSearchTerm] = useState('');
   const responsiveSize = layout === 'tablet' ? 'lg' : 'sm';
-  const debouncedSearchTerm = useDebounce(searchTerm);
```
denniskigen, line 80: I've also removed this debounce implementation. Because we're not hitting the backend when searching, we don't need to debounce the search term. We can just use the search term directly to filter the lists.

denniskigen (approved):
Thanks. This LGTM, @chibongho. I've taken the liberty of pushing a commit that fixes a couple of things.

PR #1012, opened 2024-03-01
O3-2922: Appointments: Support Visits at Parent Locations

mogoodrich:
see: https://openmrs.atlassian.net/browse/O3-2922

*Note* this PR should *NOT* be merged until a new version of the REST web services module is released with the enhancements  described in the REST web services tickets linked in the ticket above.

## Requirements

### DIFF
--- a/packages/esm-appointments-app/src/appointments/appointment-tabs.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/appointment-tabs.component.tsx
@@ -4,7 +4,6 @@ import { Tab, TabList, Tabs, TabPanel, TabPanels } from '@carbon/react';
 
 import { type ConfigObject } from '../config-schema';
 import { useConfig } from '@openmrs/esm-framework';
-import { useVisits } from '../hooks/useVisits';
 import ScheduledAppointments from './scheduled/scheduled-appointments.component';
 import UnscheduledAppointments from './unscheduled/unscheduled-appointments.component';
 import styles from './appointment-tabs.scss';
@@ -16,9 +15,6 @@ interface AppointmentTabsProps {
 const AppointmentTabs: React.FC<AppointmentTabsProps> = ({ appointmentServiceType }) => {
   const { t } = useTranslation();
   const [activeTabIndex, setActiveTabIndex] = useState<number>(0);
-
-  const { isLoading = false, visits = [], mutateVisit } = useVisits();
-
   const { showUnscheduledAppointmentsTab } = useConfig<ConfigObject>();
 
   const handleTabChange = ({ selectedIndex }: { selectedIndex: number }) => {
@@ -35,7 +31,7 @@ const AppointmentTabs: React.FC<AppointmentTabsProps> = ({ appointmentServiceTyp
           </TabList>
           <TabPanels>
             <TabPanel className={styles.tabPanel}>
-              <ScheduledAppointments {...{ visits, isLoading, appointmentServiceType, mutateVisit }} />
+              <ScheduledAppointments appointmentServiceType={appointmentServiceType} />
             </TabPanel>
             <TabPanel className={styles.tabPanel}>
               <UnscheduledAppointments />
@@ -43,7 +39,7 @@ const AppointmentTabs: React.FC<AppointmentTabsProps> = ({ appointmentServiceTyp
           </TabPanels>
         </Tabs>
       ) : (
-        <ScheduledAppointments {...{ visits, isLoading, appointmentServiceType, mutateVisit }} />
+        <ScheduledAppointments appointmentServiceType={appointmentServiceType} />
       )}
     </div>
   );
--- a/packages/esm-appointments-app/src/hooks/useVisits.ts
+++ b/packages/esm-appointments-app/src/hooks/useVisits.ts
@@ -5,15 +5,18 @@ import { useAppointmentDate } from '../helpers';
 
 /**
  * Custom hook to fetch visits from the OpenMRS REST API.
+ * This fetches all visits that started on the currently selected date and are associated with the current session location.
  * @returns An object containing the visits, isLoading flag, and error message.
  */
 export const useVisits = () => {
   const { currentAppointmentDate } = useAppointmentDate();
   const session = useSession();
 
-  const visitsUrl = `${restBaseUrl}/visit?includeInactive=true&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime)&fromStartDate=${dayjs(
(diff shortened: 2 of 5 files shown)


packages/esm-appointments-app/src/appointments/appointment-tabs.component.tsx
```diff
@@ -4,7 +4,6 @@ import { Tab, TabList, Tabs, TabPanel, TabPanels } from '@carbon/react';
 
 import { type ConfigObject } from '../config-schema';
 import { useConfig } from '@openmrs/esm-framework';
-import { useVisits } from '../hooks/useVisits';
```
mogoodrich, line 7: Instead of fetching the visits here and passing them down as a prop, I moved the fetch to the actions component, where they are actually used/

packages/esm-appointments-app/src/hooks/useVisits.ts
```diff
@@ -5,15 +5,18 @@ import { useAppointmentDate } from '../helpers';
...
   const { currentAppointmentDate } = useAppointmentDate();
   const session = useSession();
 
-  const visitsUrl = `${restBaseUrl}/visit?includeInactive=true&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime)&fromStartDate=${dayjs(
+  const visitsUrl = `${restBaseUrl}/visit?includeInactive=true&includeParentLocations=true&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime)&fromStartDate=${dayjs(
     currentAppointmentDate,
-  ).format('YYYY-MM-DD')}&location=${session?.sessionLocation?.uuid}`;
+  ).format('YYYY-MM-DD')}&toStartDate=${dayjs(currentAppointmentDate)
+    .add(1, 'day')
+    .format('YYYY-MM-DD')}&location=${session?.sessionLocation?.uuid}`;
```
mogoodrich, line 19: This is the main change, it adds a parameters to include parent locations and set a max start date when searching for visits.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.test.tsx
```diff
@@ -48,6 +48,18 @@ const appointment: Appointment = {
   extensions: [],
 };
 
+let visits;
```
mogoodrich, line 51: Is using a let like this a best pratice for defining a mocked function when I want to change the return value between tests?

packages/esm-appointments-app/src/hooks/useVisits.ts
mseaton: Assuming you make the change I suggested in the REST WS module, you should be able to set the `toStartDate` to the same value you set the `fromStartDate` to, and it will return all visits that started on that date.  Otherwise, won't this also return visits from the next date at midnight, which isn't really what you want?

mseaton (commented):
Made one comment to review @mogoodrich .  Otherwise, this doesn't necessarily need to wait to be merged until the REST fix is merged, does it?  Will the rest endpoint fail if it encounters parameters in the request that it doesn't recognize?  I would assume it will just ignore them.

packages/esm-appointments-app/src/hooks/useVisits.ts
mogoodrich: Good point re: visits that start at midnight.

packages/esm-appointments-app/src/hooks/useVisits.ts
```diff
@@ -6,16 +6,15 @@ import SelectedDateContext from './selectedDateContext';
...
  * Custom hook to fetch visits from the OpenMRS REST API.
+ * This fetches all visits that started on the currently selected date and are associated with the current session location.
  * @returns An object containing the visits, isLoading flag, and error message.
  */
 export const useVisits = () => {
   const { selectedDate } = useContext(SelectedDateContext);
   const session = useSession();
-
-  const visitsUrl = `${restBaseUrl}/visit?includeInactive=true&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime)&fromStartDate=${dayjs(
+  const visitsUrl = `${restBaseUrl}/visit?includeInactive=true&includeParentLocations=true&v=custom:(uuid,patient:(uuid,identifiers:(identifier,uuid),person:(age,display,gender,uuid)),visitType:(uuid,name,display),location:(uuid,name,display),startDatetime,stopDatetime)&fromStartDate=${dayjs(
```
mogoodrich, line 15: This is the main change, it adds a parameters to include parent locations

packages/esm-appointments-app/src/hooks/useVisits.ts
```diff
@@ -6,16 +6,15 @@ import SelectedDateContext from './selectedDateContext';
 
 /**
  * Custom hook to fetch visits from the OpenMRS REST API.
+ * This fetches all visits that started on *or after* the currently selected date and are associated with the current session location.
```
mogoodrich, line 9: Note that this useVisit fetches all visits started on *or after* the currently selected date, while I think the intent was for it *just* to fetch visits on the selected date.

From a functional perspective, I don't think this introduced any bugs because the *only* time this is actually used to trigger anything is if the day it "today", but there's a lot of needless loading.  I will ticket cleaning this up.

mseaton (approved):
LGTM @mogoodrich .  I guess that's one way to avoid the question we had about end date ;)

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.test.tsx
chibongho: You lose TypeScript's implicit typing this way. `visits` now has an `any` type. I'd just define them locally in the `it()` functions.

chibongho (approved):

mogoodrich:
Yeah @mseaton  I rabbit-holed  a bit on this, and decided to get this merged in and ticket/work on the other stuff separately.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.test.tsx
mogoodrich: I needed to find away to change the return value between test runs, I think I figured out the best practice, updating the PR.

jayasanka-sack:
Hi @mogoodrich, just a friendly reminder to include the [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label in your PR titles. 
Thank you!

PR #1020, opened 2024-03-09
(fix) O3-2698 enable configuration of status and priority styles

mseaton:
## Requirements

## Summary
The primary goal of this PR is to replace hard-coded references or inferences about specific services, priorities, and statuses with a configuration-based approach that allows for removing such references.  In doing so, secondary goals are to ensure the configured components for priority and status are used consistently in the application whether in the active visits table, the queue by status table, or elsewhere.

* Moves all rendering of the queue priority, status, and duration into dedicated components
  - Results in queue-table-by-status rendering appropriate priority tags and status icons
  - Results in active-visits table updating duration as time elapses

* Removes hard-coded references to specific priorities, statuses, and services currently used for styling

* Adds optional configuration to support styling of specific priorities and statuses that was removed
  
* Opportunistic fixes
  - Fixes default "QueueNumber" -> "Queue Number" column heading
  - Fixes rendering of status icon to ensure it does not shrink when window area is reduced
  - Fixes rendering of priority tags to ensure they do not wrap text prematurely
  

## Related Issues
https://openmrs.atlassian.net/browse/O3-2698
https://openmrs.atlassian.net/browse/O3-2774


subhamkumarr (approved):

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries-dialog.resource.ts
```diff
@@ -25,10 +26,7 @@ export async function batchClearQueueEntries(queueEntries: Array<MappedVisitQueu
           },
           signal: abortController.signal,
           body: {
-            location: queueEntries[index]?.visitLocation,
-            startDatetime: parseDate(queueEntries[index]?.visitStartDateTime),
-            visitType: queueEntries[index]?.visitTypeUuid,
-            stopDatetime: new Date(),
+            stopDatetime: endedAt,
```
mseaton, line 29: This was an incidental discovery that t I ran across it as I was making changes to MappedVisitQueueEntry, and it appears to be a bug.  Basically, if you look at the code being removed in MappedVisitQueueEntry, it had a mapping as follows:  `visitStartDateTime: queueEntry.startedAt`

Then, in this part of the code, it is setting the `startDatetime` of the visit that is attached to each queueEntry being cleared from the queue to this mapped `visitStartDateTime`.  So if the queue entry being removed is from a later point in the visit than the very beginning, this will end up moving the visit start date to this later date and closing the visit.

In general, there doesn't seem to be any good reason for setting any property on visit here other than the stopDatetime, so I am removing them all.  If anyone can tell me why we should be doing this, we can discuss.

I am also ensuring that the endDate being set is defined once and set consistently across all queueEntries and visits, rather than setting an incrementally higher value as we iterate through the list.

FWIW, I think stopping the visit at all here is a mistake, and this is something we likely want to address in a subsequent ticket / discussion.

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
```diff
@@ -34,21 +30,18 @@ export interface MappedVisitQueueEntry {
   encounters: Array<MappedEncounter>;
   name: string;
   patientAge: string;
```
raghavawasthi24, line 32: patientAge would always be a **number**. I think patientAge type should be number instead of string.

raghavawasthi24 (changes requested):
I found an issue with type of patientAge.

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
ibacher: 1. You'll note that this property isn't added by this PR.
2. Did you look at the context in which this is used? If you had you might've realised that this will always be a number.

Please read the code you are reviewing. If you are unsure of something, it's better to phrase it as a question, i.e., "Shouldn't this be a number?" or "Why isn't this a number?"

packages/esm-service-queues-app/src/queue-entry-table-components/queue-priority.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+  priorityComment?: string;
+}
+
+const QueuePriority: React.FC<QueuePriorityProps> = ({ priority, priorityComment }) => {
+  const { priorityConfigs } = useConfig<ConfigObject>();
+  const priorityConfig = priorityConfigs?.find((c) => c.conceptUuid === priority.uuid);
+  return (
+    <>
+      {priorityComment ? (
+        <DefinitionTooltip className={styles.tooltip} align="bottom-left" definition={priorityComment}>
```
chibongho, line 19: Don't need to address it in this PR, but using tooltip means it won't work for mobile/tablet. We'll need better UX guidance on this.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-status-cell.component.tsx
```diff
@@ -1,8 +1,9 @@
 import React from 'react';
 import { type QueueTableColumn, type QueueTableCellComponentProps } from '../../types';
+import QueueStatus from '../../queue-entry-table-components/queue-status.component';
 
 export const QueueTableStatusCell = ({ queueEntry }: QueueTableCellComponentProps) => {
-  return <>{queueEntry.status.display}</>;
+  return <QueueStatus status={queueEntry.status} />; // Do not pass queue into status, as we do not want to render it
```
chibongho, line 6: not sure what the comment meant...

chibongho (approved):
Looks good in general. Some minor comments.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -175,3 +193,14 @@ export interface ConfigObject {
 export interface OutpatientConfig {
   visitTypeResourceUrl: string;
 }
+
+export interface PriorityConfig {
+  conceptUuid: string;
+  tagType: string;
+  tagClassName: 'priorityTag' | 'tag' | null;
```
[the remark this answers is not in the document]
mseaton: @chibongho .  The `tagType` is the primary way to do this with these tags it seems.  There are 11 types built into carbon for various color combinations:  https://react.carbondesignsystem.com/?path=/docs/components-tag--overview .  My guess is the `tagClassName` was added so that we could have a tag that exactly matched the mockup color scheme.  My feeling is we should just remove it entirely in favor of just using one of the 11 built-in tag types that are carbon "approved", but I was just trying to maintain feature parity here for now.

packages/esm-service-queues-app/src/queue-entry-table-components/queue-priority.component.tsx
mseaton: Yep.  We can ticket that separately as needed.  My goal here was to maintain existing behavior, just do so consistently

packages/esm-service-queues-app/src/queue-table/cells/queue-table-status-cell.component.tsx
mseaton: Might be a hacky way to implement it, but if you pass both status and queue into the tag, then it will render the status as `status.display - queue.display`.  If you only pass the status in, then it will render it as `status.display` only.  So in views like the active visits table in which you may be viewing multiple queues at a location in one table, this allows the pre-existing behavior to continue to work, where both are shown.  But in the queue-table-by-status component, which only ever shows entries from a single queue, it allows that pre-existing behavior to work by excluding the queue name from the status display.

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries-dialog.resource.ts
mseaton: Follow-up here.  I have raised a separate PR to address this particular issue, as it is a standalone and important bug that needs to be documented and prioritized, so this change will ultimately not be part of this PR.  Please find the separate PR here:  #1023

makombe:
@mseaton I was trying to configure queue priority as per the example in the ticket but nothing changes. Not sure what `tagClassName` should be see the attached screenshot. The conceptUuid provided is for emergency priority. Thanks
Screenshot from 2024-03-17 16-43-59

Screenshot from 2024-03-17 16-45-15

PR #1022, opened 2024-03-11
(fix) O3-2478 ensure configuration of queue services, priorities, and statuses is consistent with the backend

mseaton:
## Requirements

## Summary

The overall goal of this ticket is to fix the app to remain compatible with changes to the queue module in which priorities and statuses can be defined for a specific queue, and for which available services, priorities, and statuses are already configured on the backend and should not be re-configured on the frontend.

During the course of this work, some bugs and improvements were also discovered and addressed, namely:
* The list of service queues presented on the "Add Queue" are now alphabetized
* Previously you could not submit a new queue entry when creating a visit with the second priority in the list, due to the fact that the second priority was always selected by default, and submitting the form requires changing the priority (via use of the content switcher).  This fixes things so that no priority is selected by default, which enables selecting the appropriate priority as a means to submit the form, which seems to have been the originally expected behavior.

## Related Issue
https://openmrs.atlassian.net/browse/O3-2478


packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
```diff
@@ -82,6 +79,11 @@ const ScheduledVisits: React.FC<{
   const { queueLocations } = useQueueLocations();
   const [selectedQueueLocation, setSelectedQueueLocation] = useState(queueLocations[0]?.id);
 
+  // TODO: This needs fixing, we cannot just take the first queue and assume that is what is wanted
```
mseaton, line 82: This is the pre-existing behavior, I am just moving this declaration higher up in the file.  I have ticketed the need to change this here:  https://openmrs.atlassian.net/browse/O3-2955 but it is unrelated to this particular issue.

packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
```diff
@@ -132,7 +129,7 @@ const ScheduledVisits: React.FC<{
                 addQueueEntry(
                   response.data.uuid,
                   patientId,
-                  priority ? priority : defaultPriority,
```
mseaton, line 135: It does not make sense to default to a priority here, as the only way this form is submitted is by clicking on the priority content switcher.

packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
```diff
@@ -238,15 +235,14 @@ const ScheduledVisits: React.FC<{
                     ) : hasPriority && ind == visitsIndex ? (
                       <ContentSwitcher
                         size="sm"
-                        selectedIndex={1}
+                        selectedIndex={null}
```
mseaton, line 238: This fixes a (unrelated, existing) bug that I discovered during testing prevented the user from clicking on whatever the second listed priority is.  So this means that no priority is selected by default, and the user can choose any of the available ones, which results in the form submission.

packages/esm-service-queues-app/src/patient-search/visit-form/queue.resource.ts
```diff
@@ -27,20 +26,12 @@ export async function addQueueEntry(
...
-          uuid: queueServiceUuid,
-        },
-        patient: {
-          uuid: patientUuid,
-        },
+        status: status,
+        priority: priority,
+        queue: queueServiceUuid,
+        patient: patientUuid,
         startedAt: new Date(),
```
mseaton, line 35: I was getting errors without making the above changes.  I'm not sure why we were submitting like we were before, but with these changes I'm able to get things posted correctly.

packages/esm-service-queues-app/src/active-visits/change-status-dialog.component.tsx
```diff
@@ -31,14 +31,14 @@ const ChangeStatus: React.FC<ChangeStatusDialogProps> = ({ queueEntry, closeModa
 
   const [priority, setPriority] = useState(queueEntry?.priority?.uuid);
   const [newQueueUuid, setNewQueueUuid] = useState('');
-  const { priorities } = usePriority();
+  const priorities = queueEntry.queue.allowedPriorities;
```
ojwanganto, line 34: +1

packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
ibacher: Maybe we can just remove this property altogether?

packages/esm-service-queues-app/src/hooks/useConcept.ts
```diff
@@ -0,0 +1,13 @@
+import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import useSWRImmutable from 'swr/immutable';
+import { type Concept } from '../types';
+
+export function useConcept(uuid: string) {
```
ojwanganto, line 5: I was wondering if such common/shared functions can be consolidated in a shared lib for reuse. Many of us will keep defining such in every app that interact with concepts. In the absence of such, we have to go this route

packages/esm-service-queues-app/src/hooks/useConcept.ts
```diff
@@ -0,0 +1,13 @@
...
+  const apiUrl = `${restBaseUrl}/concept/${uuid}`;
```
ibacher, line 6: Basically allow us to _not_ make a request if the UUID isn't yet available.

```suggestion
  const apiUrl = uuid ? `${restBaseUrl}/concept/${uuid}` : undefined;
```

packages/esm-service-queues-app/src/hooks/useSystemSetting.ts
```diff
@@ -0,0 +1,23 @@
...
+import useSWRImmutable from 'swr/immutable';
+
+export interface SystemSetting {
+  uuid: string;
+  property: string;
+  value: string;
+}
+
+export function useSystemSetting(setting: string) {
+  const apiUrl = `${restBaseUrl}/systemsetting?&v=custom:(value)&q=${setting}`;
```
ibacher, line 11: It may be less ambiguous to use:

```suggestion
  const apiUrl = `${restBaseUrl}/systemsetting/${setting}?v=custom:(value)`;
```

Since `q` is a prefix search.

ibacher (approved):
Basically LGTM

packages/esm-service-queues-app/src/hooks/useConcept.ts
ojwanganto: @ibacher what is the best practice?

ibacher: Well, I'd probably think that, for things like this, we'd have `useConceptSet()` or something. In general, my biggest concern is that loading one concept at a time is one HTTP request per concept, so usually we'd want to bulk load a series of concepts, such as using the `/conceptreferences` endpoint.

ojwanganto: That makes sense to me. I know one challenge we've had is the need to edit our dictionaries to take care of the concept set and the members. Some implementers use the dictionary as it comes from i.e. OCL exports. The guidance is clear though. Thanks

ojwanganto (approved):
Thanks @mseaton

packages/esm-service-queues-app/src/hooks/useConcept.ts
mseaton: These aren't mutually exclusive, and in this case - loading a Concept and loading a ConceptSet are exactly the same thing, as a ConceptSet is just a Concept that has setMembers.  So it isn't really an anti-pattern in this case, there is exactly one request.

I 1000% think that we should have things like "useConcept" and "useSystemSetting" already defined and in esm-core somewhere (eg. esm-api), and I am continually shocked that we don't.  It cost me a non-zero amount of time to create these, and I don't want to have to independently create (and test!) all of these everywhere we want to use them.

@ojwanganto - I fully intend to endorse moving things to esm-api.  I just didn't want any philosophical debates muddying this PR, or any dependencies on getting this merged in.  Let's ticket and do that separately.

packages/esm-service-queues-app/src/patient-search/patient-scheduled-visits.component.tsx
mseaton: Yeah, I fought this for a while.  That doesn't actually work.  The ContentSwicher component will default this to `0` if not specified, which would make it impossible to select the first element.  Setting it to null explicitly is the only thing that got it to work.  The _right_ solution is not to try to turn a content switcher into a pseudo radio + submit button.  But I'm trying to avoid getting into that here...

packages/esm-service-queues-app/src/hooks/useSystemSetting.ts
mseaton: Fixed, though I couldn't commit this as-is without changing the handling from a List to single value.

mseaton:
Thanks @ibacher  and @ojwanganto for the comments and feedback - merged in.

PR #1023, opened 2024-03-11
(fix) O3-2949 visits should not be modified when queue is cleared

mseaton:
## Requirements

- [ x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x ] My work includes tests or is validated by existing tests.

## Summary
The existing "Clear Queue" function will incorrectly shift the visit start date to the most recent queue startedAt date for each queue entry being cleared.  It may also modify the visit location and type.  This is a potentially significant issue.  The solution proposed in this PR is to eliminate the manipulation and closing of visits entirely from this operation.  Although there are definitely use cases or workflows where one could imagine the desire to close visits as well, this is not generally accepted or expected behavior, and even if it were desired, we would want to make sure this was a transactional operation, not a sequential operation that could leave the visit in an inconsistent state.  So the best solution for now is to eliminate this entirely, and create a future ticket to enhance this capability with the ability to close visits as well given appropriate configuration settings.

## Related Issue
https://openmrs.atlassian.net/browse/O3-2949

### DIFF
--- a/packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
+++ b/packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
@@ -43,10 +43,8 @@ export interface MappedVisitQueueEntry {
   service: string;
   status: QueueStatus;
   statusUuid: string;
-  visitStartDateTime: string;
   visitType: string;
   visitUuid: string;
-  visitLocation: string;
   visitTypeUuid: string;
   waitTime: string;
   queueUuid: string;
@@ -158,9 +156,7 @@ export const mapVisitQueueEntryProperties = (
   status: queueEntry.status.display as QueueStatus,
   statusUuid: queueEntry.status.uuid,
   waitTime: queueEntry.startedAt ? `${dayjs().diff(dayjs(queueEntry.startedAt), 'minutes')}` : '--',
-  visitStartDateTime: queueEntry.startedAt,
   visitType: queueEntry.visit?.visitType?.display,
-  visitLocation: queueEntry.visit?.location?.uuid,
   queueLocation: (queueEntry?.queue as any)?.location?.uuid,
   visitTypeUuid: queueEntry.visit?.visitType?.uuid,
   visitUuid: queueEntry.visit?.uuid,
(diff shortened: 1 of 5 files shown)


packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
```diff
@@ -158,9 +156,7 @@ export const mapVisitQueueEntryProperties = (
   status: queueEntry.status.display as QueueStatus,
   statusUuid: queueEntry.status.uuid,
   waitTime: queueEntry.startedAt ? `${dayjs().diff(dayjs(queueEntry.startedAt), 'minutes')}` : '--',
-  visitStartDateTime: queueEntry.startedAt,
```
mseaton, line 161: Note:  This mapping is actually the root of the problem.

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries-dialog.resource.ts
```diff
@@ -1,46 +1,6 @@
...
-          method: 'POST',
-          headers: {
-            'Content-Type': 'application/json',
-          },
-          signal: abortController.signal,
-          body: {
-            location: queueEntries[index]?.visitLocation,
-            startDatetime: parseDate(queueEntries[index]?.visitStartDateTime),
-            visitType: queueEntries[index]?.visitTypeUuid,
-            stopDatetime: new Date(),
```
mseaton, line 31: Although we could technically keep all of this in place and just fix the startDatetime mapping as indicated above, it seems much safer to remove this altogether for the reasons listed in the PR description.

packages/esm-service-queues-app/src/active-visits/active-visits-table.resource.ts
ojwanganto: Does this code change the visit information in the database? I don't have the full context but it appears to be constructing data for queue entries.

ojwanganto: I had not seen the full code. This is OK with me.

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries-dialog.resource.ts
ojwanganto: Am fine with this. I remember you already added something to take care of checked-out clients

ojwanganto (approved):

PR #1026, opened 2024-03-13
(enhc) disable submit button and show submitting status while user clicks for queue transition.

Justus446:
This commit  fixes the **Move to next service** button in service queues by:
1. Disabling the button once the request is sent to prevent re-click by users
2. Added a loading spinner (for users process is ongoing in the background)
image


makombe:
@Justus446 please look into the failing build

donaldkibet (approved):
Thanks @Justus446

PR #1033, opened 2024-03-13
(fix) O3-2474 adding a patient to a queue should use active visit

mseaton:
## Requirements

## Summary

When using the service queues app, there is an option to "Add patient to queue", however the form this enables will always attempt to first start a visit before adding a new queue entry associated with that visit.  This fails to correctly handle two scenarios:
1. If the patient selected already has a currently active visit, then the user should not be prompted to complete a start visit form that includes queue fields, but should just be prompted to fill out a form with queue fields, and this should be associated with the already active visit.
2. If the patient selected is already active on the queue, then there needs to be improved validation letting the user know that the patient cannot be added because they are already on the queue

## Screenshots
adding-to-queue-existing-visit

## Related Issue
https://openmrs.atlassian.net/browse/O3-2474
https://openmrs.atlassian.net/browse/O3-2624

### DIFF
--- a/packages/esm-service-queues-app/src/overlay.scss
+++ b/packages/esm-service-queues-app/src/overlay.scss
@@ -11,8 +11,8 @@
   border-left: 1px solid $text-03;
   background-color: $ui-01;
   overflow-y: auto;
-  height: 100vh;
-  z-index: 99999;
+  height: calc(100vh - 3rem);
+  z-index: 3;
 }
 
 .desktopOverlay::after {
--- a/packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
+++ b/packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
@@ -1,0 +1,119 @@
+import React, { useCallback, useState } from 'react';
+import { Button, ButtonSet, Form, Row, Stack } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+import {
+  type ConfigObject,
+  ExtensionSlot,
+  showSnackbar,
+  useConfig,
+  useLayoutType,
+  type Visit,
+} from '@openmrs/esm-framework';
+import { addQueueEntry, useVisitQueueEntries } from '../../active-visits/active-visits-table.resource';
+import { type SearchTypes } from '../../types';
+import styles from './visit-form.scss';
+
+interface ExistingVisitFormProps {
+  toggleSearchType: (searchMode: SearchTypes, patientUuid) => void;
+  visit: Visit;
+  closePanel: () => void;
+}
+
+const ExistingVisitForm: React.FC<ExistingVisitFormProps> = ({ visit, toggleSearchType, closePanel }) => {
+  const { t } = useTranslation();
+  const isTablet = useLayoutType() === 'tablet';
+  const [isSubmitting, setIsSubmitting] = useState(false);
+
+  const config = useConfig() as ConfigObject;
+  const visitQueueNumberAttributeUuid = config.visitQueueNumberAttributeUuid;
+  const { mutate } = useVisitQueueEntries('', '');
+
+  if (!visit) {
+    return null;
+  }
+
+  const handleSubmit = useCallback(
+    (event) => {
+      event.preventDefault();
+
+      // retrieve values from queue extension
+      const queueLocation = event?.target['queueLocation']?.value;
+      const serviceUuid = event?.target['service']?.value;
+      const priority = event?.target['priority']?.value;
+      const status = event?.target['status']?.value;
+      const sortWeight = event?.target['sortWeight']?.value;
+
+      setIsSubmitting(true);
+
+      addQueueEntry(
+        visit.uuid,
+        serviceUuid,
+        visit.patient.uuid,
+        priority,
+        status,
+        sortWeight,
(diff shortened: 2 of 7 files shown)


packages/esm-service-queues-app/src/overlay.scss
```diff
@@ -11,8 +11,8 @@
   border-left: 1px solid $text-03;
   background-color: $ui-01;
   overflow-y: auto;
-  height: 100vh;
-  z-index: 99999;
+  height: calc(100vh - 3rem);
+  z-index: 3;
```
mseaton, line 15: This is an unrelated issue, but has been bothering me for a while, in that the buttons in the Add to queue overlay are barely visible on the screen.  This adjustment here I copied from the appointments esm, which presumably had the same issue and was fixed there.  Yet another reason not to have multiple overlay implementations all over the place.

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+        },
+      );
+    },
+    [closePanel, mutate, visit, t, visitQueueNumberAttributeUuid],
+  );
+
+  return (
+    <div>
+      {isTablet && (
+        <Row className={styles.headerGridRow}>
```
mseaton, line 92: This form is really just a subset of the full visit-form, and uses the same extension slot to add in the queue fields.

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+            setIsSubmitting(false);
+            mutate();
+          }
+        },
+        (error) => {
+          let subtitle = error?.message;
+          const err = error?.responseBody?.error?.message;
+          if (err && err === '[queue.entry.duplicate.patient]') {
+            subtitle = t('patientAlreadyInQueue', 'Patient is already in the queue');
+          }
```
mseaton, line 76: This is my admittedly somewhat hacky solution to the problem presented in https://openmrs.atlassian.net/browse/O3-2624 . In future refactors, we would not get to this point, but would have pre-submission validation in place, but this is an incremental improvement to at least inform the user of the error, which is caught in back end validation with a specific error message returned.

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+        sortWeight,
+        queueLocation,
+        visitQueueNumberAttributeUuid,
+      ).then(
+        ({ status }) => {
+          if (status === 201) {
+            showSnackbar({
+              kind: 'success',
+              isLowContrast: true,
+              title: t('startAVisit', 'Start a visit'),
```
ojwanganto, line 63: You can change this label to reflect the queue operation

ojwanganto:
Looks good, @mseaton. I have flagged just a simple label change.

donaldkibet (approved):

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
mseaton: Thanks @ojwanganto - looks like this PR was merged already by @makombe .  I tried to apply this change last night, but am running into build errors at the moment that is preventing me from committing.  @makombe maybe you can fix this for me since you merged the PR?  Otherwise, I'll address it in a follow-up commit once I have my environment back.

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  const config = useConfig() as ConfigObject;
+  const visitQueueNumberAttributeUuid = config.visitQueueNumberAttributeUuid;
+  const { mutate } = useVisitQueueEntries('', '');
+
+  if (!visit) {
+    return null;
+  }
+
+  const handleSubmit = useCallback(
```
chibongho, line 35: Don't really need useCallback here (see [this](https://react.dev/reference/react/useCallback#should-you-add-usecallback-everywhere)). It makes it difficult to maintain with the list of dependencies.

packages/esm-service-queues-app/src/patient-search/visit-form/existing-visit-form.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+import { type SearchTypes } from '../../types';
+import styles from './visit-form.scss';
+
+interface ExistingVisitFormProps {
+  toggleSearchType: (searchMode: SearchTypes, patientUuid) => void;
+  visit: Visit;
+  closePanel: () => void;
+}
+
+const ExistingVisitForm: React.FC<ExistingVisitFormProps> = ({ visit, toggleSearchType, closePanel }) => {
```
chibongho, line 22: Maybe add a comment here? Just from the name, it's hard to know that it's meant to be a form to start a queue entry for a patient that already has an existing visit .

chibongho (commented):
LGTM. Just a few minor comments.

PR #1036, opened 2024-03-14
'(fix)O3-2473': Visit Type selector fails and is not needed

jwnasambu:
## Requirements

## Summary
I modify the code to remove the search by default and allow it to be toggled back on with a configuration property by following these steps:

1. Added a new prop enableSearch to control the visibility of the search functionality.
2. Updated the component's logic to conditionally render the search based on the value of enableSearch.
3. Provided a default value for enableSearch as false.
4. Allowed the search to be toggled back on by setting enableSearch to true via a configuration property.
By doing these, when you use <BaseVisitType />, by default, the search functionality will be disabled and If you want to enable it, you can set enableSearch to true.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/33891016/78080333-4f2e-43b4-ade6-21a033296eba

## Related Issue
https://openmrs.atlassian.net/browse/O3-2473

### DIFF
--- a/packages/esm-service-queues-app/src/patient-search/visit-form/base-visit-type.component.tsx
+++ b/packages/esm-service-queues-app/src/patient-search/visit-form/base-visit-type.component.tsx
@@ -12,9 +12,10 @@ interface BaseVisitTypeProps {
   onChange: (event) => void;
   patientUuid: string;
   visitTypes: Array<VisitType>;
+  enableSearch?: boolean;
 }
 
-const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes }) => {
+const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes, enableSearch = false }) => {
   const { t } = useTranslation();
   const isTablet = useLayoutType() === 'tablet';
   const [searchTerm, setSearchTerm] = useState<string>('');
@@ -39,7 +40,7 @@ const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes }) =
         [styles.tablet]: isTablet,
         [styles.desktop]: !isTablet,
       })}>
-      {results.length ? (
+      {enableSearch && (
         <>
           {isTablet ? (
             <Layer>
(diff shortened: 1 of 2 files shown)


jwnasambu:
@mseaton, @mogoodrich, @ojwanganto, @makombe kindly help me review my PR at your convinient time please!

mogoodrich:
Also added Chi Bong, since he is working a lot on queues now

packages/esm-service-queues-app/src/patient-search/visit-form/base-visit-type.component.tsx
```diff
@@ -12,9 +12,10 @@ interface BaseVisitTypeProps {
   onChange: (event) => void;
   patientUuid: string;
   visitTypes: Array<VisitType>;
+  enableSearch?: boolean;
 }
 
-const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes }) => {
+const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes, enableSearch = false }) => {
```
mseaton, line 18: I think the `enableSearch` is still a bit of a question mark.  Personally, I see no reason to have a search here.  But let's just leave the search as-is for now and we can assess more in the future.

packages/esm-service-queues-app/src/patient-search/visit-form/base-visit-type.component.tsx
```diff
@@ -56,19 +57,21 @@ const BaseVisitType: React.FC<BaseVisitTypeProps> = ({ onChange, visitTypes }) =
...
-            name="radio-button-group"
-            valueSelected={results?.length >= 1 && results[0].uuid}>
-            {results.map(({ uuid, display, name }) => (
-              <RadioButton key={uuid} className={styles.radioButton} id={name} labelText={display} value={uuid} />
-            ))}
-          </RadioButtonGroup>
         </>
+      )}
+
+      {results.length ? (
```
mseaton, line 63: This looks right - moving the search conditional around the results table, but _not_ around the actual search box itself.

mseaton (changes requested):

sam-sny (approved):

packages/esm-service-queues-app/src/patient-search/visit-form/base-visit-type.component.tsx
jwnasambu: @mseaton Kindly could you please provide more insight into the reasons behind this opinion? Let's maintain the search feature as it is for now, and we can reassess its relevance and effectiveness in the future? I am a bit confused on the changes to be made basing on the review.

mseaton: @jwnasambu - issue with the search getting hidden is a real bug.  The issue with the search not being necessary is an opinion I am expressing, but not really a problem.  Given that the designs have the search, we can leave it for now and maybe start a conversation on the ux-design-advisory asking about the thinking behind the visit type search.  It would be good to make sure we fully understand what use case this is envisioned for (eg. are there implementations out there that have so many visit types we would need an auto-complete search for them;  who/what implementation is this designed for where this is a need?

mseaton: Beyond this, if we _are_ going to have this boolean `enableSearch` prop on the component, then we will need to also support this configurability in the config schema and ensure the configured value is getting passed in and used when this component is instantiated.  I don't see that in this PR, and it's not really priority for us to have this removed, so this is why I said just to leave it.

brandones:
Fiona tested and closed this ticket.

PR #1043, opened 2024-03-15
(fix) service queues - unescape queue name when passed into t()

chibongho:
## Requirements

## Summary

minor fix to tell the `t()` function not to escape the queue name we pass in. Prior to the fix, `/` character in the queue name got incorrectly HTML-escaped 

I verified that this isn't a data /escaping issue elsewhere by seeing in the debugger that the `display` value of the queue is correctly set to "Pre-Op/PACU" prior to being passed to the `t()` function.

## Screenshots

before:
image

after:
image


mogoodrich (approved):
LGTM, thanks @chibongho !

chibongho:
> LGTM, but wouldn't we always want this to be the case? i.e. shouldn't this be the way the t function works by default?

@mseaton I was surprised that the `t()` function from `useTranslation()` does HTML escaping since React already does that to interpolated strings. That said, I still think default-escaping is still sensible (if annoying). Changing that globally is too dangerous.

PR #1053, opened 2024-03-20
(fix) add `@openmrs/esm-patient-common-lib` as peer dependency and register appointment workspace correctly.

donaldkibet:
## Requirements

## Summary
This PR moves `registerWorkspace` call outside of `startupApp` function. In addition to this I have added `@openmrs/esm-patient-common-lib` as a peer dependency. This is to fix an error when trying to launch `appointment-form-workspace` from the form-engine. cc @brandones @denniskigen @ibacher  referencing [o3 docs](https://o3-docs.openmrs.org/docs/workspaces/creating-workspaces.en-US#registering-a-workspace) should the registration for workspace happen inside `startupApp` function?

## Error
Screenshot 2024-03-20 at 18 34 51

### form schema used
```json
{
  "questions": [
    {
      "label": "Appointments",
      "required": false,
      "id": "labsWorkspaceLauncher",
      "questionOptions": {
        "rendering": "workspace-launcher",
        "buttonLabel": "Schedule appointments",
        "workspaceName": "appointments-form-workspace"
      }
    }
  ]
}
```
## Fix
https://github.com/openmrs/openmrs-esm-patient-management/assets/28008754/a1b5a5f4-57c4-4d36-a367-d31e26a75afb

### DIFF
--- a/packages/esm-appointments-app/package.json
+++ b/packages/esm-appointments-app/package.json
@@ -44,6 +44,7 @@
   },
   "peerDependencies": {
     "@openmrs/esm-framework": "5.x",
+    "@openmrs/esm-patient-common-lib": "7.x",
     "react": "18.x",
     "react-i18next": "11.x",
     "react-router-dom": "6.x",
--- a/packages/esm-appointments-app/src/index.ts
+++ b/packages/esm-appointments-app/src/index.ts
@@ -51,12 +51,6 @@ export function startupApp() {
   defineExtensionConfigSchema('cancelled-appointments-panel', cancelledAppointmentsPanelConfigSchema);
   defineExtensionConfigSchema('early-appointments-panel', earlyAppointmentsPanelConfigSchema);
 
-  registerWorkspace({
-    name: 'appointments-form-workspace',
-    load: getAsyncLifecycle(() => import('./form/appointments-form.component'), options),
-    title: translateFrom(moduleName, 'createNewAppointment', 'Create new appointment'),
-  });
-
   registerBreadcrumbs([
     {
       title: 'Appointments',
@@ -109,3 +103,9 @@ export const patientAppointmentsCancelConfirmationDialog = getAsyncLifecycle(
   () => import('./patient-chart/patient-appointments-cancel-modal.component'),
   options,
 );
+
+registerWorkspace({
+  name: 'appointments-form-workspace',
+  load: getAsyncLifecycle(() => import('./form/appointments-form.component'), options),
+  title: translateFrom(moduleName, 'createNewAppointment', 'Create new appointment'),
+});
--- a/yarn.lock
+++ b/yarn.lock
@@ -2918,6 +2918,7 @@ __metadata:
     yup: "npm:^0.32.11"
   peerDependencies:
     "@openmrs/esm-framework": 5.x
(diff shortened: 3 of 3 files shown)


mogoodrich (approved):
Oops, sorry about missing the peer dependency @donaldkibet ... I had tested this worked in general, but did not test from within a form.

Defer @brandones  and @ibacher  about whether the registerWorkspace should live within the startup function.

ojwanganto:
cc @denniskigen

denniskigen:
> Defer @brandones  and @ibacher  about whether the registerWorkspace should live within the startup function.

I defer to them about this as well. It's something I've explicitly noted down after realizing the discrepancy elsewhere.

denniskigen (approved):

ojwanganto:
@denniskigen - please go ahead and create a ticket to track any needed changes to address the expected behavior. We can merge this in for now.

ojwanganto:
@denniskigen - we can track this [in this ticket](https://openmrs.atlassian.net/browse/O3-2979)

ibacher:
> whether the registerWorkspace should live within the startup function

It should be inside the startup function. While it _may_ work, this has two downsides:

1. It's additional code that's run when this module is loaded, which works around our attempts to lazy-load functionality as user's actually need it.
2. Executable code inside the main `index.ts` has a lot of caveats about the context in which its run. It's very easy for code like this to have unexpected side-effects which yield difficult-to-debug issues. This has a lot to do with how `index.ts` ends up in the bundle.

mogoodrich:
@ibacher  do we have any idea why it needs to be outside the startup function to work when attempting to open via a form schema (as reported above, I haven't tested myself)?  I'm guessing/assuming that it is because the module isn't loaded?  Does there need to be some sort of dependency between the form engine and the appointments app.

Linked ticket is here: https://openmrs.atlassian.net/browse/O3-2979

brandones:
Yeah this would happen because the appointments app has not been loaded by the time the launch button is clicked. I think this is because apps are not loaded until a page or extension that they have is requested. But we don't presently have a mechanism by which the form app could tell the appointment app that it needs something (or indeed by which the form app could even be aware that it is the appointment app that has what it needs).

I think this motivates a switch to workspaces being registered in the routes file, alongside extensions and pages.

Ironically, if my diagnosis here is right, then workspaces registered as extensions probably don't have this shortcoming. Maybe moving to `registerWorkspace` was a bad move and we should have gone straight to workspaces as a new core system alongside extensions and pages. If that's the case, sorry, my bad.

This is connected with [this ticket](https://openmrs.atlassian.net/browse/O3-1421) to bring the right-nav and workspace into core. This ticket would be a prerequisite to making workspaces a new core system.

ibacher:
@mogoodrich Personally, I think the extension-based registration is cleaner, but its probably not a big deal either way.

mogoodrich:
Cool, I may switch back to the extensoin-based way of doing this unless there are objections?

ojwanganto:
This is fine with me @mogoodrich. At the end of the day, all that matters is the right approach.

mogoodrich:
PR with the workspace change here: https://github.com/openmrs/openmrs-esm-patient-management/pull/1068

PR #1061, opened 2024-03-25
(fix) O3-2120: Move `OpenMRS ID` patient search identifier to `defaultIdentifierTypes`

usamaidrsk:
… config-schema

## Requirements

## Summary

This PR  moves `OpenMRS ID` identifier to `defaultIdentifierTypes` in the `config-schema.ts`

## Related Issue

[O3-2120](https://openmrs.atlassian.net/browse/O3-2120)

## Other

[O3-2120]: https://openmrs.atlassian.net/browse/O3-2120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -44,12 +44,11 @@ export const configSchema = {
...
-  },
-  defaultIdentifier: {
-    _type: Type.String,
-    _description: 'Identifer type to be displayed when no defaultIdentifierTypes are provided. Default is OpenMRS ID.',
-    _default: 'OpenMRS ID',
+    _default: [
+      'b4d66522-11fc-45c7-83e3-39a1af21ae0d',
+      'f85081e2-b4be-4e48-b3a4-7994b69bb101',
+      '05a29f94-c0ed-11e2-94be-8c13b969e334',
+    ],
```
mogoodrich, line 51: IMO we shouldn't be hardcoding any identifier uuids here but this should come from the ref app configuration, but I know there's not agreement on that (and that this is existing code, not a change introduced as part of this code)

mogoodrich (commented):
Thanks @usamaidrsk !  Defer to @denniskigen  and @ibacher  who may have a better sense of how this is all used, but generally looks good to me.

packages/esm-patient-search-app/src/config-schema.ts
ibacher: What are the other identifiers then?

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
```diff
@@ -123,13 +123,7 @@ const CompactPatientBanner = forwardRef<HTMLDivElement, CompactPatientBannerProp
...
                 <span className={styles.middot}>&middot;</span>
                 {config.defaultIdentifierTypes.length ? (
-                  <>
-                    {patientIdentifiers.length > 1 ? (
-                      <DefaultIdentifiers identifiers={patientIdentifiers} />
-                    ) : (
-                      <FallbackIdentifier patient={patients[index]} identifierName={config.defaultIdentifier} />
-                    )}
-                  </>
+                  <DefaultIdentifiers identifiers={patientIdentifiers} />
```
ibacher, line 126: What happens if the configuration still specifies no patient identifiers? It seems like we still need some fallback here.

packages/esm-patient-search-app/src/config-schema.ts
denniskigen: So, these first two identifiers are Palladium-specific. Those are the Patient Clinic Number and the NUPI number. They should be provided via distro configuration instead of being hardcoded in the Patient Search config schema. @donaldkibet, can @usamaidrsk proceed to remove them from here?

usamaidrsk: From the config file there is this comment, for the first two.
> These values correspond to the Patient Clinic Number and National Unique Patient Identifier (NUPI) identifier types respectively

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
usamaidrsk: @ibacher how do you suggest we handle this callback, given that from the previous implementations it was checking for a static string `OpenMRS ID` if it existed in the patient identifiers

```typescript
const FallbackIdentifier: React.FC<CustomIdentifierProps> = ({ patient, identifierName }) => {
  const identifier = patient.identifiers.find((identifier) => identifier.identifierType.display === identifierName);

  return identifier ? <IdentifierTag identifier={identifier} /> : null;
};
```

ibacher: @usamaidrsk Technically, it's checking for a user-configurable value, `identifierName`, right? I'm not sure there's a saner implementation than that. Maybe if that's also unset we could look for the `preferred` identifier?

usamaidrsk: > Technically, it's checking for a user-configurable value, identifierName, right?.

Yes it is a configurable value, which makes it not safe.

I do agree with the idea of the `preferred` identifier, how do we get to it
cc @ibacher

packages/esm-patient-search-app/src/config-schema.ts
donaldkibet: Thanks right @mogoodrich

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
ibacher: It's a property of `patient.identifiers` or available as one (if it's excluded via a custom representation).

usamaidrsk:
https://github.com/openmrs/openmrs-esm-patient-management/assets/53287480/a77f23d1-dfc6-4e25-b3a7-e6b86a1a8c36

usamaidrsk:
> @usamaidrsk Please update this and I'll review. I think it looks good.

@brandones I have updated the PR. please take a look again.

brandones (approved):
Looks great, thanks @usamaidrsk

PR #1065, opened 2024-03-28
Appointments: The appointments metrics aren't tally correctly

gabriel090:
## Requirements

## Screenshots

Appointments: The appointment metrics aren't tally correctly. The number of patients is not equal to the total of check-in and Not arrived
metricsNotTallying

This is the modified part
metricsTallying


packages/esm-appointments-app/src/metrics/appointments-metrics.component.tsx
```diff
@@ -45,7 +45,7 @@ const AppointmentsMetrics: React.FC<AppointmentMetricsProps> = ({ appointmentSer
       <div className={styles.cardContainer} data-testid="clinic-metrics">
         <MetricsCard
           label={t('patients', 'Patients')}
-          value={totalScheduledAppointments}
+          value={totalPatients}
```
ojwanganto, line 48: What does the appointment endpoint provide? I think this should directly come from the backend without any computation on the frontend. We expect to query the backend for scheduled and checked-in clients. We can get not checked-in numbers from the two.

PR #1066, opened 2024-03-28
(fix) O3-3009 Appointment form should limit location to those tagged …

mseaton:
…as 'Appointment Location'

## Requirements

## Summary
The appointment module creates / requires a specific Location tag to be associated with any Location at which an appointment can be scheduled.  This tag is already used within the appointments esm, but it is not used within the appointment form.  The result is that appointments can be scheduled for locations at which appointments are not supported, and the list of available locations is not filtered to the appropriate list of sub-locations.

## Related Issue
https://openmrs.atlassian.net/browse/O3-3009

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/form/appointments-form.component.tsx
@@ -42,7 +42,13 @@ import { useProviders } from '../hooks/useProviders';
 import Workload from '../workload/workload.component';
 import type { Appointment, AppointmentPayload, RecurringPattern } from '../types';
 import { type ConfigObject } from '../config-schema';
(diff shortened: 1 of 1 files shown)


mseaton:
> Also, not technically part of this ticket, but would be good to confirm whether the useLocations hook paginates it's results, because we really want to fetch _all_ appointments that match the criteria in this case

Yes, that would be a separate ticket, and something that would have an impact across the entire application for all types of Location use-cases.  For this ticket, the assumption we are making is that useLocations(tag) returns the appropriate set of locations for that tag.  But if there are cases where there are 50+ locations tagged as appointment locations, then this should be handled in a separate ticket to improve the useLocations hook.

PR #1069, opened 2024-03-29
(fix) O3-3013 Appointment Form should validate invalid time values

mseaton:
## Requirements

## Summary
If you enter an invalid time in the time field input when scheduling an appointment, nothing prevents the form submission, and the invalid time is discarded and the time is set to midnight on the appointment.  This is problematic, particularly since times like "11" might be entered, which are invalid, and are just silently saved as midnight.

## Screenshots
image

## Related Issue
https://openmrs.atlassian.net/browse/O3-3013

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/form/appointments-form.component.tsx
@@ -4,34 +4,33 @@ import dayjs from 'dayjs';
 import {
   Button,
   ButtonSet,
-  DatePickerInput,
   DatePicker,
+  DatePickerInput,
   Form,
   InlineLoading,
   MultiSelect,
   NumberInput,
+  RadioButton,
+  RadioButtonGroup,
   Select,
   SelectItem,
   Stack,
-  RadioButtonGroup,
-  RadioButton,
   TextArea,
-  TimePickerSelect,
   TimePicker,
+  TimePickerSelect,
   Toggle,
 } from '@carbon/react';
 import { Controller, useForm } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
 import { z } from 'zod';
 import {
+  ResponsiveWrapper,
+  showSnackbar,
+  useConfig,
+  useLayoutType,
   useLocations,
   useSession,
-  showSnackbar,
-  useLayoutType,
-  useConfig,
-  ResponsiveWrapper,
 } from '@openmrs/esm-framework';
-import { convertTime12to24 } from '@openmrs/esm-patient-common-lib';
 import {
   saveAppointment,
   saveRecurringAppointments,
@@ -51,8 +50,11 @@ import {
 } from '../constants';
 import styles from './appointments-form.scss';
 import SelectedDateContext from '../hooks/selectedDateContext';
-import uniqBy from 'lodash-es/uniqBy';
-import { useController, Control } from 'react-hook-form';
+
+const time12HourFormatRegexPattern = '(1[0-2]|0?[1-9]):[0-5][0-9]';
+function isValidTime(timeStr) {
+  return timeStr.match(new RegExp(time12HourFormatRegexPattern));
+}
 
 const appointmentsFormSchema = z.object({
   duration: z.number(),
@@ -66,7 +68,7 @@ const appointmentsFormSchema = z.object({
   recurringPatternPeriod: z.number(),
   recurringPatternDaysOfWeek: z.array(z.string()),
   selectedDaysOfWeekText: z.string().optional(),
-  startTime: z.string(),
+  startTime: z.string().refine((value) => isValidTime(value)),
   timeFormat: z.enum(['AM', 'PM']),
   appointmentDateTime: z.object({
     startDate: z.date(),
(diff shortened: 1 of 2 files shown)


packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -4,34 +4,33 @@ import dayjs from 'dayjs';
...
+  ResponsiveWrapper,
   showSnackbar,
-  useLayoutType,
   useConfig,
-  ResponsiveWrapper,
+  useLayoutType,
+  useLocations,
+  useSession,
 } from '@openmrs/esm-framework';
-import { convertTime12to24 } from '@openmrs/esm-patient-common-lib';
```
mseaton, line 34: The use of this function from patient-common-lib was a big part of the problem, as it is what is responsible for (silently) returning a value of "midnight" if the converted time fails to match the expected format.  This gets rid of the usage of this and adds the appropriate handling directly.

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -51,8 +50,11 @@ import {
 } from '../constants';
 import styles from './appointments-form.scss';
 import SelectedDateContext from '../hooks/selectedDateContext';
-import uniqBy from 'lodash-es/uniqBy';
-import { useController, Control } from 'react-hook-form';
+
+const time12HourFormatRegexPattern = '(1[0-2]|0?[1-9]):[0-5][0-9]';
+function isValidTime(timeStr) {
+  return timeStr.match(new RegExp(time12HourFormatRegexPattern));
+}
```
mseaton, line 57: Simple validation function to check the time format - used within the carbon field component to control whether the "invalid time" message appears, and within the form schema to control whether the form submission is allowed.

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -66,7 +68,7 @@ const appointmentsFormSchema = z.object({
   recurringPatternPeriod: z.number(),
   recurringPatternDaysOfWeek: z.array(z.string()),
   selectedDaysOfWeekText: z.string().optional(),
-  startTime: z.string(),
+  startTime: z.string().refine((value) => isValidTime(value)),
```
mseaton, line 71: This seems to be needed to prevent the form submission if the value fails validation

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -725,7 +729,9 @@ function TimeAndDuration({ isTablet, t, watch, control, services }) {
           render={({ field: { onChange, value } }) => (
             <TimePicker
               id="time-picker"
-              pattern="([\d]+:[\d]{2})"
+              pattern={time12HourFormatRegexPattern}
+              invalid={!isValidTime(value)}
+              invalidText={t('invalidTime', 'Invalid time')}
```
mseaton, line 734: I don't quite know what the `pattern` property here accomplishes.  It doesn't seem to have much effect.  The `invalid` and `invalidText` seem to be what is needed to display a validation error below the field if the input value is invalid.

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -289,7 +291,9 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
     } = data;
 
     const serviceUuid = services?.find((service) => service.name === selectedService)?.uuid;
-    const [hours, minutes] = convertTime12to24(startTime, timeFormat);
+    const hoursAndMinutes = startTime.split(':').map((item) => parseInt(item, 10));
+    const hours = (hoursAndMinutes[0] % 12) + (timeFormat === 'PM' ? 12 : 0);
+    const minutes = hoursAndMinutes[1];
```
mseaton, line 296: This logic is essentially what was in the `convertTime12to24` method without the erroneous handling of invalid values.

ojwanganto (approved):

packages/esm-appointments-app/src/form/appointments-form.component.tsx
mogoodrich: @ibacher @denniskigen  assumedly we should ticket and fix the convertTime12to24 as well, unless for some reason that behavior is expected?

packages/esm-appointments-app/src/form/appointments-form.component.tsx
mogoodrich: fwiw, the pattern attribute is explained here: https://react.carbondesignsystem.com/?path=/docs/components-timepicker--overview

On quick glance, it seems like it is *supposed* to do what the invalid function is doing?  I have found that React Carbon elements don't have the greatest documentation (in this case, perhaps the pattern works in conjunction with warning, just a guess?) so if we can't get this to work and are relying on invalid, probably just remove the pattern attribute entirely to avoid future confusion.

mogoodrich (approved):
Thanks @mseaton  a couple comments, but LGTM

packages/esm-appointments-app/src/form/appointments-form.component.tsx
chibongho: `pattern` seems to be what the browser uses to check pattern on form submission. https://www.w3schools.com/TAGS/att_input_pattern.asp

I'm guess `invalid` can be set whenever, and doesn't need to be tied to form submission action.

chibongho (approved):
Looks good. A bit unnerving that `convertTime12to24` defaults to 00:00 silently on error...

PR #1079, opened 2024-04-03
(feat) O3-3030 - make new queue table support creating new queue entry

chibongho:
## Requirements

## Summary

This PR adds the ability to add queue entries via the new queue table UI. 

I also refactored and deprecated the `useVisitQueueEntry` hook in favor of the `useQueueEntry` hook. As `/queue-entry` can already return the same values (albeit with different schema), we should be able to consolidate their usage. I tested that the `<ActiveVisitTable>` load queue entries fine with and without the feature flag.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/509602/316219e4-ed3c-460a-bdfc-b20bce6483d2


packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
```diff
@@ -102,21 +100,35 @@ interface PaginationData {
...
+  }, [error?.message]);
+
   if (isLoading) {
     return <DataTableSkeleton role="progressbar" />;
   } else if (useNewActiveVisitsTable) {
-    const queueEntriesInSelectedQueue = queueEntries.filter(
-      (queueEntry) => !selectedServiceUuid || queueEntry.queue.uuid == selectedServiceUuid,
-    );
+    const queueEntriesInSelectedQueue =
+      queueEntries?.filter((queueEntry) => !selectedQueueUuid || queueEntry.queue.uuid == selectedQueueUuid) ?? [];
```
mseaton, line 131: Is this still needed?  Aren't the queueEntries already filtered by selectedQueueUuid at this point, based on the parameters sent to `useQueueEntries`?  Can we get rid of this filtering?

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
```diff
@@ -40,7 +38,11 @@ function ClinicMetrics() {
       return true;
     }
   });
-  const { visitQueueEntriesCount } = useVisitQueueEntries(currentServiceName, currentQueueLocation);
+  const { queueEntries, totalCount } = useQueueEntries({
```
mseaton, line 41: So I think it would be more efficient to use the `queue-entry-metric?metric=count` endpoint to get the `totalCount` if you don't actually need the queueEntries (unless we are already retrieving them on the same page and they are cached in swr anyway).

mseaton (approved):
Just a few quick suggestions but nothing major - looks great overall, thanks!

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
chibongho: Yeah, they are being retrieved with swr elsewhere already.

packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
chibongho: Yeah, I think so. Good catch.

PR #1084, opened 2024-04-05
O3-2685: Inconsistent / Duplicated Appointments List UI between Home App and Appointments App

mogoodrich:
https://openmrs.atlassian.net/browse/O3-2685

## Requirements


mogoodrich:
[Screencast from 04-05-2024 10:19:08 AM.webm](https://github.com/openmrs/openmrs-esm-patient-management/assets/1163015/edff1929-7cdc-4019-b253-345a8673476a)

This removing almost of the "home appointments" code in favor of uisng the table used within the appointments app.  

We can continue to iterate on what functionality/actions we want within this appointments table, while maintaining a consistency between the two views.

Wanted to get this up for review before I go on vacation, but once small nit I've noticed is the alignment of the Download button is now slightly off.

Also @denniskigen after the recent linting changes I feel like this code didn't get prettified/linted fully before I pushed it up?  Are there new steps (other than yarn prettier) to run now to format the code?  Thanks!

packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx
```diff
@@ -69,6 +69,10 @@ const AppointmentsTable: React.FC<AppointmentsTableProps> = ({
       header: t('identifier', 'Identifier'),
       key: 'identifier',
     },
+    {
+      header: t('location', 'Location'),
+      key: 'location',
+    },
```
mogoodrich, line 75: The location column was present in the Today's view but not the Appointments view, now it is in both.  I noticed that in the Appointments view, we *don't* currently filter by session location?  Not sure if the intent was to filter in that view, and therefore the location column would be redundant?  If we do implement that, we can just configure this table to hide this column as needed.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
```diff
@@ -18,7 +18,6 @@ dayjs.extend(isToday);
 
 interface AppointmentActionsProps {
   appointment: Appointment;
-  scheduleType: string;
```
mogoodrich, line 21: Not sure this scheduleType property was being properly used or if it was even needed in the long run.  I think the main usage was to render a "Add appointments" button in the "Today's Appointments" view.  We could bring that back if necessary.

packages/esm-appointments-app/src/appointments/common-components/appointments-table.scss
```diff
@@ -1,5 +1,8 @@
 @use '@carbon/styles/scss/spacing';
-@use '@carbon/styles/scss/type';
+@use '@carbon/colors';
+@use '@carbon/layout';
+@use '@carbon/type';
+@import '~@openmrs/esm-styleguide/src/vars';
```
mogoodrich, line 5: All this scss is existing scss that I moved around as needed as I moved some divs and containers around.  My intent was only to copy over scss that was used.

packages/esm-appointments-app/src/appointments/common-components/appointments-table.scss
```diff
@@ -16,3 +19,61 @@
...
+.emptyStateContent {
+  @include type.type-style('heading-compact-01');
+  color: $text-02;
+  margin-top: spacing.$spacing-05;
+  margin-bottom: spacing.$spacing-03;
+}
+
+
+// Overriding styles for RTL support
+html[dir='rtl'] {
```
mogoodrich, line 68: I have no idea what RTL support is...

packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
```diff
@@ -2,14 +2,15 @@ import React from 'react';
...
-interface AppointmentsByStatusProps {
+interface AppointmentsListProps {
   appointmentServiceType?: string;
-  status: string;
+  status?: string;
   title: string;
   date: string;
 }
-const AppointmentsByStatus: React.FC<AppointmentsByStatusProps> = ({ appointmentServiceType, status, title, date }) => {
+const AppointmentsList: React.FC<AppointmentsListProps> = ({ appointmentServiceType, status, title, date }) => {
```
mogoodrich, line 13: Not sure if "AppointmentsList" is the best name, but I renaming from "AppointmentsByStatus" because this will now fetch *all* appointments on a given day if you don't pass in a status, so the existing name seemed confusing.

packages/esm-appointments-app/src/empty-state/empty-state.component.tsx
```diff
@@ -27,17 +26,6 @@ export const EmptyState: React.FC<EmptyStateProps> = ({ headerTitle, displayText
...
-        {scheduleType === 'Scheduled' ? (
-          <p className={styles.action}>
-            {launchForm && (
-              <span>
-                <Button renderIcon={Add} kind="ghost" onClick={() => launchForm()}>
-                  {t('createAppointment', 'Create appointment')}
-                </Button>
-              </span>
-            )}
-          </p>
```
mogoodrich, line 39: This is the aforemented logic that rendered a "Create appointment" button, but only if there were no scheduled appointments on the current day.  I could see why this is "nice" functionality to have, but we already have a create appointment button on this page.

packages/esm-appointments-app/src/home/index.tsx
```diff
@@ -1,9 +1,22 @@
...
+import Overlay from '../overlay.component';
 
 const HomeAppointments = () => {
-  return <AppointmentsBaseTable />;
+  const { t } = useTranslation();
+
+  return (
+    <div className={styles.container}>
+      <AppointmentsList
+        date={dayjs().startOf('day').format('YYYY-MM-DDTHH:mm:ss.SSSZZ')}
```
ibacher, line 14: Hmmm... We should probably make the `isoFormat` constant importable... and change [`DateInput`](https://github.com/openmrs/openmrs-esm-core/blob/3ce509b6b4b78e070c829793f91c5e995ad88d9c/packages/framework/esm-utils/src/omrs-dates.ts#L19) to indicate that it can take a `dayjs.Dayjs` object (which would allow just calling `toOmrsIsoString(dayjs().startOf('day'))`.

ibacher (approved):
A few remarks, but LGTM.

packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx
```diff
@@ -107,93 +112,96 @@ const AppointmentsTable: React.FC<AppointmentsTableProps> = ({
   if (!appointments?.length) {
     return (
       <EmptyState
-        headerTitle={`${tableHeading} appointments`}
-        displayText={`${tableHeading.toLowerCase()} appointments`}
+        headerTitle={`${startCase(t(tableHeading))} ${t('appointments', 'appointments')}`}
+        displayText={`${t(tableHeading).toLowerCase()} ${t('appointments', 'appointments').toLowerCase()}`}
```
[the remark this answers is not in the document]
ibacher: I'm also not a huge fan of splitting words like this. The issue is that this works fine for "Today's Appointments", but plays merry hell with "Les rendez-vous du jour". It's also worth pointing out that `startCase()` is simply incorrect here: `rendez-vous` becomes `Rendez Vous`.

packages/esm-appointments-app/src/home/index.tsx
mogoodrich: makes sense, for now I just changed to:
toOmrsIsoString(dayjs().startOf('day').toDate()).

packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx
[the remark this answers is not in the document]
mogoodrich: Fair points @ibacher .  I just removed all the casing stuff entirely for now, I think it's fine without them.  As for the splitting words, I tried to clean that up, but unfortunately it was a bit of a rabbit hold, as the "tableHeading" is potentially an extension parameter, and I didn't want to fiddle with those at this point.  But this is something we should clean up, I will ticket.

denniskigen:
Formatting should work just as well as before the changes, @mogoodrich. If not, then there's a problem. Is your default formatter set to Prettier? When you press `Ctrl + Shift + P` and type `Format document with`, is `Prettier` set as the default?

packages/esm-appointments-app/src/appointments/common-components/appointments-table.scss
denniskigen: Yeah, we support translating content into Arabic and Hebrew.

packages/esm-appointments-app/src/appointments/common-components/appointments-table.scss
denniskigen: We don't need both `@carbon/styles/scss/spacing` and `@carbon/layout`. I'd prefer keeping `@carbon/layout` and refactoring tokens using the `spacing` namespace to use `layout` instead. So `spacing.$spacing-x` changes to `layout.$spacing-x`.

denniskigen (approved):
LGTM! Thanks @mogoodrich.

mogoodrich:
Thanks @denniskigen  @ibacher  I will merge once the tests pass... this is a big change and I'm on vacation next week, so maybe I shouldn't be merging this. Feel free to revert with prejudice is needed.  :)

Also @denniskigen  I don't all the linting/prettier checks ran properly for me, but I can't figure out how to tell lint/prettier to run against all files (not just those that are staged)

denniskigen:
Just looking at it again, @mogoodrich, and there's a problem. Thanks for flagging it. I'll push a fix tomorrow.

mogoodrich:
Awesome, thanks @denniskigen !

packages/esm-appointments-app/src/appointments/common-components/appointments-table.scss
mogoodrich: just saw this, will handle in a post-merge commit

denniskigen:
Noticed two visual regressions @mogoodrich:

- The translation string for the empty state is broken.
- The border around the empty state doesn't look right - looks like it tacks on another border on top of what the `EmptyState` component provides from Common Lib.

CleanShot 2024-04-06 at 7  25 25@2x

mogoodrich:
Thanks @denniskigen ! I can take a look... for clarification, it the follow on commit from here a fix for this?

PR #1095, opened 2024-04-11
O3-3067: (fix) Incorrect parameters passed to react hook when filtering queues by service uuid

donaldkibet:
## Requirements

## Summary
PR description here https://openmrs.atlassian.net/browse/O3-3067

## Screenshots
https://github.com/openmrs/openmrs-esm-patient-management/assets/28008754/d2d167a0-ceb0-4320-b2f4-9e6f35c7afd1

### DIFF
--- a/packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
+++ b/packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
@@ -68,6 +68,7 @@ import DefaultQueueTable from '../queue-table/default-queue-table.component';
 import { type QueueEntry } from '../types';
 import { mapVisitQueueEntryProperties, type MappedVisitQueueEntry } from './active-visits-table.resource';
 import styles from './active-visits-table.scss';
+import uniqBy from 'lodash-es/uniqBy';
 
 /**
  * FIXME Temporarily moved here
@@ -92,10 +93,10 @@ interface PaginationData {
 }
 
 function ActiveVisitsTable() {
-  const selectedServiceQueueUuid = useSelectedServiceUuid();
+  const selectedServiceUuid = useSelectedServiceUuid();
   const currentLocationUuid = useSelectedQueueLocationUuid();
   const { queueEntries, isLoading, error } = useQueueEntries({
-    service: selectedServiceQueueUuid,
+    service: selectedServiceUuid,
     location: currentLocationUuid,
     isEnded: false,
   });
@@ -126,12 +127,12 @@ function OldQueueTable({ queueEntries }: { queueEntries: QueueEntry[] }) {
   const { t } = useTranslation();
   const currentServiceName = useSelectedServiceName();
   const currentQueueLocation = useSelectedQueueLocationUuid();
-  const { queues } = useQueues(currentQueueLocation);
   const { visitQueueNumberAttributeUuid } = useConfig<ConfigObject>();
   const currentServiceUuid = useSelectedServiceUuid();
-  const visitQueueEntries = queueEntries
-    .map((entry) => mapVisitQueueEntryProperties(entry, visitQueueNumberAttributeUuid))
-    .filter((entry) => (currentServiceUuid ? entry.queue.service.uuid === currentServiceUuid : true));
+  const visitQueueEntries = queueEntries.map((entry) =>
+    mapVisitQueueEntryProperties(entry, visitQueueNumberAttributeUuid),
+  );
+  const queues = uniqBy(queueEntries?.flatMap((entry) => entry.queue) ?? [], 'uuid');
 
   const [showOverlay, setShowOverlay] = useState(false);
   const [viewState, setViewState] = useState<{ selectedPatientUuid: string }>(null);
@@ -199,8 +200,11 @@ function OldQueueTable({ queueEntries }: { queueEntries: QueueEntry[] }) {
   );
 
   const handleServiceChange = ({ selectedItem }) => {
-    updateSelectedServiceUuid(selectedItem.uuid);
-    updateSelectedServiceName(selectedItem.display);
+    const {
+      service: { display, uuid },
+    } = selectedItem ?? {};
+    updateSelectedServiceUuid(uuid);
+    updateSelectedServiceName(display);
   };
 
   const tableRows = useMemo(() => {
--- a/packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
+++ b/packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
@@ -92,10 +92,10 @@ interface PaginationData {
 }
 
 function ActiveVisitsTable() {
-  const selectedQueueUuid = useSelectedServiceUuid();
+  const selectedServiceQueueUuid = useSelectedServiceUuid();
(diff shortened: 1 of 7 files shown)


packages/esm-service-queues-app/src/add-provider-queue-room/add-provider-queue-room.component.tsx
```diff
@@ -61,7 +61,7 @@ const AddProviderQueueRoom: React.FC<AddProviderQueueRoomProps> = ({ providerUui
 
   const { mutate } = useProvidersQueueRoom(providerUuid);
   const { queues } = useQueues(currentLocationUuid);
-  const { rooms } = useQueueRooms(currentLocationUuid, currentServiceUuid);
+  const { rooms } = useQueueRooms(currentLocationUuid, currentLocationUuid);
```
ojwanganto, line 64: we are passing the same parameter. Is there a possibility to filter by service as well?

ojwanganto (changes requested):
Thanks @donaldkibet. I find this good already. I have left just one comment to be addressed.

ojwanganto:
image
I just got this in dev3

chibongho (changes requested):
We have had a bad naming for `updateSelectedServiceUuid` and `updateSelectedServiceName` to be set to queueUuid / queueName instead of serviceUuid / serviceName, This PR addresses that, which is great.

However, if we go with this, we must update all places that call `updateSelectedServiceUuid` to take in the serviceUuid instead of queueUuid. There are places missed in this PR. Also, we should change `updateSelectedServiceName` to also take in the service name instead of queue name.

It might be less dangerous to just rename `useSelectedServiceUuid` / `updateSelectedServiceName` to `updateSelectedQueueUuid` / `updateSelectedQueueName`. If we want to be selecting on a particular queue, I think selecting by queueUuid is preferable to selecting by serviceUuid (even if there is a one-to-one relationship between queue and service). Then fix the queue-entry-metrics call to use the right params.

packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
```diff
@@ -92,10 +92,10 @@ interface PaginationData {
 }
 
 function ActiveVisitsTable() {
-  const selectedQueueUuid = useSelectedServiceUuid();
+  const selectedServiceQueueUuid = useSelectedServiceUuid();
```
mseaton, line 95: Let's name this variable `selectedServiceUuid`, not `selectedServiceQueueUuid`

packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
```diff
@@ -128,11 +128,11 @@ function OldQueueTable({ queueEntries }: { queueEntries: QueueEntry[] }) {
...
   const { visitQueueNumberAttributeUuid } = useConfig<ConfigObject>();
-  const visitQueueEntries = queueEntries.map((entry) =>
-    mapVisitQueueEntryProperties(entry, visitQueueNumberAttributeUuid),
-  );
-
   const currentServiceUuid = useSelectedServiceUuid();
+  const visitQueueEntries = queueEntries
+    .map((entry) => mapVisitQueueEntryProperties(entry, visitQueueNumberAttributeUuid))
+    .filter((entry) => (currentServiceUuid ? entry.queue.service.uuid === currentServiceUuid : true));
+
```
mseaton, line 135: I wouldn't think this filter is needed at all.  We should just be able to use `queueEntries` as-is, since it is already filtering by both the selected service and location.  So let's get rid of this `visitQueueEntries` and just use the `queueEntries` returned from the hook.

mseaton (changes requested):
See comments.

donaldkibet:
> We have had a bad naming for `updateSelectedServiceUuid` and `updateSelectedServiceName` to be set to queueUuid / queueName instead of serviceUuid / serviceName, This PR addresses that, which is great.
> 
> However, if we go with this, we must update all places that call `updateSelectedServiceUuid` to take in the serviceUuid instead of queueUuid. There are places missed in this PR. Also, we should change `updateSelectedServiceName` to also take in the service name instead of queue name.
> 
> It might be less dangerous to just rename `useSelectedServiceUuid` / `updateSelectedServiceName` to `updateSelectedQueueUuid` / `updateSelectedQueueName`. If we want to be selecting on a particular queue, I think selecting by queueUuid is preferable to selecting by serviceUuid (even if there is a one-to-one relationship between queue and service). Then fix the queue-entry-metrics call to use the right params.

@chibongho i have updated the function to pick correct parameters

packages/esm-service-queues-app/src/active-visits/active-visits-table.component.tsx
```diff
@@ -126,13 +127,13 @@ function OldQueueTable({ queueEntries }: { queueEntries: QueueEntry[] }) {
   const { t } = useTranslation();
   const currentServiceName = useSelectedServiceName();
   const currentQueueLocation = useSelectedQueueLocationUuid();
-  const { queues } = useQueues(currentQueueLocation);
   const { visitQueueNumberAttributeUuid } = useConfig<ConfigObject>();
+  const currentServiceUuid = useSelectedServiceUuid();
   const visitQueueEntries = queueEntries.map((entry) =>
     mapVisitQueueEntryProperties(entry, visitQueueNumberAttributeUuid),
   );
+  const queues = uniqBy(queueEntries?.flatMap((entry) => entry.queue) ?? [], 'uuid');
```
chibongho, line 135: I think this will make the dropdown list display only queues that have at least one queue entry. Are we sure we want that? The order of those queues can be unpredictable based on the order of the returned queue entries.

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
```diff
@@ -26,6 +27,7 @@ function ClinicMetrics() {
 
   const currentQueueLocation = useSelectedQueueLocationUuid();
   const { queues } = useQueues(currentQueueLocation);
+  const uniqueQueues = unionBy(queues ?? [], ({ service }) => service?.uuid);
```
chibongho, line 30: I'm guessing we need this because there can be multiple queues for the same service?

packages/esm-service-queues-app/src/add-provider-queue-room/add-provider-queue-room.resource.ts
```diff
@@ -2,7 +2,7 @@ import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
 import useSWR from 'swr';
 import { type ProvidersQueueRoom, type QueueRoom } from '../types';
 
-export function useQueueRooms(location: string, queueUuid: string) {
+export function useQueueRooms(location: string, queueUuid: string = '') {
```
chibongho, line 5: This technically works, but we should use `null` to denote absence of a value instead of empty string.

chibongho (changes requested):
There are still places using the `updateSelectedServiceUuid()` function that needs updating, namely in `add-provider-queue-room.component.tsx` and `default-queue-table.component.tsx`

brandones:
Closing as stale.

PR #1101, opened 2024-04-17
 (feat) O3-3068 - service queues - remove ActiveVisitsTable and have its current columns and actions configurable

chibongho:
## Requirements

## Summary

This PR deprecates the `<ActiveVisitsTable>` in the service queues app in favor of the more customizable `<DefaultQueueTable>`, while attempting to preserve all columns and actions that the current table supports.
- the `queueNumber` column is shown when the `visitQueueNumberAttributeUuid` config is set
- the current actions columns (with bell / transfer / Overflow menu) if the `defaultStatusConceptUuid` and `defaultTransitionStatus` configs are set.

Things not done yet:
- ~~slotting in the "Clear Queue" button~~ This is now hard-coded into the default queue table.
- ~~testing the `useShowProviderQueueRoomModal` hook, which opens the showProviderQueueRoomModal dialog~~ The modal is intended for users to persist their location / queue dropdown selection across app load. I think we can solve it by just persisting their dropdown selection directly into location storage.

## Screenshots

with `visitQueueNumberAttributeUuid`, `defaultStatusConceptUuid` and `defaultTransitionStatus` configured:
image

with `visitQueueNumberAttributeUuid`, `defaultStatusConceptUuid` and `defaultTransitionStatus` configured to null:
image


packages/esm-service-queues-app/src/active-visits/active-visits-row-actions.component.tsx
```diff
@@ -26,7 +26,7 @@ export const ActiveVisitRowActionsCell = ({ queueEntry }: QueueTableCellComponen
 };
 
 export const activeVisitActionsColumn: QueueTableColumn = {
-  headerI18nKey: '',
+  header: translateFrom('@openmrs/esm-service-queues-app', 'actions', 'Actions'),
```
mseaton, line 29: You should just be able to use the standard `t` function here, right?  translateFrom would be if the translation is coming from a different esm.

packages/esm-service-queues-app/src/add-provider-queue-room/add-provider-queue-room.resource.ts
```diff
@@ -74,3 +78,31 @@ export function useProvidersQueueRoom(providerUuid: string) {
     mutate,
   };
 }
+
```
mseaton, line 81: Is the below something that would typically be found in a resource?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -74,14 +74,6 @@ export const configSchema = {
...
       _default: '5242AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
     },
-    generalPatientNoteUuid: {
-      _type: Type.ConceptUuid,
-      _default: '165095AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
-    },
-    midUpperArmCircumferenceUuid: {
-      _type: Type.ConceptUuid,
-      _default: '1343AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
-    },
```
mseaton, line 84: Why are these getting deleted?  Are they unused?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -142,6 +134,11 @@ export const configSchema = {
     _default: '',
     _description: 'Custom label for patient chart button',
   },
+  useActiveVisitsTable: {
+    _type: Type.Boolean,
+    _default: false,
+    _description: 'If true, renders the ActiveVisitsTable instead of the default queue table',
+  },
```
mseaton, line 141: Can't this go away?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -158,7 +155,6 @@ export interface ConfigObject {
     heightUuid: string;
     weightUuid: string;
     respiratoryRateUuid: string;
-    midUpperArmCircumferenceUuid: string;
```
mseaton, line 161: Why are you removing this?  Seems unrelated to this PR

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -173,6 +169,7 @@ export interface ConfigObject {
   customPatientChartUrl: string;
   defaultFacilityUrl: string;
   customPatientChartText: string;
+  useActiveVisitsTable: boolean;
```
mseaton, line 172: This should no longer exist

packages/esm-service-queues-app/src/queue-table/cells/queue-table-action-cell.component.tsx
```diff
@@ -61,7 +61,7 @@ export function QueueTableActionCell({ queueEntry }: QueueTableCellComponentProp
 }
 
 export const queueTableActionColumn: QueueTableColumn = {
-  headerI18nKey: '',
+  header: translateFrom('@openmrs/esm-service-queues-app', 'actions', 'Actions'),
```
mseaton, line 64: Same comment as above, just use the `t` function (this goes for all of the other similar changes to other files in this PR, I'm not going to keep making the same comment).  Or if this needs to be like this, can you describe why?

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -22,40 +22,75 @@ import { queueTableStatusColumn } from './cells/queue-table-status-cell.componen
...
+
   const [showOverlay, setShowOverlay] = useState(false);
   const [viewState, setViewState] = useState<{ selectedPatientUuid: string }>(null);
 
+  useShowProviderQueueRoomModal();
+
+  const config = useConfig<ConfigObject>();
+  const { visitQueueNumberAttributeUuid, concepts } = config;
+  const { defaultStatusConceptUuid, defaultTransitionStatus } = concepts ?? ({} as any);
+
```
mseaton, line 66: The below is consistent with what we discussed.  Might want to add a todo that this is here for backwards-compatibility pending further design of the configuation

packages/esm-service-queues-app/src/active-visits/active-visits-row-actions.component.tsx
chibongho: I tried that, but unfortunately no. The `t` function is usually obtained from `useTranslation()`, which must be called from a react component. I don't think I'm using the [`translateFrom` function](https://github.com/openmrs/openmrs-esm-core/blob/13067210cba73a461e65345388a887186d9da2e5/packages/framework/esm-translations/src/index.ts#L39) as intended, but I don't see any other function that exposes the `i18n` object. 

An alternative would be to turn `QueueTableColumn` into a function that takes in the `t` function passed in from a react component, but I don't think that's clean either.

@ibacher @denniskigen is there a standard way to define translation strings outside of a react component? When I define them using `translateFrom()`, the strings also don't get picked up by the translation pre-commit git hook.

packages/esm-service-queues-app/src/add-provider-queue-room/add-provider-queue-room.resource.ts
chibongho: Yes, it's a typical pattern to define react hooks within *resource.ts. I try to consolidate hooks that are used in multiple places in hooks.ts, but I don't foresee this being used in more than one place.

packages/esm-service-queues-app/src/config-schema.ts
chibongho: Yes.

packages/esm-service-queues-app/src/config-schema.ts
chibongho: Yes, will change.

packages/esm-service-queues-app/src/active-visits/active-visits-row-actions.component.tsx
mseaton: Couldn't you just make the header cells react components, similar to the data cells?

ibacher: Our standard practice with situations like this has been to obtain the `t()` function inside a React component and then pass it in to the place that it's needed. The `t()` function itself isn't React-dependent but the `useTranslation()` hook and the contextual namespace is.

chibongho:
Ready for feedback now. Thanks for syncing up with me today @ojwanganto

packages/esm-service-queues-app/src/active-visits/active-visits-row-actions.component.tsx
```diff
@@ -1,12 +1,13 @@
-import React, { FC } from 'react';
+import React from 'react';
 import TransitionMenu from '../queue-entry-table-components/transition-entry.component';
 import { type QueueTableCellComponentProps, type QueueTableColumn } from '../types';
 import ActionsMenu from '../queue-entry-table-components/actions-menu.component';
 import EditMenu from '../queue-entry-table-components/edit-entry.component';
-import { useConfig } from '@openmrs/esm-framework';
+import { translateFrom, useConfig } from '@openmrs/esm-framework';
```
mseaton, line 6: Remove

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries-dialog.component.tsx
```diff
@@ -1,25 +1,25 @@
-import React, { useCallback, useState } from 'react';
-import { useTranslation } from 'react-i18next';
-import styles from './clear-queue-entries-dialog.scss';
 import { Button, ButtonSkeleton, ModalBody, ModalFooter, ModalHeader } from '@carbon/react';
 import { showSnackbar } from '@openmrs/esm-framework';
-import { batchClearQueueEntries } from './clear-queue-entries-dialog.resource';
-import { type MappedVisitQueueEntry } from '../active-visits/active-visits-table.resource';
+import React, { useCallback, useState } from 'react';
+import { useTranslation } from 'react-i18next';
```
mseaton, line 4: I've seen feedback on other PRs that React should be the top import

packages/esm-service-queues-app/src/home.component.tsx
```diff
@@ -1,19 +1,19 @@
 import React from 'react';
-import { type ConfigObject, useConfig } from '@openmrs/esm-framework';
-import ActiveVisitsTable from './active-visits/active-visits-table.component';
+import { useConfig, useFeatureFlag } from '@openmrs/esm-framework';
```
mseaton, line 2: useFeatureFlag is no longer needed here

packages/esm-service-queues-app/src/queue-entry-table-components/queue-duration.component.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import dayjs from 'dayjs';
-import styles from '../active-visits/active-visits-table.scss';
+import styles from '../queue-table/queue-table.scss';
```
mseaton, line 4: Is this scss even used?  Do we need this import?

packages/esm-service-queues-app/src/queue-table/cells/queue-table-action-cell.component.tsx
```diff
@@ -61,7 +61,11 @@ export function QueueTableActionCell({ queueEntry }: QueueTableCellComponentProp
 }
 
 export const queueTableActionColumn: QueueTableColumn = {
-  headerI18nKey: '',
+  HeaderComponent: () => {
+    const { t } = useTranslation();
+    return t('actions', 'Actions');
+  },
+  key: 'actions',
```
mseaton, line 68: I guess if there is no more efficient way to avoid this repetitive pattern...?

packages/esm-service-queues-app/src/queue-table/queue-table.component.tsx
```diff
@@ -76,8 +76,10 @@ function QueueTable({ queueEntries, queueTableColumns, ExpandedRow, tableFilter
               <TableHead>
                 <TableRow>
                   {ExpandedRow && <TableExpandHeader enableToggle {...getExpandHeaderProps()} />}
-                  {headers.map((header) => (
-                    <TableHeader {...getHeaderProps({ header })}>{header.header}</TableHeader>
+                  {headers.map((header: QueueTableColumn) => (
+                    <TableHeader {...getHeaderProps({ header })}>
+                      <header.HeaderComponent />
```
mseaton, line 81: So I know you created the HeaderComponent based on my previous feedback regarding translation, but what isn't clear is why you didn't just continue to do the translation here when the column is rendered? i.e. what was wrong with just having an i18nKey like you had before, and then running it through the translation system here?

chibongho: The translation pre-commit git hook looks at places where we call the `t` function with hard-coded strings to determine what keys to add/remove in the translation json files. It does not have the ability to recognize those keys when they were passed into the `t` function like we do currently.

We got away with doing that because the `<ActiveVisitsTable>` hard-coded most of the same translation keys anyway, but with that gone, we have to do it differently.

packages/esm-service-queues-app/src/home.component.tsx
chibongho: you're right, will remove.

packages/esm-service-queues-app/src/active-visits/active-visits-row-actions.component.tsx
ibacher: Even in a library, we can still require a `TFunction` as a parameter... Especially since the library would have no way of knowing the possible translation strings whatsoever (i.e., I don't even see how we could use the `translateFrom()` function).

mseaton (approved):
Looks good to me @chibongho - nice work!

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries.component.tsx
```diff
@@ -1,24 +1,31 @@
...
 }
 
-const ClearQueueEntries: React.FC<ClearQueueEntriesProps> = ({ visitQueueEntries }) => {
+/**
+ * Button to nd queue entries of all patients in a queue table and end their visits.
+ * TODO: Remove this button once we have a better way to end queue entries.
+ * @param param0
+ * @returns
+ * @deprecated
+ */
```
denniskigen, line 19:
```suggestion
/**
 * Button to end queue entries of all patients in a queue table and end their visits.
 * TODO: Remove this button once we have a better way to end queue entries.
 * @param param0
 * @returns
 * @deprecated
 */
```

packages/esm-service-queues-app/src/index.ts
```diff
@@ -139,10 +139,4 @@ export function startupApp() {
   registerBreadcrumbs([]);
 
   defineConfigSchema(moduleName, configSchema);
-
-  registerFeatureFlag(
-    'new-queue-table',
-    'New Queue Table',
-    'Use a newer implementation of the queue table in the home dashboard of the queues app.',
-  );
 }
```
denniskigen, line 142: We can safely remove the unused `registerFeatureFlag` import.

packages/esm-service-queues-app/src/queue-table/default-queue-table.test.tsx
```diff
@@ -106,17 +106,17 @@ describe('ActiveVisitsTable: ', () => {
     expect(john).toBeInTheDocument();
     expect(john).toHaveAttribute('href', 'someUrl');
 
-    const expectedColumnHeaders = [/name/, /priority/, /status/, /waitTime/];
+    const expectedColumnHeaders = ['Name', 'Priority', 'Coming from', 'Status', 'Queue', 'Wait time', 'Actions'];
```
denniskigen, line 109: nit: Regex matches might be preferable to avoid getting burned by changes in case.

denniskigen (approved):
Nice work, @chibongho! LGTM.

PR #1110, opened 2024-04-22
(fix) O3-3080-Remove Location dropdown

kb019:
## Requirements

## Summary
Removed location dropdown and related code.

## Screenshots
image

## Related Issue
(https://openmrs.atlassian.net/browse/O3-3080)

### DIFF
--- a/packages/esm-service-queues-app/src/patient-queue-header/patient-queue-header.component.tsx
+++ b/packages/esm-service-queues-app/src/patient-queue-header/patient-queue-header.component.tsx
@@ -1,36 +1,15 @@
-import React, { useCallback } from 'react';
+import React from 'react';
 import { useTranslation } from 'react-i18next';
 import { Calendar, Location } from '@carbon/react/icons';
-import { Dropdown } from '@carbon/react';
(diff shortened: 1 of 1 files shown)


ojwanganto:
Thank you. cc @chibongho

chibongho (changes requested):
Thanks @kb019. This ticket is a bit more involved. We need to remove the `useQueueLocations` function. All places that calls it need to be replaced with using the session location instead. See [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/18021e5690237d31c851d571c860a23caeb335d2/packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx#L39-L41) for example.

kb019:
> Thanks @kb019. This ticket is a bit more involved. We need to remove the `useQueueLocations` function. All places that calls it need to be replaced with using the session location instead. See [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/18021e5690237d31c851d571c860a23caeb335d2/packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.tsx#L39-L41) for example.

Sure @chibongho , will do it and update the PR.

PR #1111, opened 2024-04-24
O3-2396: Appointments: Refine logic for whether a patient is checked …

mogoodrich:
…in or not

## Requirements

### DIFF
--- a/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
@@ -6,7 +6,7 @@ import { Button } from '@carbon/react';
 import { TaskComplete } from '@carbon/react/icons';
 import { useTranslation } from 'react-i18next';
 import { navigate, showModal, useConfig } from '@openmrs/esm-framework';
-import { type Appointment } from '../../types';
+import { type Appointment, AppointmentStatus } from '../../types';
 import { type ConfigObject } from '../../config-schema';
 import { useTodaysVisits } from '../../hooks/useTodaysVisits';
 import CheckInButton from './checkin-button.component';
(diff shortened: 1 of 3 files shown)


mogoodrich:
see my comments here: https://openmrs.atlassian.net/browse/O3-2963 ... we can have broader discussions about making this change there.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
```diff
@@ -22,13 +22,9 @@ interface AppointmentsActionsProps {
...
   const { t } = useTranslation();
   const { checkInButton, checkOutButton } = useConfig<ConfigObject>();
-  const { visits, mutateVisit } = useTodaysVisits();
+  const { mutateVisit } = useTodaysVisits();
   const patientUuid = appointment.patient.uuid;
   const visitDate = dayjs(appointment.startDateTime);
-  const hasActiveVisitToday = visits?.some((visit) => visit?.patient?.uuid === patientUuid && visit?.startDatetime);
-  const hasCheckedOutToday = visits?.some(
-    (visit) => visit?.patient?.uuid === patientUuid && visit?.startDatetime && visit?.stopDatetime,
-  );
```
mseaton, line 31: As we discovered when looking at this today, this change can't be made in isolation of the changes to the checkout functionality, which currently will end the active visit.  Always ending the active visit makes sense when your checkout logic is based on the presence of an active visit.  But once you change the check-out logic to be independent of the visit status, you'll need to check that the visit is not already ended and/or prompt if the user wants to also close the visit, when you issue the modal dialog and perform the operation.

mogoodrich: Yeah, I likely want to tweak this.

mogoodrich:
Converting this to a draft, we don't want to merge this in until we properly are able to switch appointment actions via the default paths

PR #1114, opened 2024-04-26
(feat) - O3-3096 service queues - onfiguration-driven queue table

chibongho:
## Requirements

## Summary

This is a draft proposal of how to address the problem describe in [O3-3096](https://openmrs.atlassian.net/browse/O3-3096):

We need to configure the queue table in different contexts. For example, there can be a view of queue entries from all queues, a view of queue entries in a specific queue, or even a view of queue entries in a specific queue AND status. Depending on the context, we want to configure the queue table to display different columns. Additionally, the queue table usually provides a way to perform actions on each queue entry. The set of available actions could be specific to a queue, specific to a status, or specific to a queue+status combination.

The PR proposes a solution that allows us to specify `TableDefinition`s in the config. Each `TableDefinition` specifies a list of columns to use based on an `appliedTo` criteria, which can be applied to specific queue, a specific status, a specific queue+status combination, or (in the case when `appliedTo` not specified) to any table. 

We predefine a set of `QueueTableColumn`s that we should reasonably support (like name, priority, status) and how they should get rendered. Some of these columns can further be configured. (For example, this PR shows how we can configure the `patientId` column to display a particular `patientIdentifierType`.) We will also define another `QueueTableColumn` (not currently in this PR) that only renders an extension slot, allowing implementors to custom-define whatever column they want.

Also not currently in this PR yet, but we will also need more flexibility in how we define actions we can perform to each queue entry based on its current queue and status. We can predefined a set of `QueueEntryAction`s that we should reasonably support (like editing / transitioning / ending / undoing queue entries). Some of these actions can further be customized (For example, we can define a quick action that tranistions a queue entry to a particular status). In addition, we can also define another `QueueEntryAction` that only renders an extension slot, allowing implementors to custom-define whatever action they want. We can use a very similar approach that we used for configuring columns and apply it to `QueueEntryAction`s
 

## Screenshots

Table listing queue entries from multiple queues:
image

Table listing queue entries from a specific queue. Note that the columns rendered are different.
image

## Other

[O3-3096]: https://openmrs.atlassian.net/browse/O3-3096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -60,32 +50,15 @@ function DefaultQueueTable() {
...
   useShowProviderQueueRoomModal();
-
-  const config = useConfig<ConfigObject>();
-  const { visitQueueNumberAttributeUuid, concepts } = config;
-
-  // TODO: these two configs are here for backwards compatibility with the actions we show in each row.
-  // There might be futher changes pending future design of the config schema.
-  const { defaultStatusConceptUuid, defaultTransitionStatus } = concepts ?? ({} as any);
-
-  const columns = [
```
chibongho, line 71: This (and similar columns arrays) are no longer hard-coded.

packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts
```diff
@@ -0,0 +1,93 @@
...
+        );
+      }),
+    [tableDefinitions, queue, status],
+  );
+
+  const columns = tableDefinition?.columns?.map((column) => columnsMap.get(column));
+  return columns;
+}
+
+function getColumnFromDefinition(t: TFunction, columnDef: ColumnDefinition): QueueTableColumn {
```
chibongho, line 48: Big switch statement here to mapping columnType (defined in config) to actual columns that are used to render header and cells of a column in the DataTable. The switch statement enumerates all the predefined columns we support. Implementators wishing to have a custom column can use the `extension-column` columnType.

packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts
```diff
@@ -0,0 +1,93 @@
...
+import { queueTableWaitTimeColumn } from './queue-table-wait-time-cell.component';
+import { useMemo } from 'react';
+import { type ColumnDefinition, type ConfigObject } from '../../config-schema';
+import { useConfig } from '@openmrs/esm-framework';
+import { activeVisitActionsColumn } from '../../active-visits/active-visits-row-actions.component';
+
+// returns the columns to display for a queue table of a particular queue + status.
+// For a table displaying all entries of a particular queue, the status param should be null
+// For a table displaying all entries from all queues, both params should be null
+export function useColumns(queue: string, status: string): QueueTableColumn[] {
```
chibongho, line 18: Function to determine the columns to display for a queue table of a particular queue + status

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -1,9 +1,60 @@
-import { Type, restBaseUrl, validators } from '@openmrs/esm-framework';
+import { Type, validators } from '@openmrs/esm-framework';
 import vitalsConfigSchema, { type VitalsConfigObject } from './current-visit/visit-details/vitals-config-schema';
 import biometricsConfigSchema, {
   type BiometricsConfigObject,
 } from './current-visit/visit-details/biometrics-config-schema';
 
+export const defaultTablesConfig: TablesConfig = {
```
chibongho, line 7: Definition of default columns to show in the queue table of the service queues app main page.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -1,9 +1,60 @@
...
+    },
+    {
+      id: 'actions',
+      columnType: 'active-visits-actions-column',
+    },
+  ],
+  tableDefinitions: [
+    {
+      columns: ['patient-name', 'priority', 'comingFrom', 'status', 'wait-time', 'actions'],
+      appliedTo: { queue: '3113f164-68f0-11ee-ab8d-0242ac120002' },
```
chibongho, line 50: 2 different `TableDefinition`s here. The first one is for illustration purpose, to show that we can define the columns for a specific queue. The second one applies to all tables.

packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts
```diff
@@ -0,0 +1,93 @@
...
+  const { tablesConfig } = config;
+  const { columnDefinitions, tableDefinitions } = tablesConfig;
+
+  const columnsMap = useMemo(() => {
+    const map = new Map<string, QueueTableColumn>();
+    for (const columnDef of columnDefinitions) {
+      map.set(columnDef.id, getColumnFromDefinition(t, columnDef));
+    }
+    return map;
+  }, [columnDefinitions, t]);
```
mseaton, line 30: Do we need to get the `t` function here and pass it down, rather than just get it in the getColumnFromDefinition method directly?

packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts
```diff
@@ -0,0 +1,93 @@
...
+  const columns = tableDefinition?.columns?.map((column) => columnsMap.get(column));
+  return columns;
+}
+
+function getColumnFromDefinition(t: TFunction, columnDef: ColumnDefinition): QueueTableColumn {
+  const { id, header, columnType } = columnDef;
+
+  switch (columnType) {
+    case 'patient-name-column': {
+      return queueTableNameColumn(id, header ?? t('name', 'Name'));
```
mseaton, line 53: The custom `header` here should _also_ be run through the translation function, no?  Potentially also with another optional property on `columnDef` to indicate which module the header translation string is defined in?

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-id-cell.component.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { type PatientIdColumnConfig } from '../../config-schema';
+import { type QueueEntry, type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
+
```
mseaton, line 4: `Id` -> `Identifier` search/replace in this file

packages/esm-service-queues-app/src/types/index.ts
```diff
@@ -429,15 +428,22 @@ export interface QueueTableCellComponentProps {
   queueEntry: QueueEntry;
 }
 
-export type QueueTableColumn = (t: TFunction) => {
-  header: string; // header of the column, MUST be unique for each column in the queue table
+export type QueueTableColumn = {
+  key: string; // key used by DataTable, MUST be unique for each column in the queue table
+  header: string; // header of the column
```
mseaton, line 433: This supports a message code right?  Might want to indicate this can be a translation string or actual column value

packages/esm-service-queues-app/translations/en.json
```diff
@@ -172,9 +172,7 @@
   "program": "Program",
```
mseaton, line 172: We probably don't actually want these message strings to be removed, and there may be some we need to see added from the different built-in columns.  There is a trick here that you need to do to ensure that the build scripts behave correctly - you can see how this is done in the root component of patient-chart-app.  Look in the codebase for a comment that reads something like:  `DO NOT REMOVE THIS COMMENT` and `THE TRANSLATION KEYS AND VALUES USED IN THE COMMON LIB IS WRITTEN HERE`

mseaton (commented):
@chibongho looks great.  A few comments.  I see this is still in Draft - would be good to get a sense of what is remaining before moving it into a state that could be merged.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -157,17 +213,63 @@ export interface ConfigObject {
...
+  | { columnType: 'current-queue-column' }
+  | { columnType: 'wait-time-column' }
+  | { columnType: 'visit-start-time-column' }
+  | { columnType: 'actions-column' }
+  | { columnType: 'active-visits-actions-column' }
+  | { columnType: 'extension-column'; config: object }
+);
+
+export interface VisitAttributeQueueNumberColmnConfig {
+  visitQueueNumberAttributeUuid: string;
```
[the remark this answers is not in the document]
chibongho: This is the same name as config used in the appointments app and patient-chart. This is technically a new schema, but I think I should retain the same name for consistency.

chibongho:
> @chibongho looks great. A few comments. I see this is still in Draft - would be good to get a sense of what is remaining before moving it into a state that could be merged.

@mseaton There are a few more `QueueTableColumn`s currently marked as TODO that I'd like to define as part of this PR. They are relatively easy to do. The configuration of actions will be trickier and probably warrant a separate PR.

packages/esm-service-queues-app/translations/en.json
chibongho: I got around this by changing the `extract-translations` git hook to read in *.ts files.

chibongho:
This is now ready for review. Thanks.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-age-cell.component.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import dayjs from 'dayjs';
+import { type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
+
+export const QueueTableWaitTimeCell = ({ queueEntry }: QueueTableCellComponentProps) => {
```
mseaton, line 5: This should be renamed to `QueueTablePatientAgeCell`

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-age-cell.component.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const todaydate = dayjs();
+  const age = todaydate.diff(birthdate, 'years');
+
+  return <span>{age}</span>;
+};
+
+export const queueTablePatientAgeColumn: QueueTableColumnFunction = (key, header) => ({
+  key,
+  header,
+  CellComponent: QueueTableWaitTimeCell,
```
mseaton, line 16: Same, rename once above is renamed

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-identifier-cell.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { type QueueEntry, type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
+
+export const queueTablePatientIdentifierColumn: QueueTableColumnFunction = (
+  key,
+  header,
+  config: PatientIdentifierColumnConfig,
+) => {
+  const { identifierType } = config;
+
+  const getPatientId = (queueEntry: QueueEntry) => {
```
mseaton, line 12: `getPatientId` -> `getPatientIdentifier`

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-identifier-cell.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+    for (const identifier of queueEntry.patient.identifiers) {
+      if (identifier.identifierType?.uuid == identifierType) {
```
mseaton, line 14: We likely want to check beyond this.  If a patient has > 1 identifier of the same type, then we should prioritize a) any that are "preferred = true" first, then likely the most recently created second.  Ideally the API could give you this, but otherwise should be done here.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-age-cell.component.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import dayjs from 'dayjs';
+import { type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
+
+export const QueueTableWaitTimeCell = ({ queueEntry }: QueueTableCellComponentProps) => {
+  const birthdate = dayjs(queueEntry.patient.person.birthdate);
+  const todaydate = dayjs();
+  const age = todaydate.diff(birthdate, 'years');
```
mseaton, line 8: I have seen logic around O3 that attempts to standardize some of this in places.  Not sure if we want to try to use that here, or if that should be a different column type.  The logic I've seen is something like "if the patient is < 1 month old, show age in days, if the patient is < 2 years old show age in months, else show age in years".  Don't rely on that exactly, but it's something like that.  @ibacher / @denniskigen do you remember?

packages/esm-service-queues-app/src/queue-table/cells/queue-table-patient-identifier-cell.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+        return identifier.identifier;
+      }
+    }
+    return null;
+  };
+
+  const QueueTablePatientIdCell = ({ queueEntry }: QueueTableCellComponentProps) => {
```
mseaton, line 21: `QueueTablePatientIdCell` -> `QueueTablePatientIdentifierCell`

packages/esm-service-queues-app/src/queue-table/cells/queue-table-queue-cell.component.tsx
```diff
@@ -1,13 +1,14 @@
...
+import { type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
 
 // reprevents a column showing which queue a queue entry belongs to
 export const QueueTableQueueCell = ({ queueEntry }: QueueTableCellComponentProps) => {
   return <>{queueEntry.queue.display}</>;
 };
 
-export const queueTableQueueColumn: QueueTableColumn = (t) => ({
-  header: t('queue', 'Queue'),
+export const queueTableQueueColumn: QueueTableColumnFunction = (key, header) => ({
```
mseaton, line 9: This is really `queueName` right?  Let's rename the component to this

packages/esm-service-queues-app/src/queue-table/cells/queue-table-visit-start-time-cell.component.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { type QueueTableColumnFunction, type QueueTableCellComponentProps, type QueueEntry } from '../../types';
+
+export const queueTableVisitStartTimeColumn: QueueTableColumnFunction = (key, header) => {
+  function getVisitStartTime(queueEntry: QueueEntry) {
+    return queueEntry.visit.startDatetime;
```
mseaton, line 6: visit is theoretically nullable on queueEntry, so may want to make this a `?.`

packages/esm-service-queues-app/src/queue-table/cells/queue-table-visit-start-time-cell.component.tsx
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  const QueueTablePriorityCell = ({ queueEntry }: QueueTableCellComponentProps) => {
```
mseaton, line 9: Copy/pasted code needs to be renamed.

packages/esm-service-queues-app/translations/en.json
```diff
@@ -196,7 +197,6 @@
   "queueLocation": "Queue Location",
   "queueLocationRequired": "Queue location is required",
   "queueName": "Queue name",
-  "queueNumber": "Queue Number",
```
mseaton, line 199: Why is this getting removed?

mseaton (approved):
Looks great to me @chibongho .  I'm approving, but there are some minor changes to be made, mostly where you've copied/pasted code around and failed to rename everything.  I don't need to re-review assuming you clean up the comments indicated.

I would be good to get at least one or two reviews from some of the more experienced react developers

packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts
ibacher: There's also the framework's 	`translateFrom()` function, which might be easier than maintaining two `t` functions (since presumably the defaults should still come from this namespace).

packages/esm-service-queues-app/translations/en.json
chibongho: Turns out I forgot to have a case statement for this column. Thanks for catching.

PR #1116, opened 2024-04-29
(fix) O3-3080: Service queues - location dropdown improvements

mccarthyaaron:
## Requirements

## Summary
This tickets aims at tackling the requirements outlined in the linked issue.
1. This PR changes the label of the location dropdown from "View" to "Location".
2. This PR adds a configuration property in the config-schema which allows the user to limit the available queue locations in the location picker dropdown to the session location and its child locations if they exist.
3. The value of that configuration can be passed to the useQueuesLocations() hook. By default(without passing that value) to it, the hook will return all locations with a queue tag. But when passed that configuration value the hook will return the current session location and its children if they do exist as the available queue locations.
4. If the resulting number of queue locations is just one, the location dropdown will only have that location as an option with the "All" option elimintated. If the possible locations are more than one, then the "All" option is  included. Furthermore these locations are organised alphabetically based on their "name" property by using lodash.
5. The PR also changes the initial values for the location name and location uuid from being read from local storage to being read from session storage. The onChange hanlder for the location dropdown rather than merely updating state also saves the new values to session storage. 
6. Also the PR adds a  hook to test if a given location is a queue location. If the currentLocation is a queue location then it will be displayed as the default value of the location dropdown but if it's not then 'All' will be displayed.

## Related Issue
https://openmrs.atlassian.net/browse/O3-3080


vasharma05:
Closing in view of #1195

PR #1118, opened 2024-04-30
UHM-7439, add queue table metrics

cioan:
@chibongho and @mseaton , before I go and implement all those configurable statuses and the number of patients in each table, just wanted to make sure if the following components are the right components to implement? Thanks!

Screenshot 2024-04-30 at 10 27 11 AM

### DIFF
--- a/packages/esm-service-queues-app/src/queue-table/queue-table-metrics-card.component.tsx
+++ b/packages/esm-service-queues-app/src/queue-table/queue-table-metrics-card.component.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import classNames from 'classnames';
+import { useTranslation } from 'react-i18next';
+import { Layer, Tile } from '@carbon/react';
+import { ConfigurableLink } from '@openmrs/esm-framework';
+import styles from '../patient-queue-metrics/metrics-card.scss';
+
+interface QueueTableMetricsCardProps {
+  value: number | string;
+  headerLabel: string;
+  children?: React.ReactNode;
+}
+
+const QueueTableMetricsCard: React.FC<QueueTableMetricsCardProps> = ({ value, headerLabel, children }) => {
+  const { t } = useTranslation();
+
+  return (
+    <Layer
+      className={classNames(styles.container, {
+        [styles.cardWithChildren]: children,
+      })}>
+      <Tile className={styles.tileContainer}>
+        <div className={styles.tileHeader}>
+          <div className={styles.headerLabelContainer}>
+            <label className={styles.headerLabel}>{headerLabel}</label>
+            {children}
+          </div>
+        </div>
+        <div className={styles.link}>
+          <ConfigurableLink className={styles.link} to={''}>
+            {value}
+          </ConfigurableLink>
+        </div>
+      </Tile>
+    </Layer>
+  );
+};
+
+export default QueueTableMetricsCard;
--- a/packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
+++ b/packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
@@ -1,0 +1,21 @@
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import QueueTableMetricsCard from './queue-table-metrics-card.component';
+import styles from '../patient-queue-metrics/clinic-metrics.scss';
+
+function QueueTableMetrics() {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <div className={styles.cardContainer} data-testid="clinic-metrics">
+        <QueueTableMetricsCard value={'38'} headerLabel={'Total Patients'} />
+        <QueueTableMetricsCard value={'3'} headerLabel={'Waiting for Service'} />
+        <QueueTableMetricsCard value={'4'} headerLabel={'In Service'} />
+        <QueueTableMetricsCard value={'5'} headerLabel={'Waiting for Transfer'} />
+      </div>
+    </>
+  );
+}
+
+export default QueueTableMetrics;
--- a/packages/esm-service-queues-app/src/patient-queue-metrics/metrics-card.scss
+++ b/packages/esm-service-queues-app/src/patient-queue-metrics/metrics-card.scss
@@ -28,6 +28,10 @@
   margin-bottom: spacing.$spacing-03;
 }
 
+.centerValue {
+  text-align: center;
+}
+
 .headerLabel {
   @include type.type-style('heading-compact-01');
   color: $text-02;
--- a/packages/esm-service-queues-app/src/views/queue-tables-for-all-statuses.component.tsx
+++ b/packages/esm-service-queues-app/src/views/queue-tables-for-all-statuses.component.tsx
@@ -16,6 +16,7 @@ import { queueTableActionColumn } from '../queue-table/cells/queue-table-action-
 import PatientSearch from '../patient-search/patient-search.component';
 import { Add } from '@carbon/react/icons';
 import { Search } from '@carbon/react';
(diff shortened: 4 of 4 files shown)


packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from 'react';
```
mseaton, line 1: I'm not sure this needs to be in it's own component, it can always be refactored out as needed in the future.  I'd probably just include this directly in the page where it is used for now

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { useTranslation } from 'react-i18next';
+import QueueTableMetricsCard from './queue-table-metrics-card.component';
+import styles from '../patient-queue-metrics/clinic-metrics.scss';
+
+function QueueTableMetrics() {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <div className={styles.cardContainer} data-testid="clinic-metrics">
```
mseaton, line 11: All of the below are incorrect in the sense that:
a) The headers all need to be translation strings if they are hard-coded (eg. Total Patients)
b) The cards that relate to specific statuses need to be done dynamically by iterating over the queue, getting it's allowed statuses, and then displaying a card based on that queue/status combination

packages/esm-service-queues-app/src/queue-table/queue-table-metrics-card.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Layer, Tile } from '@carbon/react';
+import { ConfigurableLink } from '@openmrs/esm-framework';
+import styles from '../patient-queue-metrics/metrics-card.scss';
+
+interface QueueTableMetricsCardProps {
+  value: number | string;
+  headerLabel: string;
+  children?: React.ReactNode;
+}
+
```
mseaton, line 13: The style of card already present on the service-queues home page should serve as the example of how this should be styled.

mseaton (changes requested):

chibongho:
Overall idea looks good to me.

cioan:
Thanks @mseaton and @chibongho ! I've got the queue statuses to render dynamically along with the number of queue entries with each status:
Screenshot 2024-04-30 at 5 13 46 PM

Screenshot 2024-04-30 at 5 13 33 PM

Next, I will work on the styling to see if I could get it to look close to the designs. Thanks!

cioan:
@mseaton , can I merge in this PR? Thanks!

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+function QueueTableMetrics({ selectedQueue }: QueueTableMetricsProps) {
+  const { t } = useTranslation();
+
+  const allowedStatuses = selectedQueue.allowedStatuses;
+  const { queueEntries } = useQueueEntries({ queue: selectedQueue.uuid, isEnded: false });
+
+  return (
+    <div className={styles.metricsBorder}>
+      <QueueTableMetricsCard value={queueEntries.length} headerLabel={t('totalPatients', 'Total Patients')} />
```
chibongho, line 20: use `totalCount` returned from `useQueueEntries()` instead. `queueEntries.length` is technically incorrect because the REST endpoint limits the max returned result count to 50 by default.

This also means the per-status metrics are not calculated correctly as is, but I don't really know what's the best solution to that. We have a [related ticket](https://openmrs.atlassian.net/browse/O3-3015) regarding this.

cioan: Thanks @chibongho ! I have updated the PR to use totalCount. Thanks!

mseaton: If the same `totalCount` can't be applied to the per-status counts @cioan , then one way to solve this would be via multiple calls to the `queue-entry-metrics` endpoint that is here:  https://github.com/openmrs/openmrs-module-queue/blob/main/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java .  Not sure if this will be too much overhead with the multiple calls, but it would provide the most accurate data.

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface QueueTableMetricsProps {
+  selectedQueue: Queue;
+}
+
+function QueueTableMetrics({ selectedQueue }: QueueTableMetricsProps) {
+  const { t } = useTranslation();
+
+  const allowedStatuses = selectedQueue.allowedStatuses;
+  const { totalCount } = useQueueEntries({ queue: selectedQueue.uuid, isEnded: false });
```
mseaton, line 16: Should this also use the metric endpoint for consistency?

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div className={styles.metricsBorder}>
+      <QueueTableMetricsCard value={totalCount} headerLabel={t('totalPatients', 'Total Patients')} />
+      {allowedStatuses?.map((status) => {
+        return (
+          <QueueTableMetricsCard
+            queueUuid={selectedQueue.uuid}
+            serviceUuid={selectedQueue.service.uuid}
+            status={status.display}
```
mseaton, line 26: Why are you passing in the status display name here?  This should be status.uuid

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const { totalCount } = useQueueEntries({ queue: selectedQueue.uuid, isEnded: false });
+
+  return (
+    <div className={styles.metricsBorder}>
+      <QueueTableMetricsCard value={totalCount} headerLabel={t('totalPatients', 'Total Patients')} />
+      {allowedStatuses?.map((status) => {
+        return (
+          <QueueTableMetricsCard
+            queueUuid={selectedQueue.uuid}
+            serviceUuid={selectedQueue.service.uuid}
```
mseaton, line 25: serviceUuid is totally unnecessary - you are already passing in the queue, the service is redundant.

mseaton (changes requested):

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
mseaton: Looking pretty good @cioan .  A few additional comments to address

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
cioan: Sure, np. I have updated the PR.

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
cioan: Changed to status.uuid. Thanks!

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
cioan: OK, I removed it.

mseaton (approved):

PR #1119, opened 2024-04-30
(fix) O3-3136: Appointments: Don't Allow Checking In to Cancelled App…

mogoodrich:
…ointments

## Requirements

### DIFF
--- a/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
@@ -48,28 +48,27 @@ const AppointmentsActions: React.FC<AppointmentsActionsProps> = ({ appointment }
 
   const renderVisitStatus = () => {
     switch (true) {
-      case hasCheckedOutToday && isTodaysAppointment && !isCancelled:
-        return (
-          <Button kind="ghost" renderIcon={TaskComplete} iconDescription={t('checkedOut', 'Checked out')} size="sm">
-            {t('checkedOut', 'Checked out')}
-          </Button>
-        );
-      case checkOutButton.enabled && hasActiveVisitToday && isTodaysAppointment && !isCancelled:
-        return (
-          <Button onClick={handleCheckout} kind="danger--tertiary" size="sm">
-            {t('checkOut', 'Check out')}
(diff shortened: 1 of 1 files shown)


chibongho (approved):
Looks fine to me. I would have put the isCancelled case first to make the condition for other cases less complex.

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
```diff
@@ -47,21 +48,27 @@ const AppointmentsActions: React.FC<AppointmentsActionsProps> = ({ appointment }
...
-      case checkInButton.enabled && !hasActiveVisitToday && isTodaysAppointment: {
+      case checkInButton.enabled && !hasActiveVisitToday && isTodaysAppointment && !isCancelled: {
         return <CheckInButton patientUuid={patientUuid} appointment={appointment} />;
       }
+      case isCancelled:
+        return (
+          <Button kind="danger--ghost" iconDescription={t('cancelled', 'Cancelled')} size="sm">
+            {t('cancelled', 'Cancelled')}
+          </Button>
+        );
```
mseaton, line 71: It's getting a little hard to follow.  Wouldn't it be easier and clearer if this were just a series of if then return statements, and at the top putting the Cancelled case?

```if (isCancelled) return cancel button```

Then everything below this can proceed without having to worry about the Cancelled case...

mseaton (approved):
I'm sure it's fine, but see my comment, might be clearer...

packages/esm-appointments-app/src/appointments/common-components/appointments-actions.component.tsx
mogoodrich: Both you and @chibongho commented pretty much the same thing, so I'll switch I guess (we don't need to change to and if/then, fwiw, just put the Cancel as the first switch statement)

FWIW, I did have cancel first originally, but thought there was value in explicitly doing the logic for each case, even if it was repetitive, in case someone unknowing moved things around... (but I think you both have convinced me)

mogoodrich:
Thanks @mseaton  and @chibongho , I made the change you are suggesting, merging in now...

PR #1122, opened 2024-05-02
(chore)O3-2505: Multiple Requests to identifiersource

jwnasambu:
## Requirements


packages/esm-patient-registration-app/src/patient-registration/patient-registration.resource.ts
```diff
@@ -47,14 +47,33 @@ export function saveEncounter(encounter: Encounter) {
...
   const abortController = new AbortController();
 
-  return openmrsFetch(`${restBaseUrl}/idgen/identifiersource/${source}/identifier`, {
+  return openmrsFetch(`${restBaseUrl}/idgen/identifiersource`, {
     headers: {
       'Content-Type': 'application/json',
     },
-    method: 'POST',
-    body: {},
+    method: 'GET',
```
ibacher, line 54: I think this is patching the wrong method altogether. This one is for generating identifiers and isn't the requests causing the issue.

jwnasambu: @ibacher thanks for the correction.  Its supposed to be  `generateIdentifier` method.

ibacher:
Ultimately, the goal of the ticket is to replace [this function](https://github.com/openmrs/openmrs-esm-patient-management/blob/1461160f4bb1f1b24b9270868077f8e47b2d4ff0/packages/esm-patient-registration-app/src/offline.resources.ts#L128) and the places that it's used so that it only makes a single request instead of a request per identifier type.

jwnasambu:
@ibacher kindly do you mind having a look at the changes I have pushed?

ibacher (approved):
Weirdly, this looks like the idea works. Thanks @jwnasambu!

ibacher:
@jwnasambu Can you take a look at the failing test? It appears the user is no longer redirected correctly? You can find the playwright output here: https://github.com/openmrs/openmrs-esm-patient-management/actions/runs/8977383799?pr=1122

ibacher (changes requested):
Please fix the failing test

jwnasambu:
@ibacher 
basing on the error message am encountering I think its related to timeouts in the Playwright tests :
```

Error 1: Timeout waiting for a specific URL pattern after attempting to register a new patient.
Error 2: Timeout waiting for a success toast notification after attempting to register an unknown patient.

``` indicating that the test is taking longer than expected to complete or the expected condition is not being met within the specified time frame. Kindly on my side everything is running okay and I don't have permission to re-run the build on github.

File: packages/esm-patient-registration-app/src/offline.resources.ts [hunk 2]
```diff
@@ -75,17 +75,21 @@ export async function fetchPatientIdentifierTypesWithSources(): Promise<Array<Pa
...
 
+  const allIdentifierSources = identifierSourcesResponse.data.results;
+
   for (let i = 0; i < identifierTypes?.length; i++) {
-    identifierTypes[i].identifierSources = allIdentifierSources[i].data.results.map((source) => {
-      const option = find(autoGenOptions.data.results, { source: { uuid: source.uuid } });
-      source.autoGenerationOption = option;
-      return source;
-    });
+    identifierTypes[i].identifierSources = allIdentifierSources
```
ibacher, line 86:
```suggestion
    identifierTypes[i].identifierSources = allIdentifierSources.data.results
```

packages/esm-patient-registration-app/src/offline.resources.ts
```diff
@@ -75,17 +75,21 @@ export async function fetchPatientIdentifierTypesWithSources(): Promise<Array<Pa
...
+      .filter((source) => source.identifierTypeUuid === identifierTypes[i].uuid)
```
ibacher, line 87: AFAICT, the source doesn't have an identifierTypeUuid. I think that's ok, we just need the map, not the filter.

ibacher (changes requested):

packages/esm-patient-registration-app/src/offline.resources.ts
jwnasambu: @ibacher Thanks so much! let me do a fix.

jwnasambu:
@ibacher Kindly its fixed
Screenshot 2024-05-16 at 12 58 02 AM

kdaud (approved):

denniskigen:
I've tested this locally and it doesn't seem to work as intended. This is the network trace from dev3 where multiple calls are made to the endpoint:

CleanShot 2024-05-17 at 3  40 32@2x

Compared to this PR, which makes this request:

CleanShot 2024-05-17 at 3  40 53@2x

Whose response only returns the first `identifierType` passed as a query param in the request:

CleanShot 2024-05-17 at 3  43 15@2x

@ibacher I don't imagine that we should be concatenating the identifier types as is the case above, right?

ibacher:
Yeah, I think the ticket description is correct. I wrote it while tracing the code involved, so no filtering is probably the right move.

denniskigen:
So for this one call to `idgen/identifiersource/${source}/identifier` to get all the identifiers, what should the `source` param be?

denniskigen:
Right, any ideas for getting past that limitation?

ibacher:
@denniskigen I just created another. You're right that multiple `&identifierType=` parameters are ignored and it just gets the first.

dkayiwa:
@jwnasambu is there a reason why you do not just make a request to `/ws/rest/v1/idgen/identifiersource` in order to fetch all identifier sources without specifying the `identifierType` parameter?

jwnasambu:
@dkayiwa thanks for the review am on it right now.

ibacher (approved):
LGTM. I imagine the test failures are due to branch age.

PR #1124, opened 2024-05-07
(fix) O3-3015: Fetch all the queue entries to support filters on the frontend

vasharma05:
## Requirements

## Summary
Currently, the queue entries only fetch the top 50 results and the filters only work on those top 50 results. Now, we will be fetching all the queue entries using `useSWRInfinite` to fetch all the results and display them accordingly.

This PR also includes some UI improvements on the Table Toolbar.

### Before
image

image

### After
image

image

## Related Issue
https://issues.openmrs.org/browse/O3-3015


packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -1,30 +1,61 @@
...
+
+  const { data, ...rest } = useSWRInfinite<QueueEntryResponse, Error>(getUrl, openmrsFetch, {
+    initialSize: totalCount ? Math.ceil(totalCount / chunkSize) : 1,
+  });
+
+  useEffect(() => {
+    if (data?.[0]?.data?.totalCount && data[0].data.totalCount !== totalCount) {
+      setTotalCount(data[0].data.totalCount);
+    }
+  }, [data?.[0], totalCount, setTotalCount]);
```
chibongho, line 52: Is there a reason to put the setter function in the dependency array? I've seen this pattern elsewhere in this code base, but I don't understand why.

packages/esm-service-queues-app/src/clear-queue-entries-dialog/clear-queue-entries.component.tsx
```diff
@@ -29,9 +30,8 @@ const ClearQueueEntries: React.FC<ClearQueueEntriesProps> = ({ queueEntries }) =
 
   return (
     <Button
-      size="sm"
+      size={isDesktop(layout) ? 'sm' : 'lg'}
```
chibongho, line 33: thanks for this!

chibongho (changes requested):
I tested this locally and found a couple issues.
- The assumption that the chunkSize == 50 is incorrect for PIH distros, as it causes the number of fetched entries to be incorrect. (Some of them get double-fetched)
- The subsequent fetches to get the next chunk does not seem to get triggered immediately after the first fetch. I was able to trigger it by switching focus out of the page and back in. See:

https://github.com/openmrs/openmrs-esm-patient-management/assets/509602/7a29a740-85e2-448b-a949-661794eaca86

chibongho:
In this PR, the `useQueueEntries()` hook is modified to make multiple fetches as needed to get all queue entries. I think this is fine for this use case as, typically, the number of active queue entries is in the order of ~10s or ~100s.

It would be nice for use cases like this to have a usePaginate hook where the totalCount and the first batch of results are loaded on first fetch, and it delays fetching subsequent entries until the user paginates to the corresponding page in the DataTable. Thoughts? @ibacher @denniskigen @mseaton

ibacher:
I agree. That sounds like a useful utility to have.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -1,30 +1,61 @@
...
 const repString =
   'custom:(uuid,display,queue,status,patient,visit:(uuid,display,startDatetime,encounters:(uuid,display,diagnoses,encounterDatetime,encounterType,obs,encounterProviders,voided)),priority,priorityComment,sortWeight,startedAt,endedAt,locationWaitingFor,queueComingFrom,providerWaitingFor,previousQueueEntry)';
 
 export function useQueueEntries(searchCriteria?: QueueEntrySearchCriteria, rep: string = repString) {
-  const searchParam = new URLSearchParams();
-  for (let [key, value] of Object.entries(searchCriteria)) {
-    if (value != null) {
-      searchParam.append(key, value?.toString());
+  const [totalCount, setTotalCount] = useState<number>();
+  const chunkSize = 50;
```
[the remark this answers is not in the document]
vasharma05: I wasn't aware of this scenario, thank you @chibongho for pointing this out!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
vasharma05: Both the value and the setter are put in the dependency array, AFAIK

vasharma05:
> It would be nice for use cases like this to have a usePaginate hook where the totalCount and the first batch of results are loaded on first fetch, and it delays fetching subsequent entries until the user paginates to the corresponding page in the DataTable. Thoughts? @ibacher @denniskigen @mseaton

Yes, this is exactly what we want, but if you see, the main action of the table is to
1. Sort
2. Search the rows

Since this is not yet supported by the REST endpoint, we have to fetch all the entries and then handle the sorting and searching the whole results.
If we just fetch the first result, the search will only happen on the first batch, not the whole data.

vasharma05:
Hi @chibongho !
Can you please provide me the PIH server where I can test my implementation better, since there is not much data in the dev3 right now?
Thanks!

ibacher:
I will say that making the queue table sortable, in my view, is a mistake. Queues are supposed to be implicitly ordered. I'd rather we applied filtering to allow the users to zoom in on particular statuses, etc.

chibongho:
> Hi @chibongho ! Can you please provide me the PIH server where I can test my implementation better, since there is not much data in the dev3 right now? Thanks!

@vasharma05 just DM'd you the server + credentials

> I will say that making the queue table sortable, in my view, is a mistake. Queues are supposed to be implicitly ordered. I'd rather we applied filtering to allow the users to zoom in on particular statuses, etc.

We have no sorting in the queue table (and plan to keep it that way). Vineet's point stands regarding the search filtering though.

vasharma05:
Hi @chibongho!
I've made the changes, please re-review the changes.
Thanks!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
ibacher: @chibongho It's mostly about consistency. We've had issues with deps that should've be included being omitted, leading to inconsistent behavior, so its easiest across a code-base developed by a large number of contributors with varying degrees of familiarity with React, etc., just to include everything in the deps array.

Technically, setters from `useState()` don't need to be in the dependency array because part of the contract of `useState()` is that the setter remain consistent across renders. However, that's not necessarily the case for functions returned from other hooks

chibongho (approved):
Works locally. Thanks!

PR #1072, opened 2024-03-31
O3- 2998 monthly calender in appointment form fixed

Madhu-mac:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary
 This PR Fixes the issue with new enhanced monthly calender view 

- In this new enhanced monthly calendar view the user can able to go to the future months with the arrow buttons.
- And it only lets the user to select current and future dates which is the actual requirement needed here. The user should not 
select past dates. 
- The selected date gets highlighted in the calendar for better experience

## Screenshots
 

https://github.com/openmrs/openmrs-esm-patient-management/assets/99901964/454006d5-88a3-43c9-ae28-2afcad7c382d

## Related Issue
https://github.com/openmrs/openmrs-esm-patient-management/pull/1049

PR #1132, opened 2024-05-09
(fix) O3-3167: The recurring-appointment end date field does not get focused when it is required and the form does not submit

mccarthyaaron:
## Requirements

## Summary
This PR aims at fixing the issue where if the appointment form is submitted with a missing end date the form does not submit(which is okay) but the field causing the error is not put into focus.
We usually simply pass the ref callback from react-hook-form to the ref prop of a component and that does the job. With the date picker in the appointments form two issues arise:

The first issue is with the DatePickerInput. When it is passed as a child(children) to the DatePicker as we do, what ends up being rendered by DatePicker is not the DatePickerInput in its pure form. Rather, Carbon clones the DatePickerInput and overrides the ref prop with another value. This means that whatever ref value we pass to DatePickerInput(as long as it's being passed as child to DatePicker) never gets passed down to the element that is finally rendered by DatePicker.
This PR proposes a potential solution of manually calling the ref callback from react-hook-form with the date input element.

The second issue is with the DatePicker. Although the ref callback gets passed down to the rendered element, for some reason the value that the ref callback is called with(tested using ref={(node) => console.log(node)} ends up being an object rather than DOM node. The object looks something like {calendar: {...bunch-of-properties}. This was previously causing the error “elm.focus is not a function” because rather than elm being a DOM node it is the object mentioned above. Therefore elm.focus returned undefined which when called throws. This PR therefore also eliminates passing of a ref to the DatePicker component.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/121826239/df92e3bc-feba-4647-8109-a4c26c94383a

## Related Issue
https://openmrs.atlassian.net/browse/O3-3167


packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -195,6 +195,20 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
 
   useEffect(() => setValue('formIsRecurringAppointment', isRecurringAppointment), [isRecurringAppointment]);
 
+  const {
```
mogoodrich, line 198: would be good to add a brielf comment here as to why we are doing all this

mogoodrich (approved):
LGTM, though I'm flagging @denniskigen  and @ibacher  as they likely have more familarity with react-hook-form and the Carbon datepicker than me.

I made one comment about adding a comment about why we are doing this.

packages/esm-appointments-app/src/form/appointments-form.component.tsx
mccarthyaaron: @mogoodrich if you mean the useEffect that calls setValue() then that is from a previous PR not this one.

ibacher:
Couple of thoughts: 

1. You can pass the `ref` property to the `DatePickerInput` instead of the `DatePicker`
2. There should be a separate controller for the start-date and end-date fields

Basically, that would avoid three hooks, since we could just declaratively set the refs.

packages/esm-appointments-app/src/form/appointments-form.component.tsx
mogoodrich: I just meant the stuff you added as mentioned in the PR.  Everything looks good to me, I would just include a brief summary here so that someone in the future doesn't refactor this without understanding the reasoning behind it.

mccarthyaaron: @mogoodrich We are basically trying to manually call the ref callback from react-hook-form with the DOM node that we want to be focused. Rather than passing it as a ref prop which seems not be working

mogoodrich: Oh, sorry for not being clear @mccarthyaaron .  When I asked to include a brief summary "here", I meant to add a brief comment to the code itself, so a year or two down the line someone doesn't try to "simplify" the code and remove your changes without understanding why you added them.  Otherwise, this looks good to go!

mccarthyaaron: @mogoodrich let me work on it

mccarthyaaron: @mogoodrich I was also hoping to get some feedback from @ibacher and @denniskigen regarding this issue

ibacher: This is, unfortunately, basically the right approach. I think it's slightly better to use `getElementById('startDatePickerInput')` vs `querySelector()`, but they're essentially equivalent. Unfortunately, the `ref`s passed directly into the Carbon components point to the div rather than the `input`s (using `ref` pointing to the `DatePicker` div and then calling `ref.querySelector('input:nth-of-type(1)')` and `ref.querySelector('input:nth-of-type(2)`' might be marginally better since it doesn't restrict us from supporting two date editing components on the same page, but I'm not sure we need that flexibility.

mogoodrich:
Thanks @ibacher !  @mccarthyaaron  I think we are good to merge, let me know if you need me to merge.

mogoodrich:
Oh, and unfortunately, it looks like there are some conflicts to resolve.

ibacher:
@mogoodrich It would actually be a good idea to hold off on merging anything new here until we get the new release cut (should be later today)

mogoodrich:
Okay... but, oops, I think I did merge something else earlier today.

ibacher:
🤷 We'll deal with it. I wasn't intending to completely freeze things.

mccarthyaaron:
@mogoodrich @ibacher   This PR has some major conflicts with this PR https://github.com/openmrs/openmrs-esm-patient-management/pull/1072. The mentioned PR though already merged seems to have some issues considering the past work we have done. For example it eliminates an end date for recurring appointments and re-introduces an we already solved. So now a user can't submit the form after turning on isRecurring appointment because they don't have where to input the recurring appointment's end date. Also for this current PR, we are trying to focus the recurring pattern end date field if its not provided but the mentioned PR deletes the field all together. I am kindly suggesting that we revisit the mentioned PR and see a way forward that takes all the issues into account.

mogoodrich:
Thanks for the heads up, I will take a look @mccarthyaaron

mogoodrich:
I see what you mean, @mccarthyaaron ... I will look into this tomorrow but will likely revert #1072 at that time, will keep you posted.

mogoodrich:
@mccarthyaaron  I reverted the other commit and updated this ticket to latest main line of the code, checks should be running now.  Can you confirm everything now works correctly and this is ready for merge?

mccarthyaaron:
@mogoodrich its now okay. I think it's ready to merge

mogoodrich:
Thanks @mccarthyaaron , I am merging this in.

PR #1133, opened 2024-05-09
( fix ) - O3-3171 Replace overlays in the Appointments app

kb019:
## Requirements

## Summary
To replace the overlay in the esm-appointments-app with the workspaces. Also i think the size of the workspaceOverlay seems quite big which comes from the min-width property assigned in the esm-framework.

https://openmrs.atlassian.net/browse/O3-3171

**Video**

[OpenMRS - Personal - Microsoft_ Edge 2024-05-09 12-45-54.webm](https://github.com/openmrs/openmrs-esm-patient-management/assets/115476530/1fef98a0-570b-4d05-b7ec-8bc7242a5e2b)


packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx
```diff
@@ -217,14 +217,11 @@ const AppointmentsTable: React.FC<AppointmentsTableProps> = ({ appointments, isL
...
-                                      <AppointmentsForm
-                                        appointment={matchingAppointment}
-                                        context="editing"
-                                        closeWorkspace={closeOverlay}
-                                      />,
-                                    )
+                                    launchWorkspace('edit-appointments-form', {
+                                      appointment: matchingAppointment,
+                                      context: 'editing',
+                                      closeWorkspace: closeWorkspace.bind(null,"edit-appointments-form"),
```
brandones, line 223:
```suggestion
```

`closeWorkspace` is set automatically

packages/esm-appointments-app/src/metrics/metrics-header.component.tsx
```diff
@@ -24,11 +22,11 @@ const MetricsHeader: React.FC = () => {
     const props = {
       patientUuid: patientUuid,
       context: 'creating',
-      closeWorkspace: closeOverlay,
+      closeWorkspace: closeWorkspace.bind(null, 'create-appointment'),
```
brandones, line 25:
```suggestion
```

brandones:
Please get rid of the explicit passing of `closeWorkspace`.

You can probably delete `'./overlay.component'` and its associated files at this point.

At some point we should probably refactor the appointments system. It is weird that it has three different form names for the same component and that there is a `context: 'editing'` that maybe makes that component behave differently. Don't worry about that for this PR though.

brandones (changes requested):
As above

kb019:
Hey @brandones 😃. Thank you so much for reviewing. I have made the changes suggested by you. Also a overlay is used by a widget in appointments app which is rendered in extension slot of home app . So should i replace that by workspaceOverlay or keep as it is .

packages/esm-appointments-app/src/index.ts
```diff
@@ -78,6 +79,14 @@ export const appointmentsList = getSyncLifecycle(appointmentsListComponent, opti
 
 export const earlyAppointments = getSyncLifecycle(earlyAppointmentsComponent, options);
 
+export const editAppointmentsForm = getSyncLifecycle(appointementsForm, options);
+
+export const searchPatient = getSyncLifecycle(patientSearch, options);
+
+export const createAppointment = getSyncLifecycle(appointementsForm, options);
+
+export const addAppointment = getSyncLifecycle(appointementsForm, options);
```
mogoodrich, line 88: Defer to @brandones  but wouldn't it make sense to the just create a single "appointmentForm" here instead of "editAppointmentForm", "createAppointment" and "addAppointment" and then just that same component when registering the three different workspaces below?

brandones: Yeah that would be great, thanks @mogoodrich , thanks @kb019

brandones:
> Hey @brandones 😃. Thank you so much for reviewing. I have made the changes suggested by you. Also a overlay is used by a widget in appointments app which is rendered in extension slot of home app . The Component is [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/e073617a6a58a9eb37878fa2e9e1454562367860/packages/esm-appointments-app/src/home/home-appointments.component.tsx) . So should i replace that by workspaceOverlay or keep as it is .

We want to migrate everything to the new workspace system. There should only be one workspace renderer per page, and the home page already has one, so we probably don't need that overlay component there at all. Please make sure to find where it is in the UI and verify that everything still works after you make the change.

kb019:
Hi @brandones and @mogoodrich , I have made the necessary changes . Could you please review it. I have attached the updated video.Thanks @mogoodrich .Thanks @brandones .
[OpenMRS - Personal - Microsoft_ Edge 2024-05-19 20-39-42.webm](https://github.com/openmrs/openmrs-esm-patient-management/assets/115476530/41d7785f-3fc3-4ef2-8fd1-c67fcde79ead)

mogoodrich (approved):
LGTM,  assuming all the overlays have been tested and still open, thanks @kb019 !

I also just triggered a run of the tests to confirm they pass.

kb019:
> LGTM, assuming all the overlays have been tested and still open, thanks @kb019 !
> 
> I also just triggered a run of the tests to confirm they pass.

Thanks @mogoodrich.

brandones (approved):

brandones:
> @mogoodrich : LGTM, assuming all the overlays have been tested and still open,

No need to assume, that's what the video is for :laughing: 

Thanks @kb019 , and thanks for the review @mogoodrich !

PR #1134, opened 2024-05-10
(feat) O3-3159: Add patient header on appointment form

usamaidrsk:
## Requirements

## Summary

This PR adds Patient Banner at the top, below the “Appointment Form” header but above the “Location” so that the user is reminded of which patient they are creating an appointment for.

## Screenshots

image
image

## Related Issue

[O3-3159](https://openmrs.atlassian.net/browse/O3-3159)


packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -369,6 +368,7 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
   return (
     <Form onSubmit={handleSubmit(handleSaveAppointment, onError)}>
       <Stack gap={4}>
+        <AppointmentsFormPatientBanner patientUuid={patientUuid} />
```
mogoodrich, line 371: Oh, sorry @usamaidrsk ... I think the appropriate thing here would be just to include the extension slot for the patient banner, not recreate it...  ie just use something like:

<ExtensionSlot
            name="patient-header-slot"
            state={{
              patient,
              patientUuid: patientUuid,
              hideActionsOverflow: true,
            }}
          />

@denniskigen  @ibacher  that is the current way we are incorporating the banner, correct?

mogoodrich (commented):
Thanks @usamaidrsk , see my comment, i will also flag Dennis and Ian to confirm my thoughts...

packages/esm-appointments-app/src/form/appointments-form.component.tsx
usamaidrsk: Thanks @mogoodrich 

That slot is not really existing, AFAIK.
I actually first tried with this and setting it up in the `esm-patient-search-app` as `patient-banner-slot`, though had some challengers with the the `PatientSearchContext` in the banner component.  To get it right, looked like I had to do changes with in the component 

I would love to here suggestions too from @denniskigen @ibacher

ibacher: So, simply put, we do not want a large number of components that render a patient banner. The idea with the patient banner is to have a single component that can be used in multiple places and provide consistent rendering of important patient details. Basically, if we need to make changes to the header, it's strongly preferable to only have to do this in a single place. This pattern would require us to update it in multiple places, which greatly increases the maintenance and the chance of getting things wrong.

@usamaidrsk I don't really understand what you mean with "hallengers with the the PatientSearchContext in the banner component". Maybe you could elaborate? The banner slot does need both a patient and patient uuid object

mogoodrich: Thanks @ibacher !  And @usamaidrsk ... I'm not sure what you mean about the slot not existing, I believe my above code will work, though I haven't tested.  Check out the visit-form in the patient chart.

ibacher (approved):
LGTM

mogoodrich (commented):
Thanks @usamaidrsk , merging this in!

mogoodrich:
Or, oops, @ibacher  already merged, thanks @ibacher  :)

PR #1136, opened 2024-05-14
(feat) - O3-3200 service queues - allow manually enter and edit the d…

chibongho:
…ate/time of queue transitions

## Requirements

## Summary

This PR adds an additional field in the queue entry transition / edit form to change the startedAt date. 
## Screenshots

Transition:
image
Edit:
image

## Other

- We currently rely on a queue entry's startedAt time to determine its previous queue entry (which should have matching endedAt time along with same patient and visit). Allowing edits to the startedAt field can cause this to break.
- There is browser-side logic, but no server-side logic, to prevent setting the transition date to the future. This might actually be desirable in the rare case where there is clock drift between client and server.


mseaton (commented):
Code itself looks generally good to me, with a few comments:

* Editing date/time is likely going to be relatively rare, and also something we don't want to encourage for real-time use.  I wonder if we want to default to having this behind some sort of hidden toggle, which one can expand to change the date/time, and which normally submits these as null to allow them to be set server-side in the typical case

* It seems odd to place the date/time widgets between the "priority" and "priorityComment" fields, which would seem to go together

* This is a lot of code to have to write to support a date/time picker for a field, and especially all the stuff around converting 12 to 24, dealing with AM/PM, validating, etc.  We should consider improvements to what is available in our styleguide and other framework components to support this without needing all this here.  Just something to consider but not a blocker for this PR.

chibongho:
> Code itself looks generally good to me, with a few comments:
> 
> * Editing date/time is likely going to be relatively rare, and also something we don't want to encourage for real-time use.  I wonder if we want to default to having this behind some sort of hidden toggle, which one can expand to change the date/time, and which normally submits these as null to allow them to be set server-side in the typical case
> * It seems odd to place the date/time widgets between the "priority" and "priorityComment" fields, which would seem to go together
> * This is a lot of code to have to write to support a date/time picker for a field, and especially all the stuff around converting 12 to 24, dealing with AM/PM, validating, etc.  We should consider improvements to what is available in our styleguide and other framework components to support this without needing all this here.  Just something to consider but not a blocker for this PR.

Good points. I added a toggle to enable editing of the date time fields, and rearranged the fields so that priority and priorityComment are together.

I second the point on working with date and time pickers. Not pleasant to use atm.

image

mseaton (approved):
Generally looks ok to me, though would be good to get another set of eyes on it.  Will also likely need some definite testing to ensure timezone handling is as expected.

PR #1149, opened 2024-05-21
(fix) O3-3260: When editing an appointment, changing the service prevents form submission due to duration field error

mccarthyaaron:
## Requirements

## Summary
When editing an appointment, selecting a different service refuses the form from submitting due to an error in the duration field. This is caused by the onChange handler of the “selectedService” field setting the value of the “duration” field to the durationMins property of the service. This has no issues when creating an appointment from scratch. But when editing an appointment which already has a duration it sets the value to the durationMins property of the service which is null at times. So even though the form field displays a duration value like 30, react-hook-form is passed null so it refuses submission.

The solution presented by this PR is that setting the appointment duration based on the service selected, should only be limited to when creating a new appointment. Such that appointments being edited should maintain their initial duration unless changed manually by the user.

## Related Issue
https://openmrs.atlassian.net/browse/O3-3260


brandones:
Hi @mccarthyaaron , thank you very much for the well-documented bug. You made a ticket with an excellent title, clear issue description and video. Well done and thank you.

On this PR, I'm not sure this is the right solution. I'm going to loop in some doctor & design folks on the ticket for advice.

mccarthyaaron:
thanks @brandones... will be waiting for the feedback

brandones:
@mccarthyaaron Sounds like Ciaran favors this:

The duration should be updated to the default duration of the new service type only if there is one and if the existing duration is the one automatically set by the previous service type.

So that is to say, if the duration is the previous default, update it automatically (if the new service type has a non-null default duration), otherwise leave it.

All clear, you can move forward with that? Any questions?

mccarthyaaron:
@brandones let me phrase it in my own understanding to see if i get you. When the appointment is being created, the user selects the service type  and as a result the duration is set based off that service type. when the appointment is being edited, the fact that the duration was set by a service type means that changing the service type should then change the duration to that of the newly selected service type but only if the newly selected service type actually has one. Which means that for the other scenario where the duration was set manually by the user in the duration field, when editing that appointment, changing the service type should not change the duration. Is that it?

brandones:
Yep, that sounds right @mccarthyaaron

brandones:
@mccarthyaaron Any progress on this?

mccarthyaaron:
@brandones pushed some changes. Could you take a look?

brandones (approved):
LGTM, thank you!

PR #584, opened 2023-03-06
feat: Initial work to migrate both appointments and service queues to  home page

donaldkibet:
## Requirements

- [ ] This PR has a title that briefly describes the work done, including the ticket number if there is a ticket.
- [ ] My work conforms to the [**OpenMRS 3.0 Styleguide**](https://om.rs/styleguide).
- [ ] I checked for feature overlap with [**existing widgets**](https://om.rs/directory).

## Summary

Kapture 2023-03-06 at 16 44 12

## Screenshots

*None.*

## Related Issue

*None.*

## Other

*None.*

PR #1150, opened 2024-05-23
(fix) O3-3273: Scheduled appointments card always shows "Not arrived' even for past dates

mccarthyaaron:
## Requirements

## Summary
This PR addresses the issue where the scheduled appointments card on the appointments page displays "Not arrived" even for past dates. The suggested solution is to use the selected date from the SelectedDateContext and check if its a date before today's date.

## Screenshots
https://github.com/openmrs/openmrs-esm-patient-management/assets/121826239/d3e3a93e-a410-4360-a138-3f92f37b1a7e

## Related Issue
https://openmrs.atlassian.net/browse/O3-3273


packages/esm-appointments-app/src/metrics/metrics-card.component.tsx
```diff
@@ -29,7 +30,8 @@ const MetricsCard: React.FC<MetricsCardProps> = ({
   appointmentDate,
```
brandones, line 30: Can you clean up the MetricsCard props? The only usage I see is [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/20499f906983a27b1dac0f80a07382280fdbaaa4/packages/esm-appointments-app/src/metrics/appointments-metrics.component.tsx#L47-L66). Which means several of these props are not being used anywhere. You have deleted the only line that used `appointmentDate`, which means now you should remove `appointmentDate` from the interface. Looks like `view` isn't used by anyone anywhere either. Nor is `children`.

brandones (changes requested):

brandones:
Please don't check the box that says "My work includes tests or is validated by existing tests" unless your work includes tests or is validated by existing tests.

packages/esm-appointments-app/src/metrics/metrics-card.component.tsx
brandones: For record-keeping: looks like `view` was added by @CynthiaKamau [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/338) and partially removed by @donaldkibet [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/652) and [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/584). Dunno about the other props.

brandones (approved):

brandones:
Thanks @mccarthyaaron !

PR #1152, opened 2024-05-26
(fix) O3-3290: Resolve flickering scroll bar issue in service queues table

Twiineenock:
## Requirements

## Summary

This PR is aimed at solving this issue.
Service queues table scroll bar appears and disappears on hovering the table
This is demonstrated in the files attached to the issue [here](https://openmrs.atlassian.net/browse/O3-3290)

Implemented:
 - vertical and horizontal scrollbars to only appear when content size exceeds table height/width
 - Scrollbars must persist even at hoover states.

## Screenshots

after fix:
service-queues#

## Related Issue

[https://openmrs.atlassian.net/browse/O3-3290](https://openmrs.atlassian.net/browse/O3-3290)

### Screancasts
[service-queues.webm](https://github.com/openmrs/openmrs-esm-patient-management/assets/122885804/9d4ebd13-367d-44c0-a564-6592f3ea470d)


chibongho (changes requested):
Sorry for the late review. I think it would be better if we actually just forgoing the table scrollbar in favor of making the table height longer (and having just one vertical scrollbar for the entire page). Can you change it to that instead?

Twiineenock:
> Sorry for the late review. I think it would be better if we actually just forgoing the table scrollbar in favor of making the table height longer (and having just one vertical scrollbar for the entire page). Can you change it to that instead?

@chibongho, I updated this PR to meet this. Thanks!

denniskigen (approved):
Nice! Thanks, @Twiineenock!

PR #1162, opened 2024-05-29
(fix) O3-2955 Add patient to queue defaults to the first queue at the currently selected location

brandones:
## Requirements

## Summary

This makes it so that if the user is on the page for a particular service queue, that queue is selected by default when adding a patient to a queue.

This also includes some minor refactoring and renaming.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/1031876/0693271d-b775-44e3-9fdc-2ff2ff3121a9

## Related Issue

https://openmrs.atlassian.net/issues/O3-2955


brandones:
Whoops, gotta fix the build

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
```diff
@@ -1,47 +1,54 @@
...
   useEffect(() => {
-    if (priority === emergencyPriorityConceptUuid) {
+    if (priority === config.concepts.emergencyPriorityConceptUuid) {
       setSortWeight(1);
     }
   }, [priority]);
 
   useEffect(() => {
     if (queues?.length > 0) {
-      setSelectedService(queues[0].uuid);
```
mseaton, line 38: I'm not entirely sure this solves the issue presented in the ticket @brandones .  The issue is that the active visits table is not limited to a specific queue, but allows the user to add a new queue entry without specifying which queue they want to add the queue entry into.  As far as I can tell, that behavior remains after these changes, as below the `selectedService` just defaults to the first queue available in the given location, and there is nothing in place that prompts the user to select the queue / service of interest in this situation.

mseaton (commented):
See comment @brandones .  I might be missing something though.

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
brandones: Oh, I think I understand what you meant in the ticket now. At present the behavior is that when a patient is added from a view that has multiple queues, there is a default value for the queue to add to, which happens to be the first queue in the list. The desired behavior is that there should be no default value for the queue when the user is in a context with multiple queues.

brandones:
@mseaton I think this now does what it is supposed to. Ready for re-review.

https://github.com/openmrs/openmrs-esm-patient-management/assets/1031876/b46e4496-b2d1-432e-91fe-2b93d8ba3e43

I have also deleted some dead code and some useless tests.

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
```diff
@@ -105,59 +102,40 @@ const StartVisitQueueFields: React.FC = () => {
         )}
       </section>
```
mseaton, line 104: Is this among the dead code you are referring to?  I would have thought that we'd want to allow the user to choose the status, while providing a default, but I don't see this happening on dev3

mseaton (approved):
Thanks @brandones .  To be honest, I'm not 100% sure any more of the  exact status of things, as a lot of different changes have already been implemented since I raised this ticket, but this seems to look good.  Personally I find the naming of things confusing, as often we dealing with `queues`, but we are calling them `services`, but services are a specific concept property on queue, so it can be really hard to follow what's what, but from what I can tell this looks good.

chibongho (approved):
Looks good. One minor comment.

packages/esm-service-queues-app/src/patient-search/visit-form/visit-form.scss
```diff
@@ -46,6 +46,7 @@
   display: flex;
   flex-direction: column;
   justify-content: space-between;
+  height: calc(var(--desktop-workspace-window-height) - 7rem - 2px); // 7rem patient banner, plus border
```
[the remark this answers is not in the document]
brandones: Yeah, if we can figure out how to do that that would be great. This is part of the motivation for https://openmrs.atlassian.net/browse/O3-3340

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
brandones: Yeah, that `styles.sectionHidden` makes it not show up. And it's not toggled anywhere. Not sure why someone would have decided to make it not appear on the screen, rather than deleting it. Worse than commented-out code.

denniskigen (approved):

packages/esm-service-queues-app/src/patient-search/visit-form-queue-fields/visit-form-queue-fields.component.tsx
brandones: Oops, the status part of the form was being used, passed through the form submit event, which made it not subject to type-checking. I have fixed this in a follow-up PR: https://github.com/openmrs/openmrs-esm-patient-management/pull/1171

PR #1166, opened 2024-05-31
(fix) O3-3336 Fix up service queue table schema

brandones:
## Requirements

## Summary

At present the queue table config is unstructured, un-typed, and un-validated. This improves the config schema so that it is structured, typed, and validated.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/1031876/a79c84de-1db2-4742-955a-20eb6efe65e0

## Related Issue

https://issues.openmrs.org/browse/O3-3336


ojwanganto:
Thanks @brandones. The changes look great

denniskigen (commented):
Nice work! I've left some minor suggestions.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -1,90 +1,117 @@
...
+  'blue',
+  'teal',
+  'cyan',
+  'gray',
+  'green',
+  'warm-gray',
+  'cool-gray',
+  'high-contrast',
+  'outline',
+] as const;
```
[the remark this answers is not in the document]
brandones: This list is in the order provided by the Carbon documentation, which I figure we will continue to copy-paste it from in the future.

chibongho (approved):
Generally approve. I have a couple minor comments, but looks good overall. Thanks for working on this.

@ojwanganto you're probably aware already, but this will affect the service queues configurations if you are adding custom config to the queue tables.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -196,21 +218,150 @@ export const configSchema = {
...
-    _description: `Configurations of columns to show for the queue table.
-      Multiple configurations can be provided, each can be applied generally, or to tables 
-      for a particular queue, particular status, or even particular queue+status combination.
-      If multiple configs are defined, the first config with matching appliedTo condition
-      is used.
-      See https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/config-schema.ts
-      for full schema definition and example.
-    `,
-    _default: defaultTablesConfig,
+  ...columnConfigSchemaFragment,
```
chibongho, line 221: This line is unintentional, right? `columnConfigSchemaFragment` also appears in the `queueTables` Object.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-extension-cell.component.tsx
```diff
@@ -1,23 +1,10 @@
...
+import { ExtensionSlot } from '@openmrs/esm-framework';
 import { type QueueTableColumnFunction, type QueueTableCellComponentProps } from '../../types';
-import { type ExtensionColumnConfig } from '../../config-schema';
-
-export const queueTableExtensionColumn: QueueTableColumnFunction = (key, header, config: ExtensionColumnConfig) => {
-  const state = config?.state ?? {};
 
+export const queueTableExtensionColumn: QueueTableColumnFunction = (key, header) => {
   const QueueTableExtensionCell = ({ queueEntry }: QueueTableCellComponentProps) => {
-    const extensions = useRenderableExtensions('queue-table-extension-column-slot');
```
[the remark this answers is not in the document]
brandones: Yeah `useRenderableExtensions` AFAICT has all the same overhead as an extension slot. Used the way as it is at the cell level there's no real difference: compare with the [ExtensionSlot](https://github.com/openmrs/openmrs-esm-core/blob/eeb48a4c573111d7323956164d75e7f64723543b/packages/framework/esm-react-utils/src/ExtensionSlot.tsx) implementation. I don't know how this hook could be used to make things lighter. Maybe @ibacher has an idea. I think it's unlikely that this kind of thing will work until we have [framework-native mounting](https://openmrs.atlassian.net/browse/O3-1418).

packages/esm-service-queues-app/src/config-schema.ts
brandones: This makes it so these properties can be passed at the "global" level or per-column. So if you set `priorityConfigs` at the top level, it will apply to all columns of type `priority` unless overridden. I also added a test for this. This kind of thing is already [in the MCOE config](https://github.com/PIH/openmrs-config-pihsl/blob/ac6f5914cbd0d5c41923903634d128c849bc84f5/configuration/frontend/config.json#L255-L275). However, it presumably wasn't doing anything, because there was no code to actually handle column config elements passed at the top level.

chibongho: Hm... not sure if I like this. If we create a `TableDefinition` with no `appliedTo` criteria, it will also apply globally. When there are multiple `TableDefinition` with matching `appliedTo` criteria, currently we just pick the first matching one. This `global` column config makes the order of precedence more complex. I can imagine there are cases where this simplifies things, but I think it makes the config (even) harder to understand.

brandones: Sure, fair enough. I'll remove that.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-extension-cell.component.tsx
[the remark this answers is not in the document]
ojwanganto: Thanks @chibongho for the heads up. CC @donaldkibet @makombe

packages/esm-service-queues-app/src/config-schema.ts
brandones: Removed, including the test which tested this behavior.

packages/esm-service-queues-app/src/queue-table/cells/queue-table-extension-cell.component.tsx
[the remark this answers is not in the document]
ibacher: Hmmm... the idea here was basically to end up with a parcel-per-column and then feed each parcel the row data based on props, e.g., so that we end up with 8 parcels rendered # rows times rather than 8 * # rows parcels. That said, it may not be a substantial reduction in rendering times. I think implementation-wise, that'd only work if `const extensions` is assigned outside of the `QueueTableExtensionCell` component though.

chibongho (approved):
LGTM!

denniskigen (approved):
Thanks, @brandones. LGTM too. It looks like there are some Typescript errors related to the `visitQueueNumberAttributeUuid`.

PR #1167, opened 2024-06-03
(fix) Use correct concept UUID for registration form image uploads

denniskigen:
## Requirements

## Summary

This PR updates the key used to fetch the concept UUID for [POSTing image uploads](https://rest.openmrs.org/#query-observations-by-uuid) from the [style guide](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-styleguide/src/config-schema.ts#L26). 

It also fixes a deprecation warning for the `atob` function in the `dataURItoFile` function, replacing it with `window.atob`.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
@@ -131,14 +131,15 @@ export class FormManager {
 
       await this.saveObservations(values.obs, savePatientResponse, currentLocation, currentUser, config);
 
-      const { patientPhotoUuid } = await getConfig('@openmrs/esm-styleguide');
-      if (patientPhotoUuid && capturePhotoProps?.imageData) {
+      const { patientPhotoConceptUuid } = await getConfig('@openmrs/esm-styleguide');
+
+      if (patientPhotoConceptUuid && capturePhotoProps?.imageData) {
         await savePatientPhoto(
           savePatientResponse.data.uuid,
           capturePhotoProps.imageData,
           `${restBaseUrl}/obs`,
           capturePhotoProps.dateTime || new Date().toISOString(),
-          patientPhotoUuid,
+          patientPhotoConceptUuid,
         );
       }
     }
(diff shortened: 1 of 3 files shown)


packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -131,14 +131,15 @@ export class FormManager {
 
       await this.saveObservations(values.obs, savePatientResponse, currentLocation, currentUser, config);
 
-      const { patientPhotoUuid } = await getConfig('@openmrs/esm-styleguide');
-      if (patientPhotoUuid && capturePhotoProps?.imageData) {
+      const { patientPhotoConceptUuid } = await getConfig('@openmrs/esm-styleguide');
```
brandones, line 134: This seems like a bug that typing should have prevented. Can we add the typing for the config object?

denniskigen: It does, can you suggest a change?

brandones: I would have, if I'd been sure that such a type already existed. Looked it up and it does.

brandones, line 134:
```suggestion
      const { patientPhotoConceptUuid } = await getConfig<StyleguideConfigObject>('@openmrs/esm-styleguide');
```

brandones: You'll probably need to add the import anyway

denniskigen: Doh 

CleanShot 2024-06-03 at 6  25 00@2x

brandones: Merged that

denniskigen:
This should be good to go, @brandones!

brandones (approved):
Thanks @denniskigen !

PR #1169, opened 2024-06-03
(feat) O3-2473: Service Queues - Visit Type selector fails and is not needed

brandones:
## Requirements

## Summary

I have marked this as `feat` rather than `fix` because it changes the UI. The search bar is no longer shown if there are 5 or fewer options.

Fixed the bug with the search bar, wrote some tests and improved existing tests (including ones that didn't do what they said they did), and then made it so the search bar won't appear if there are 5 or fewer options.

The UI is still technically buggy if there is a search bar, because it is supposed to paginate results, and there is no pagination implemented. `usePagination` had been used just to truncate the results, giving the false impression that there was some kind of pagination happening.

## Screenshots

What it looks like with 5 visit types

https://github.com/openmrs/openmrs-esm-patient-management/assets/1031876/669f3689-71fe-4f2f-bc61-ad9b405feaba

Here's a demo of the search bar, for which I've reduced the `MAX_RESULTS` value to 3, just for demo purposes.

https://github.com/openmrs/openmrs-esm-patient-management/assets/1031876/3363ea43-4d88-492f-b10d-8b1852f0a051

## Related Issue
https://openmrs.atlassian.net/issues/O3-2473

## Other

Heads up about the behavior change @ojwanganto


packages/esm-service-queues-app/src/patient-search/visit-form/visit-form.component.tsx
```diff
@@ -314,24 +309,22 @@ const VisitForm: React.FC<VisitFormProps> = ({ patientUuid, toggleSearchType, cl
               </ContentSwitcher>
             )}
             {config.showRecommendedVisitTypeTab && contentSwitcherIndex === 0 && !isLoading && (
```
mseaton, line 311: Presumably we can get rid of the isLoading check here, as this is handled by the skeleton within the selector component?

mseaton (approved):
LGTM @brandones thanks!  Made one small comment, but looks like a nice improvement.

PR #1184, opened 2024-06-10
O3-3210 ward app - configuration system for ward patient cards 

chibongho:
## Requirements

## Summary

This PR introduces a configuration system to ward app, as mocked up [here](https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda1d3119350891d20dcc7). 

The configuration models each Ward Patient Card as a list of rows. In the link above, a card has 3 rows: the header, middle section (baby info), and footer. On other pages in the mock up, there can be other types of rows as well. (For example, [here](https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda2368de2754cb052ef8c) some cards can have extra rows.) The patient's card (and sometimes only its header) can also appear in different contexts, either [fully](https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda236f0ad0df6e5d786ec) (in the workspace) or [patially with the header](https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda28704d2debed99316ab) (also in the workspace).

The most common type of row is the "Bento Box" row, which can contain 1 or more elements displaying patient attributes. In the first link, the header, baby info, and footer rows should theoretically all be modeled as a bento box row.

The configuration system defines a card as having the header row, and any other additional row can be slotted in as extension. The header row gets special treatment (instead of it being just another extension) as it can sometimes be rendered (with the other rows) in different contexts.

In this PR, 4 types of bento elements (bed number, patient name, patient age, patient address) are defined and can be configured to appear in any bento boxes. In future PRs, more bento elements will be defined to support other elements defined in the mockups.

## Screenshots

image


packages/esm-ward-app/src/beds/occupied-bed.component.tsx
```diff
@@ -0,0 +1,40 @@
...
+  patients: Patient[];
+  bed: Bed;
+}
+const OccupiedBed: React.FC<OccupiedBedProps> = ({ patients, bed }) => {
+  return (
+    <div className={styles.occupiedBed}>
+      {patients.map((patient, index: number) => {
+        const last = index === patients.length - 1;
+        return (
+          <div key={patient.uuid + ' ' + index}>
```
ibacher, line 19: Shouldn't `patient.uuid` be unique?

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
```diff
@@ -0,0 +1,40 @@
...
+    </div>
+  );
+};
+
+const BedShareDivider = () => {
+  const { t } = useTranslation();
+  return (
+    <div className={styles.bedDivider}>
+      <div className={styles.bedDividerLine}></div>
+      <Tag>{t('bedshare', 'Bed share')}</Tag>
```
ibacher, line 34: Not entirely sure I follow what this corresponds to in the designs?

packages/esm-ward-app/src/beds/occupied-bed.scss
```diff
@@ -0,0 +1,24 @@
+@use '@carbon/styles/scss/spacing';
+@use '@carbon/styles/scss/type';
+@import '~@openmrs/esm-styleguide/src/vars';
```
ibacher, line 3:
```suggestion
@use '@openmrs/esm-styleguide/src/vars';
```

Requires `vars.$ui-02`, etc., but slightly friendlier to the Sass compiler.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+  'patient-name',
+  'patient-age',
+  'patient-address',
+  'admission-time',
+];
+
+export const configSchema: ConfigSchema = {
+  wardPatientCards: {
+    _description: 'Configure the display of ward patient cards',
+    bentoElementDefinitions: {
```
ibacher, line 27: What is a `bentoElementDefinition`? I'm not entirely sure this nomenclature is clear.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+
+export interface WardPatientCardsConfig {
+  bentoElementDefinitions: Array<BentoElementDefinition>;
+  cardDefinitions: Array<WardPatientCardDefinition>;
+}
+
+export interface WardPatientCardDefinition {
+  card: {
+    id: string;
+    header: Array<string>; // an array of (either built-in or custom) bento element ids
```
ibacher, line 103: Convert this into a doc string (`/** */`)

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+  };
+  appliedTo?: Array<{
+    location: string; // locationUuid. If given, only applies to patients at the specified ward locations. (If not provided, applies to all locations)
```
ibacher, line 106: This would also be useful as a docstring.

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-age.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useTranslation } from 'react-i18next';
+import { type WardPatientCardBentoElement } from '../../types';
+
+const WardPatientAge: WardPatientCardBentoElement = ({ patient }) => {
+  const { t } = useTranslation();
+  return (
+    <div>
+      {t('yearsOld', '{{age}} yrs', {
+        age: patient?.person?.age,
+      })}
```
ibacher, line 11: I'd probably suggest using esm-framework's `age()` function here. That handles years, months, weeks, and days (may need some extension for hours).

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-name.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { type WardPatientCardBentoElement } from '../../types';
+import styles from '../ward-patient-card.scss';
+
+const WardPatientName: WardPatientCardBentoElement = ({ patient }) => {
+  // TODO: make server return patient.display and use that instead
+  const { givenName, familyName } = patient?.person?.preferredName;
```
ibacher, line 7: Better to use `patient?.person?.display` (`patient?.display` has the primary identifier as well).

ibacher (commented):
Mostly LGTM. A few q.s and suggestions.

packages/esm-ward-app/src/beds/empty-bed.component.tsx
```diff
@@ -10,7 +10,9 @@ interface EmptyBedProps {
 const EmptyBed: React.FC<EmptyBedProps> = ({ bed }) => {
   return (
     <div className={styles.container}>
-      <span className={admittedPatientHeaderStyles.bedNumber}>{bed.bedNumber}</span>
+      <span className={`${wardPatientCardStyles.wardPatientBedNumber} ${wardPatientCardStyles.empty}`}>
+        {bed.bedNumber}
+      </span>
       <p className={styles.emptyBed}>Empty Bed</p>
```
mseaton, line 16: This needs to be translated.

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { type Patient } from '@openmrs/esm-framework';
+import { type Bed } from '../types';
+import styles from './occupied-bed.scss';
+import { Tag } from '@carbon/react';
+import { useTranslation } from 'react-i18next';
+import WardPatientCard from '../ward-patient-card/ward-pateint-card.component';
```
mseaton, line 7: ward-pateint-card.component is misspelled

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
```diff
@@ -0,0 +1,40 @@
...
+  bed: Bed;
+}
+const OccupiedBed: React.FC<OccupiedBedProps> = ({ patients, bed }) => {
+  return (
+    <div className={styles.occupiedBed}>
+      {patients.map((patient, index: number) => {
+        const last = index === patients.length - 1;
+        return (
+          <div key={patient.uuid + ' ' + index}>
+            <WardPatientCard patient={patient} bed={bed} status={'admitted'} />
```
mseaton, line 20: An occupied bed != admitted, it just means the bed is occupied.  Admission is a separate, orthongonal concern.

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
mseaton: @ibacher see images in Zeppelin, also summarized here:  https://openmrs.atlassian.net/browse/O3-3198

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+const defaultWardPatientCard: WardPatientCardDefinition = {
+  card: {
+    id: 'default-card',
+    header: ['bed-number', 'patient-name', 'patient-age', 'patient-address'],
+  },
+  appliedTo: null,
+};
 
-export interface ConfigObject {}
+export const defaultBentoElementConfig: BentoElementConfig = {
```
mseaton, line 12: I don't understand why this is and why it has addressFields in it.

packages/esm-ward-app/src/config-schema.ts
mseaton: I think something like `patientCardElement` might be clearer.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+          _type: Type.String,
+          _description: 'The unique identifier for this custom bento element',
+        },
+        elementType: {
+          _type: Type.String,
+          _description: 'The bento element type',
+          _validators: [validators.oneOf(bentoElementTypes)],
+        },
+        config: {
+          addressFields: {
```
mseaton, line 41: I really don't like this here.  A specific type of `patientCardElement` should have a configuration that takes in an address field.  Just like other types of `patientCardElement` (like Gravity) might take in a `Concept`.  This should not be baked into a generic config schema.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+      _default: [defaultWardPatientCard],
+      _description: `An array of card configuration. A card configuration can be applied to different ward locations.
+         If multiple card configurations apply to a location, only the first one is chosen.`,
+      _elements: {
+        id: {
+          _type: Type.String,
+          _description: 'The unique identifier for this card definition. Currently unused, but that might change.',
+        },
+        card: {
+          header: {
```
mseaton, line 59: Per our discussion, I'm still of the belief that a card should be made up of 1-N sections, which contain 1-N elements.  And these sections could have a type associated with them, or otherwise be configured to indicate if they should appear up in particular views (eg. in a minimal view that is used in the Admission Request workspace, or in a maximal view used in the ward cards.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+  }>;
+}
+
+export type BentoElementDefinition = {
+  id: string;
+  elementType: BentoElementType;
+  config?: BentoElementConfig;
+};
+
+export interface PatientAddressBentoElementConfig {
```
mseaton, line 116: This is what I would expect to see, not to have this at the top level

packages/esm-ward-app/src/types/index.ts
```diff
@@ -1,4 +1,32 @@
...
+export interface WardPatientCardProps {
+  patient: Patient;
+  bed: Bed;
+  status?: WardPatientStatus;
+}
+
+export type WardPatientCardRow = React.FC<WardPatientCardProps>;
+export type WardPatientCardBentoElement = React.FC<WardPatientCardProps>;
+
+export type WardPatientStatus = 'admitted' | 'pending';
```
mseaton, line 20: Patients might be in a bed but not admitted.

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-age.tsx
mseaton: +1

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-name.tsx
mseaton: Also there is likely a utility method to get a patient name (or should be) for display.  There may be active work ongoing around this, but the way a patient name is formatted is configurable on the backend, and should likely be respected.

chibongho: There is a `displayName` function from esm-framework, but it takes in a FHIR patient instead of an OpenMRS Patient. I'll use `patient?.person?.display` for now.

chibongho: Actually... the REST endpoint we are using (`/adminLocation`) does not return `patient.person.display` even with `v=full`. I'll keep this (and the TODO) for now.

packages/esm-ward-app/src/config-schema.ts
chibongho: A similar change was made to the queues app. I believe it was to facilitate adding validators on the fields. I'm not sure if it's possible to have a validator work by defining specific configs per element type, but I can play around with it. 
@brandones what do you think?

packages/esm-ward-app/src/config-schema.ts
chibongho: Yeah, it's the individual elements that can show up within a patient card row (like bed number, patient name, patiant age, patient address). I'll change it to `PatientCardRowElement`.

packages/esm-ward-app/src/config-schema.ts
chibongho: I think your suggestion of having every Patient Card Rows be modeled as a "bento box", even with rows that don't fit neatly within that paradigm as a bento box with one element (possibly an extension), is worth exploring. I've tried modeling this by having different row types and wasn't that satisfied with how complex it was.

packages/esm-ward-app/src/config-schema.ts
chibongho: It's supposed to be a common config object for various types of bentoElements. (See comment below)

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-age.tsx
chibongho: The backend isn't returning the birthdate either. I'll add a TODO.

ojwanganto: There is [this PR](https://github.com/Bahmni/openmrs-module-appointments/pull/148). May be we push for its review

ibacher: Ah... I hadn't looked into the backend, but it looks like the fields returned from patient [are hard-coded](https://github.com/Bahmni/openmrs-module-bedmanagement/blob/2058e237664331f90a2172a8529920e9a53cee00/omod/src/main/java/org/openmrs/module/bedmanagement/rest/resource/AdmissionLocationResource.java#L190). Here's a ticket to adjust that: https://openmrs.atlassian.net/browse/BED-10.

packages/esm-ward-app/src/ward-patient-card/bento-elements/ward-patient-name.tsx
ibacher: `patient?.person?.display`, when it exists, is already formatted according to the backend name template. We sort of need a utility function for FHIR, because the display name is (in FHIRPath syntax) `Patient.name.where(use='usual').text.first()`, but we have a fallback if the `text` element is not defined which happens in cases where we've constructed a pseudo-FHIR patient out of the results of the REST API.

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
ibacher: Is this a placeholder tag? I was looking at the images and not really getting what from those images this implements.

chibongho: It's this divider for when there are multiple patients assigned to the same bed:

image

ibacher: Got it! I missed that one. Thanks!

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
brandones: However you should append a namespace for the key. We don't want random components around the app all having their key equal the patient UUID.


```suggestion
          <div key={`occupied-bed-pt-${patient.uuid}`}>
```

packages/esm-ward-app/src/beds/occupied-bed.scss
brandones: I didn't know about this! Does it compile faster, or result in better output, or both?

@denniskigen should we add something about this to the docs, if we have docs on SCSS/style?

packages/esm-ward-app/src/config-schema.ts
brandones: I see where you're coming from with `PatientCardRowElement` @chibongho but unfortunately I think the formal logic of the name does not align with intuition—I think it is very easy to read `PatientCardRowElement` and think the element is supposed to be a row. I think @mseaton 's suggestion of `patientCardElement` is clearest.

brandones: So this would be `patientCardElementDefinitions`

packages/esm-ward-app/src/config-schema.ts
brandones: @mseaton This is the way we handle this kind of thing in registration and service queues; it's the best pattern we've figured out for this.

@chibongho You will definitely want to add a `_default` value for `addressFields` so that it is optional, since indeed it would be very bad if we required implementers to configure `addressFields` for an element of type `bed-number`. We want implementers to be able to provide config like

```json
{
  "patientCardElementDefinitions": [
    { "id": "bednum", "elementType": "bed-number" },
    { "id": "addr", "elementType": "patient-address", "config": { "addressFields": ["cityVillage"] } } 
  ]
}
```

and then we can add validators that ensure that implementers *don't* provide an element config that doesn't make sense with the element type.

@mseaton The only way I can imagine supporting what you're talking about would be to have a separate definitions array for each type:

```json
{
  "patientCardBedNumberElementDefinitions": [
    { "id": "bednum" }
  ],
  "patientCardPatientAddressElementDefinitions": [
    { "id": "addr", "addressFields": ["cityVillage"] }
  ]
}
```

So that would be our other option here.

packages/esm-ward-app/src/beds/occupied-bed.scss
ibacher: Theoretically, it compiles faster and produces more predictable output. However, the strongest reason is that the Sass team has said they're planning on deprecating `@import` altogether. [See here for more](https://sass-lang.com/documentation/at-rules/import/). (I say theoretically, because the `vars` file itself doesn't really add any bloat, but `@import` parts of Carbon definitely would).

packages/esm-ward-app/src/config-schema.ts
brandones: Given the frequency with which this pattern appears, I'm definitely open to thinking about whether we can add a schema feature that supports it better. It would be really nice if the config system had some way of knowing what element config schema should be used based on the element type.

packages/esm-ward-app/src/config-schema.ts
ibacher: Yeah, this is a situation which the config schema doesn't handle very elegantly. Do we really have a use-case, though, for cards with different elements that display different parts of an address?

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -1,5 +1,120 @@
...
+              _description: 'The ID of the (bulit-in or custom) bento element',
+              _validators: [
+                validator(
+                  (bentoElementId: string) => {
+                    const validBentoElementIds: string[] = [...bentoElementTypes];
+                    return validBentoElementIds.includes(bentoElementId);
+                  },
+                  (bentoElementId: string) => 'Invalid bento element id: ' + bentoElementId,
+                ),
+              ],
```
brandones, line 72:
```suggestion
              _validators: [validators.oneOf(bentoElementTypes)],
```

(I think that syntax is right, sorry if it's not quite)

If this gives you a strange type error try upgrading core; I fixed the typing of that validator in the last week or two.

packages/esm-ward-app/src/ward-patient-card/ward-pateint-card.component.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
brandones, line 1: file name has a typo

packages/esm-ward-app/src/config-schema.ts
ibacher: Quick first pass. This usually comes up in the context of arrays of elements right? Maybe for arrays we can support something like the following:

```js
someKey: {
  _type: Type.Array,
  _default: [],
  _key: {
    id: {
      _type: Type.String,
      _description: 'The config-internal identifier for this element',
    },
  }
  _elements: [{
    id: 'address',  // applies to element whose key matches this
    elementType: {
      _type: Type.String,
      _description: 'The bento element type',
      _validators: [validators.oneOf(bentoElementTypes)],
    },
    config: {
        _type: Type.Array,
        _description: 'For bentoElementType "patient-address", defining which address fields to show',
    }
  }, {
    elementType: {
      _type: Type.String,
      _description: 'The bento element type',
      _validators: [validators.oneOf(bentoElementTypes)],
    },
  }]
}
```

That could be typed like:

```ts
type ConfigElement = {
  id: string;
  elementType: string;
}

type AddressConfigElement = ConfigElement & {
  id: 'address';
  config: Array<string>;
}

someKey: Array<ConfigElement>;
```

chibongho:
Updated with PR feedback and changed the configuration to allow for multiple rows. Hopefully that simplifies things.

packages/esm-ward-app/src/beds/occupied-bed.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { mockAdmissionLocation } from '../../../../__mocks__/wards.mock';
+import { bedLayoutToBed, filterBeds } from '../ward-view/ward-view.resource';
+import { getDefaultsFromConfigSchema, useConfig } from '@openmrs/esm-framework';
+import { configSchema, defaultPatientCardElementConfig } from '../config-schema';
+
+const defaultConfigSchema = getDefaultsFromConfigSchema(configSchema);
+
+jest.mocked(useConfig).mockReturnValue({
+  ...defaultConfigSchema,
+});
```
brandones, line 13:
```suggestion
const defaultConfig = getDefaultsFromConfigSchema(configSchema);

jest.mocked(useConfig).mockReturnValue({
  ...defaultConfig,
});
```

mseaton (approved):
This looks great to me @chibongho , thanks!  I'm still a little unsure about the address configuration, but it sounds like it has been given lots of thought and attention, so defer more to the experts on that, and if this is the best practice identified, it works for me.

PR #1193, opened 2024-06-17
(feat) O3-3215 - ward app - add patient card element to display obs f…

chibongho:
…or a particular concept

## Requirements

## Summary

Adds a configurable element to display an obs value.  See mockup:
https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda1a0534bb1e5f88a6640

## Screenshots

image


mseaton (changes requested):
This does not cover all the use cases described in https://openmrs.atlassian.net/browse/O3-3215 , and I would not expect the default behavior to limit to only obs within the visit .

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-obs.tsx
```diff
@@ -0,0 +1,38 @@
+import { useVisit } from '@openmrs/esm-framework';
+import React from 'react';
+import { type PatientObsElementConfig } from '../../config-schema';
+import { type Encounter, type WardPatientCardElement } from '../../types';
+import styles from '../ward-patient-card.scss';
+import { SkeletonText } from '@carbon/react';
+
+// prettier-ignore
+export const visitCustomRepresentation =
```
brandones, line 9: Good idea, I like this

chibongho:
Added the configs specified in https://openmrs.atlassian.net/browse/O3-3215.

brandones:
@chibongho , I commented on the Jira ticket to this effect, but I think this would be much easier both to use and maintain if the bentobox obs and the risk factors row were totally separate things. They are very different pieces of UI with very different normative usage patterns.

packages/esm-ward-app/src/hooks/useObs.ts
```diff
@@ -0,0 +1,22 @@
+import useSWR from 'swr';
+import { type Observation } from '../types';
+import { type Link, openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+
+type QueueEntrySearchCriteria = Record<'patient', string> | Record<'encounter', string> | Record<'q', string>;
```
mseaton, line 5: What is this QueueEntrySearchCriteria here for?

packages/esm-ward-app/src/hooks/useObs.ts
```diff
@@ -0,0 +1,22 @@
...
+
+// TODO: right now, the obs endpoint can only take in one of "patient", "encounter" and "q' as param
+// We should make it able to query by patient AND concept type
```
mseaton, line 8: Wow, that Obs endpoint is definitely in need of work.  Let's ticket that.

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
```diff
@@ -1,23 +1,18 @@
...
 import WardPatientCard from '../ward-patient-card/ward-patient-card';
+import { type WardBedProps } from '../ward-view/ward-bed.component';
+import styles from './occupied-bed.scss';
 
-export interface OccupiedBedProps {
-  patients: Patient[];
-  bed: Bed;
-}
-const OccupiedBed: React.FC<OccupiedBedProps> = ({ patients, bed }) => {
+const OccupiedBed: React.FC<WardBedProps> = ({ patientInfos, bed }) => {
```
mseaton, line 8: What is `patientInfos`?

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-obs.tsx
```diff
@@ -0,0 +1,73 @@
...
+    const {
+      conceptUuid,
+      onlyWithinCurrentVisit,
+      orderBy,
+      limit,
+      label,
+      labelI18nModule: labelModule,
+      displayType,
+    } = config;
+    const { data, isLoading } = useObs({ patient: patient.uuid }, obsCustomRepresentation);
```
mseaton, line 29: This overall looks great, but this in particular is not going to perform, assuming I am understanding that this gets _all_ obs for the patient, and then filters down to those of interest.  We will definitely need a more efficient endpoint that does much of this at the DB +/- server level.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -61,7 +61,9 @@ const WardViewByLocation = ({ location }: { location: Location }) => {
         {bedLayouts.map((bedLayout, i) => {
           const { patient } = bedLayout;
```
mseaton, line 62: This will be a list of patients once we fix bed management

mseaton (approved):
Overall looks good.  Certainly no real reason not to commit and then ticket any necessary improvements.  The obs endpoint and client-side filtering of all obs for a patient needs to be ticketed and prioritized.

packages/esm-ward-app/src/hooks/useObs.ts
chibongho: copy pasted error. Should have been `ObsSearchCriteria`

packages/esm-ward-app/src/hooks/useObs.ts
chibongho: What @mogoodrich suggested works (thanks!). I'm changing it to pass in just patient and concept

packages/esm-ward-app/src/beds/occupied-bed.component.tsx
chibongho: It's an array of compound object that, as of now, only has the patient and the active visit object. It'll have more things like encounters soon.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-obs.tsx
chibongho: Shouldn't be a concern after I change `useObs` to also take in the concept as Mark suggested.

chibongho:
> @chibongho , I commented on the Jira ticket to this effect, but I think this would be much easier both to use and maintain if the bentobox obs and the risk factors row were totally separate things. They are very different pieces of UI with very different normative usage patterns.

@brandones I'll admit this is a half-hearted attempt at also having the [risk factor stuff ](https://openmrs.atlassian.net/browse/O3-3222) done in this PR. I'll remove the `displayType` thing for now. That said, I'm not convinced that the config and component for this and the risk factor stuff is really that different, especially since we are now going with modelling a custom row with a just a bentobox row with one element.

PR #1158, opened 2024-05-28
(feat) O3-3122: Improve UI for Queue by status view in the service queues

vasharma05:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary
Improved UI for "Queue entries by Queue and status" in the service queues dashboard.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/51502471/83513895-1f20-4207-87b0-343c0d0145c3

## Related Issue
https://issues.openmrs.org/browse/O3-3122

## Other
None

PR #1196, opened 2024-06-18
(feat) O3-3123: Improve the queue status metrics on the service queue page

vasharma05:
## Requirements

## Summary
This PR makes the following changes to the Queue status metrics:
1. Queue status metric header is an extension slot with slot name: `queue-table-metrics-slot`
2. Added extensions for displaying metric cards for the following statuses: Waiting, In service and Finished service (Based on the reference data)
3. The queue metric card header will redirect and display status specific queue entry table.

## Screenshots

https://github.com/openmrs/openmrs-esm-patient-management/assets/51502471/1e9a9f9e-79ae-48cb-a6a8-8a21a44ec404

## Related Issue
https://issues.openmrs.org/browse/O3-3123

## Other
Built over PR #1158


mseaton (changes requested):
If this is the only way to implement this, we probably need to rethink it entirely.  This is basically taking us back to a point that we already refactored away from, which is hard-coding particular notions of what statuses should be rendered into the frontend.

For this particular view, we always know what queue we are viewing - this page is for a specific queue.  For that queue, there are certain statuses that are available, and this can vary on a queue-by-queue basis.

We should not be hard-coding particular statuses, but rendering dynamic extensions into slots based on the statuses that are available.

packages/esm-service-queues-app/src/queue-table-metrics-extensions/in-service-status.extension.tsx
```diff
@@ -0,0 +1,20 @@
...
+export default function InServiceStatusQueueMetrics({ queueUuid, statusUuid }: InServiceStatusQueueMetricsProps) {
+  const config = useConfig<ConfigObject>();
+
+  if (!config?.concepts?.inServiceQueueStatusConceptUuid) {
+    console.error(new Error('No concept UUID found for Queue Status = In service'));
+    return null;
+  }
+
+  return <StatusQueueMetrics queueUuid={queueUuid} statusUuid={config?.concepts?.inServiceQueueStatusConceptUuid} />;
+}
```
mseaton, line 20: Same comment as above.  Delete this.

packages/esm-service-queues-app/src/queue-table-metrics-extensions/finished-service-status.extension.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+  if (!config?.concepts?.finishedServiceQueueStatusConceptUuid) {
+    console.error(new Error('No concept UUID found for Queue Status = Waiting for transfer'));
+    return null;
+  }
+
+  return (
+    <StatusQueueMetrics queueUuid={queueUuid} statusUuid={config?.concepts?.finishedServiceQueueStatusConceptUuid} />
+  );
+}
```
mseaton, line 25: I don't understand the point of this file. How does this add any value above the existing StatusQueueMetrics component? Why does this both pull a status concept from config and take one in as a prop? It should only take one in as a prop, and that is exactly what the StatusQueueMetrics already does. Delete this file.

packages/esm-service-queues-app/src/queue-table-metrics-extensions/total-patients-in-queue.extension.tsx
```diff
@@ -0,0 +1,25 @@
...
+  });
+
+  return (
+    <QueueTableMetricsCard
+      value={count}
+      headerLabel={t('totalPatients', 'Total Patients')}
+      linkTo={`\${openmrsSpaBase}/home/service-queues/queue-table-by-status/${queueUuid}`}
+    />
+  );
+}
```
mseaton, line 25: You shouldn't need this class.  Just change the `StatusQueueMetrics` component -> `QueueMetrics` component and make the `status` prop optional.  Then render the appropriate metric based on queue+status or just queue.

packages/esm-service-queues-app/src/queue-table-metrics-extensions/waiting-status.extension.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default function WaitingStatusQueueMetrics({ queueUuid, statusUuid }: WaitingStatusQueueMetricsProps) {
+  const config = useConfig<ConfigObject>();
+
+  if (!config?.concepts?.waitingQueueStatusConceptUuid) {
+    console.error(new Error('No concept UUID found for Queue Status = Waiting'));
+    return null;
+  }
+
+  return <StatusQueueMetrics queueUuid={queueUuid} statusUuid={config?.concepts?.waitingQueueStatusConceptUuid} />;
+}
```
mseaton, line 23: Same as earlier comments.  Delete this class.

packages/esm-service-queues-app/src/queue-table/queue-table-by-status-skeleton.component.tsx
```diff
@@ -2,31 +2,31 @@ import { DataTableSkeleton, SkeletonText, Tab, TabList, TabPanel, TabPanels, Tab
...
 
   return (
     <div className={styles.container}>
-      <div className={styles.headerContainer}>
-        <div className={!isDesktop(layout) ? styles.tabletHeading : styles.desktopHeading}>
-          <h3>
-            <SkeletonText />
-          </h3>
-        </div>
+      <div className={styles.statusTableContainer}>
```
mseaton, line 12: This seems to make the assumption that there are always 3 statuses.  If you are doing what I think you are doing here, you should add a DataTableSkeleton for each status configured on the queue, not just 3 arbitrarily.

packages/esm-service-queues-app/src/queue-table/queue-table-metrics.component.tsx
```diff
@@ -1,29 +1,20 @@
...
+import React from 'react';
 import styles from './queue-table-metrics.scss';
-import { type Queue } from '../types';
+import type { Concept, Queue } from '../types';
+import { ExtensionSlot } from '@openmrs/esm-framework';
 
 interface QueueTableMetricsProps {
   selectedQueue: Queue;
+  selectedStatus: Concept;
 }
```
mseaton, line 10: This is where we need to figure out a way to render extensions based on what extensions are configured on the slot, and what statuses are available on the current queue.  We _may_ be able to use the new `useRenderableExtensions` function from esm-framework (esm-react-utils) to achieve this.  I'd be interested in @chibongho  and @ibacher given the similar work they recently tried to achieve with the queue table.

packages/esm-service-queues-app/src/routes.json
```diff
@@ -102,6 +102,30 @@
...
+      "name": "in-service-status-queue-metrics",
+      "component": "inServiceStatusQueueMetrics",
+      "slot": "queue-table-metrics-slot",
+      "order": 2
+    },
+    {
+      "name": "finished-service-status-queue-metrics",
+      "component": "finishedServiceStatusQueueMetrics",
+      "slot": "queue-table-metrics-slot",
+      "order": 3
```
mseaton, line 128: We don't want to hard-code these here.  This should be removed in favor of some other technique that does not require hard-coding particular metadata or statuses into our default configurations.

mseaton (changes requested):

vasharma05:
> We should not be hard-coding particular statuses, but rendering dynamic extensions into slots based on the statuses that are available.

Hi @mseaton!
This is what I had in mind, we want to use the `allowedStatus` for the given queue, which is dynamic. Now, creating extensions dynamically and then attaching the extensions to the extension slot seems to doing the task the longer way. If the need is to only hide a few statuses, I believe we can do so my configuration, defining which statuses to hide. We can make the header an extension slot, but rendering dynamic extensions is still doubtful to me.

I might be missing something, looking forward for your reply.
Thanks!

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -118,12 +135,26 @@ export const configSchema = {
...
+    inServiceQueueStatusConceptUuid: {
+      _type: Type.ConceptUuid,
+      _description: 'The UUID of the concept that represents Status = In Service.',
+      _default: '167408AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
+    },
+    finishedServiceQueueStatusConceptUuid: {
+      _type: Type.ConceptUuid,
+      _description: 'The UUID of the concept that represents Status = Finished service.',
+      _default: 'b559fb77-4e1e-4285-b9b7-1d03e0ba983f',
+    },
```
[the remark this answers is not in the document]
brandones: They are the default values corresponding to the default implementation, which is the reference application. Having default values allows us to ship an EMR which is a working product, which can then be configured by organizations that wish to use custom metadata. If it's configurable, it's not hard-coded. Whether these defaults are present or not makes absolutely no difference to how PIH's config will work, but it allows organizations that are not PIH to have a much more pleasant experience getting set up with OpenMRS.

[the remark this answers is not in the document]
brandones: (Deleted comment because I'd misunderstood something)

packages/esm-service-queues-app/src/index.ts
```diff
@@ -156,6 +156,29 @@ export const patientSearchWorkspace = getAsyncLifecycle(() => import('./patient-
...
+export const inServiceStatusQueueMetrics = getAsyncLifecycle(
+  () => import('./queue-table-metrics-extensions/in-service-status.extension'),
+  { featureName: 'in-service-status-queue-metrics', moduleName },
+);
+
+export const finishedServiceStatusQueueMetrics = getAsyncLifecycle(
+  () => import('./queue-table-metrics-extensions/finished-service-status.extension'),
+  { featureName: 'finished-service-status-queue-metrics', moduleName },
+);
+
```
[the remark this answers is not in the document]
brandones: I agree that there should be a single component configurable in the way @mseaton describes, using extension config.

However, I'd also suggest that these three extensions should also still be here, as thin wrappers around the configurable component which provide the correct default configuration for the RefApp. So the RefApp ships with these three attached, and then if someone wants to swap them out they just use the generic queue metric extension.

It would be nice if we had a better way of dealing with this kind of thing (attaching an extension to a slot, possibly multiple times, with configuration set for the attachment), but this is how we've been doing it so far.

brandones:
Yeah I don't think we should be "attaching extensions dynamically." If something should be setting its contents dynamically, it should not be an extension slot. (Extensions that set their own contents dynamically, on the other hand, are great and fine).

So one way of solving this would be to not use extensions at all, but just have the metrics section be a component that renders cards based on information from the API and the config. That gives a lot of flexibility about how to implement this.

Using the extension system puts much stronger constraints about how a thing ought to be implemented. If we want to use extensions for this, then I strongly recommend something close the following structure.
- A generic QueueMetricCard component which takes queue and status as props.
- A generic queue-metric-card extension which uses [`defineExtensionConfigSchema`](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-framework/docs/API.md#defineextensionconfigschema) to allow itself to be configured to show a queue and a status. It is a thin wrapper around the QueueMetricCard component which passes in the config as props.
- Three extensions for the default queue metrics for the RefApp. These are thin wrappers around the QueueMetricCard which pass in the correct queue and status info for the RefApp. I think it would not be unreasonable to also make these configurable using `defineExtensionConfigSchema`, so that their queue and concept UUIDs are not hard-coded, but can be changed.

brandones:
Closing as stale.

PR #1199, opened 2024-06-20
(fix) for O3-2827: Appointments: Refactor (Part I ?)

mogoodrich:
## Requirements

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.scss
+++ b/packages/esm-appointments-app/src/form/appointments-form.scss
@@ -4,31 +4,10 @@
 
 $openmrs-background-grey: #ededed;
 
-.formWrapper {
-  display: flex;
-  flex-direction: column;
-  justify-content: space-between;
-}
-
-:global(.omrs-breakpoint-lt-desktop) .formWrapper {
-  background-color: $openmrs-background-grey;
-  margin-top: spacing.$spacing-09;
-}
-
-:global(.omrs-breakpoint-gt-tablet) .formWrapper {
-  background-color: colors.$white;
-}
-
 .formGroup {
   display: flex;
   margin-bottom: spacing.$spacing-02;
   padding: spacing.$spacing-05;
-
-  checkbox {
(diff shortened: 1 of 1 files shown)


packages/esm-appointments-app/src/form/appointments-form.scss
```diff
@@ -97,17 +76,3 @@ $openmrs-background-grey: #ededed;
   max-inline-size: fit-content;
   padding-top: spacing.$spacing-03;
 }
-
-:global(.cds--select-input):focus {
-  outline: 2.5px solid colors.$red-60;
-}
```
mogoodrich, line 103: This (and the one similar ones below) were the problematic cases.  They existed previous to the refactor, but one result of the refactor was that the appointment form was being used in multiple places, so there are now more times when this global override was being applied.  

It looked like the main use case here was to make sure a field was highlighted in red when the form submittal handler set focus on the input (say if for instance it was a mandatory field that had not been selected).  I considered adding a wrapper or something so that this would *only* be applied to inputs on this form, but I realized that we probably don't want to apply this non-standard UI pattern.  Now, when a mandatory field does not have a value, it still switched focus to that field and the highlights it, just in the default outline color, not red.

usamaidrsk (approved):

brandones:
@mogoodrich Please include screenshots or video in your PRs.

mogoodrich:
Screen uploaded.  The only change in this PR should be that when focus shifts to a field in the appointments form it is highlighted in blue, not red.  This color change had also errorneously been applied globally. See my comments above for more details.

mogoodrich:
Thanks @usamaidrsk !  Screenshot added above.  I am starting on a 2 week vacation starting tomorrow, so going to go ahead and merge this in so I confirm there are no issues before I take off.

brandones:
Thanks @mogoodrich , much appreciated. Enjoy your vacation!

ojwanganto:
Thanks, @mogoodrich

PR #1205, opened 2024-06-22
Chore/more robust refetch mechanism for service queues #3263

PiusKariuki:
## Requirements

## Summary
This PR addresses an incosistency in re-retching the table contents in the services queues page when a write operation like adding a patient, clearing the queue or transitioning the patient fails to repopulate the table with the updated data.
We achieve this by employing custom events
```
export const emitRefetchQueuesEvent = (queueUUID: string = '') => {
  const event = new CustomEvent('refetchQueues', { detail: queueUUID });
  window.dispatchEvent(event);
};
```

## Related Issue
https://openmrs.atlassian.net/browse/O3-3418

### DIFF
--- a/packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
+++ b/packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
@@ -35,7 +35,7 @@ This is used in the main dashboard of the queues app. (Currently behind a featur
 function DefaultQueueTable() {
   const selectedQueueUuid = useSelectedServiceUuid();
   const currentLocationUuid = useSelectedQueueLocationUuid();
-  const { queueEntries, isLoading, isValidating, error } = useQueueEntries({
+  const { queueEntries, isLoading, isValidating, error, mutate } = useQueueEntries({
     queue: selectedQueueUuid,
     location: currentLocationUuid,
     isEnded: false,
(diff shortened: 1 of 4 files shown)


packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -83,6 +83,15 @@ function DefaultQueueTable() {
...
   }, [queueEntries, searchTerm]);
 
+  useEffect(() => {
+    const handlePatientAddedToQueue = () => mutate();
+
+    window.addEventListener('refetchQueues', handlePatientAddedToQueue);
+    return () => {
+      window.removeEventListener('refetchQueues', handlePatientAddedToQueue);
+    };
+  }, []);
```
vasharma05, line 93: `mutate` will be added as a dependency`

vasharma05:
Hi @PiusKariuki!
I reviewed the changes you made in the PR and have the following points to put forward:
`useQueueEntries` is the hook that fetches the queue entries. I see you are calling the `mutate` for the default queue table. Now, if you note, the `mutate` function not only needs to revalidate the response for the given `service` and `status` but for all the statuses and services.

Alongside, please note that the default queue table is only used on the service queues home page, but we have more pages in the service queues (refer to `packages/esm-service-queues-app/src/root.component.tsx`) which will not be refreshed in this scenario. The `QueueTable` component is the reusable component for every table rendered inside the service queues app.

My one more concern is the use of events in the same app, where we only want to update the SWR Cache of the same app, this seems to be an approach, that is alright to use, but still goes beyond the current practices.

Thanks!

PiusKariuki:
Thanks @vasharma05 for these pointers that I had not taken into account. I am going to use them to rethink my implementation then try again.
Thank you!

brandones (changes requested):
This needs documentation. I commented in the ticket to this effect. There is no documentation of the bug—screenshots, video, reproduction instructions, nothing. And there is no documentation that this fixes any bug (video or screenshots that show it working). Finally, if we are getting rid of `useMutateQueueEntries`, we should do a proper job and delete it entirely. However it's likely that if there is a bug, the thing we want to do is simply to fix `useMutateQueueEntries` and ensure it is used correctly everywhere.

PR #1212, opened 2024-06-25
(fix): Update date picker props and add mock for date picker

NethmiRodrigo:
## Requirements

## Summary

This PR updates the props passed on to the `OpenmrsDatepicker` and adds a mock date picker to the test case. Also bumps up the openmrs framework.

## Related Issue

https://issues.openmrs.org/browse/O3-3470

## Other

To Do:

- [x] Have to fix the failing DOB test case by changing how the test case updates the value and how it tries to get the field.
- [x] Have to add to the DOB component a div that renders the error text shown when the field is invalid as it is no longer handled by the date picker itself.

### DIFF
--- a/e2e/specs/register-new-patient.spec.ts
+++ b/e2e/specs/register-new-patient.spec.ts
@@ -31,8 +31,10 @@ test('Register a new patient', async ({ page, api }) => {
     await patientRegistrationPage.waitUntilTheFormIsLoaded();
   });
 
-  await test.step('And then I click on fill new values into the registration form and then click the `Submit` button', async () => {
-    await patientRegistrationPage.fillPatientRegistrationForm(formValues);
+  test.describe.fixme('Broken test due to the openmrs date picker and should be fixed', async () => {
+    await test.step('And then I click on fill new values into the registration form and then click the `Submit` button', async () => {
+      await patientRegistrationPage.fillPatientRegistrationForm(formValues);
+    });
   });
 
   await test.step("Then I should be redirected to the new patient's chart page and a new patient record should be created from the information captured in the form", async () => {
--- a/e2e/specs/register-new-patient.spec.ts
+++ b/e2e/specs/register-new-patient.spec.ts
@@ -6,7 +6,8 @@ import { deletePatient, getPatient } from '../commands';
 
 let patientUuid: string;
 
-test('Register a new patient', async ({ page, api }) => {
+// TODO: O3-3482 Broken due to the date picker and should be fixed
+test.skip('Register a new patient', async ({ page, api }) => {
   test.setTimeout(5 * 60 * 1000);
   const patientRegistrationPage = new RegistrationAndEditPage(page);
 
@@ -31,10 +32,8 @@ test('Register a new patient', async ({ page, api }) => {
     await patientRegistrationPage.waitUntilTheFormIsLoaded();
   });
 
-  test.describe.fixme('Broken test due to the openmrs date picker and should be fixed', async () => {
-    await test.step('And then I click on fill new values into the registration form and then click the `Submit` button', async () => {
-      await patientRegistrationPage.fillPatientRegistrationForm(formValues);
-    });
+  await test.step('And then I click on fill new values into the registration form and then click the `Submit` button', async () => {
+    await patientRegistrationPage.fillPatientRegistrationForm(formValues);
   });
 
   await test.step("Then I should be redirected to the new patient's chart page and a new patient record should be created from the information captured in the form", async () => {
@@ -48,7 +47,7 @@ test('Register a new patient', async ({ page, api }) => {
 
     await expect(person.display).toBe(`${givenName} ${middleName} ${familyName}`);
     await expect(person.gender).toMatch(new RegExp(sex[0], 'i'));
-    await expect(dayjs(person.birthdate).format('DD/MM/YYYY')).toBe(formValues.birthdate);
+    // await expect(dayjs(person.birthdate).format('DD/MM/YYYY')).toBe(formValues.birthdate);
     await expect(person.preferredAddress.address1).toBe(formValues.address1);
     await expect(person.preferredAddress.cityVillage).toBe(formValues.cityVillage);
     await expect(person.preferredAddress.stateProvince).toBe(formValues.stateProvince);
--- a/e2e/specs/register-new-patient.spec.ts
+++ b/e2e/specs/register-new-patient.spec.ts
@@ -7,52 +7,54 @@ import { deletePatient, getPatient } from '../commands';
 let patientUuid: string;
 
 // TODO: O3-3482 Broken due to the date picker and should be fixed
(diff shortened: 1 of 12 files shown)


ibacher:
Re: validation: Making an API more similar to Carbon's is on the TODO list, but for now, React-Aria supports a `validation` property that takes a function with this signature:

```ts
(date: CalendarDate) => true | string
```

Basically, the function takes the current value and either returns `true`, indicating no error, or a `string` that's the input text to display. The component should handle things from there.

NethmiRodrigo:
@ibacher regarding the DOB test case, we have a test case that does the below which fails because that string isn't a `CalendarDate` and brings up the error in the updated onDateChange function saying `birthDate.toDate is not a function`. Was wondering if I should check in the function if its a string and convert it to a date if so? What do you think is the best approach here?
```
await user.type(dateInput, '10/10/2022');
expect(dateInput).toHaveValue('10/10/2022');
```

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
```diff
@@ -46,8 +47,8 @@ export const DobField: React.FC = () => {
   );
 
   const onDateChange = useCallback(
-    (birthdate: Date) => {
-      setFieldValue('birthdate', birthdate);
+    (birthdate: CalendarDate) => {
```
chibongho, line 50: Shouldn't this be of type `DateValue`? I thought that's what the onChange callback expects.

chibongho (commented):
Not a blocker for this PR, but there is one more file (obs-field.component.tsx) in patient-management that's using `<OpenmrsDatePicker>` that needs changing as well.

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
NethmiRodrigo: I think that was before the framework was updated in the module.

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
```diff
@@ -104,17 +103,15 @@ export const DobField: React.FC = () => {
...
-              invalidText={birthdateMeta.error && t(birthdateMeta.error)}
+              isInvalid={!!(birthdateMeta.touched && birthdateMeta.error)}
               value={birthdate.value}
-              carbonOptions={{
-                placeholder: placeHolder,
-              }}
             />
+            {!!(birthdateMeta.touched && birthdateMeta.error) && (
+              <div className={styles.radioFieldError}>{birthdateMeta.error && t(birthdateMeta.error)}</div>
+            )}
```
vasharma05, line 114: Shouldn't we also make the error field part of the component itself?

package.json
```diff
@@ -17,6 +17,7 @@
   },
   "dependencies": {
     "@hookform/resolvers": "^3.3.1",
+    "@internationalized/date": "^3.5.4",
```
ibacher, line 20: It's probably preferable to inherit this from ems-framework unless it's required here?

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
ibacher: Yes... Upcoming feature.

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.test.tsx
```diff
@@ -36,18 +49,20 @@ describe('Dob', () => {
     expect(screen.getByRole('textbox', { name: /date of birth/i })).toBeInTheDocument();
   });
 
-  it('typing in the date picker input sets the date of birth', async () => {
-    const user = userEvent.setup();
+  // TODO : Fix this test case.
```
ibacher, line 52: Instead of commenting this out, just use `xit()` in place of `it()`

ibacher (commented):
Thanks @NethmiRodrigo! A few small things, but basically LGTM

package.json
NethmiRodrigo: Would using it in in the dob component for the onChange function count as required?

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.test.tsx
NethmiRodrigo: Sorry Ian, I had replaced it with `it.skip` before seeing this comment, would that suffice, or if `xit` preferred?

ibacher: Oh! That's also fine to... I thought jest didn't support `.skip` with `it` but I guess I'm wrong!

package.json
ibacher: I don't think so; it's a nice-to-have, but you should get the dependency transitively from the styleguide. My worry here is that I had some issues with Yarn not resolving ^3.5.0 and ^3.5.4 to the same version in core that lead to a lot of typescript errors.

packages/esm-patient-registration-app/src/patient-registration/field/dob/dob.component.tsx
ibacher: Yeah, for the DatePicker, at any rate, this will always be a `CalendarDate`. Would it be helpful to have a wrapper here that instead returned a normal JS date?

chibongho: I'm indifferent about that, as it's just one (albeit non-obvious) line to convert `CalendarDate` to JS date. I'd be happy to see more utilities in core to help handle datetime in general though. (so things like combining the input values of `DatePicker` and `TimePicker` into a single Date is standardized)

gracepotma:
@eudson this would be a good time for your team to test this and share any feedback here ;)

chibongho:
E2E tests are still failing. (I think those are easy fixes though). I can go ahead and fix those.

e2e/specs/register-new-patient.spec.ts
```diff
@@ -6,7 +6,8 @@ import { deletePatient, getPatient } from '../commands';
 
 let patientUuid: string;
 
-test('Register a new patient', async ({ page, api }) => {
+// TODO: O3-3482 Broken due to the date picker and should be fixed
+test.skip('Register a new patient', async ({ page, api }) => {
```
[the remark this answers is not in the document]
NethmiRodrigo: Yeah I was supposed to wrap the fixme around this whole test but I was trying to see if I can edit the test to use age instead of dob.

package.json
NethmiRodrigo: Would you please mind guiding me through how to update the dependency? I'm not sure I know how

ibacher: What you have here is fine.

ibacher (approved):

PR #1218, opened 2024-06-27
(feat) O3-3223: Add configurable element to include alerts on pending items to patient card

usamaidrsk:
## Requirements

## Summary

Add pending inpatient request on the patient card

## Screenshots

image

## Other

[O3-3223](https://openmrs.atlassian.net/browse/O3-3223)

[O3-3223]: https://openmrs.atlassian.net/browse/O3-3223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brandones:
It remains obvious that [this is the wrong approach](https://github.com/openmrs/openmrs-esm-patient-management/pull/1193#issuecomment-2179015366). I wonder how bad we need to let it get before we can start re-implementing all this correctly. The idea of an "hourglass type element" is hilarious. It's like we're building a Lego set for an EMR, not an actual EMR.

Whenever this is ready for review I'll review it and we can put it in.

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
```diff
@@ -0,0 +1,56 @@
...
+    const encounterKey = `${restBaseUrl}/encounter/${order.encounter.uuid}?v=${encounterRepresentation}`;
+
+    const concept = await openmrsFetch<Concept>(conceptKey);
+    const encounter = await openmrsFetch<Encounter>(encounterKey);
+
+    return { concept, encounter };
+  };
+
+  useEffect(() => {
+    testOrders.forEach(async (order) => {
```
brandones, line 30: This is going to execute in series. We probably want to execute this in parallel. Use `Promise.all` and `map`.

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
```diff
@@ -0,0 +1,56 @@
...
+  'custom:(uuid,encounterDatetime,encounterType,location:(uuid,name),' +
+  'patient:(uuid,display),encounterProviders:(uuid,provider:(uuid,name)),' +
+  'obs:(uuid,obsDatetime,voided,groupMembers,formFieldNamespace,formFieldPath,order:(uuid,display),concept:(uuid,name:(uuid,name)),' +
+  'value:(uuid,display,name:(uuid,name),names:(uuid,conceptNameType,name))))';
+
+const conceptRepresentation =
+  'custom:(uuid,display,name,datatype,set,answers,hiNormal,hiAbsolute,hiCritical,lowNormal,lowAbsolute,lowCritical,units,' +
+  'setMembers:(uuid,display,answers,datatype,hiNormal,hiAbsolute,hiCritical,lowNormal,lowAbsolute,lowCritical,units))';
+
+export function useTestOrderCount(testOrders: Order[]) {
```
brandones, line 16: What does this function do? It isn't clear to me why this takes a `testOrders` object but then still needs to make a bunch of requests, especially if all it's doing is counting the number of them that are test orders... which if the param is called `testOrders` should be all of them. This is all very confusing. Please write a docstring and consider some better names.

brandones (dismissed):
Needs a lot of work.

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
```diff
@@ -0,0 +1,95 @@
...
+      return { concept, encounter };
+    };
+
+    const countPendingOrders = async () => {
+      setIsLoading(true);
+      try {
+        const pendingCounts = await Promise.all(
+          orders.map(async (order) => {
+            if (order.orderType['javaClassName'] === 'org.openmrs.TestOrder') {
+              const { concept, encounter } = await fetchOrderData(order);
```
mogoodrich, line 58: Is the assumption here that the test result obs is going to be on the same encounter as the order?  I don't think that will normally be the case, unfortunately.

mogoodrich: It also looked like according to the ticket this should be limited to orders placed on the current visit (which should hopefully simplify things)

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { useTranslation } from 'react-i18next';
+
+const WardPatientTransfer: WardPatientCardElement = () => {
+  const { t } = useTranslation();
+  // TODO update the transfer details by the active disposition
+  return (
+    <div className={styles.waitingForItemContainer}>
+      <Movement className={styles.movementIcon} size="24" />
+      {t('transferToDeliveryWard', 'Transfer to Delivery ward')}
+    </div>
```
mogoodrich, line 14: This is just a more or less a placeholder at this point, correct?

mogoodrich (commented):
Thanks @usamaidrsk !  I added a few comments, but as I'm not 100% familiar with this functionality, I'll hold on approving and wait for @mseaton  and @chibongho  to return.

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
```diff
@@ -0,0 +1,95 @@
...
+                const encounterData = encounter.data;
+
+                const testResultObs = encounterData?.obs?.find((obs) => obs.concept.uuid === conceptData.uuid);
+
+                if (conceptData.setMembers.length > 0) {
+                  return conceptData.setMembers.filter(
+                    (memberConcept: { uuid: string }) =>
+                      !testResultObs?.groupMembers?.find((obs) => obs.concept.uuid === memberConcept.uuid),
+                  ).length;
+                } else if (conceptData.setMembers.length === 0 && !testResultObs?.value['display']) {
```
brandones, line 71: What's that second condition, `!testResultObs?.value['display']`, about?

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
brandones: It looks like it. @usamaidrsk ?

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
brandones: @usamaidrsk please answer Mark's first question

I would be inclined to have the caller of this hook handle filtering orders down to only the ones placed on this visit. The filtered list of orders would be passed in as the argument. So here the change would go in ward-patient-pending-orders. @usamaidrsk can you make that change?

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
usamaidrsk: The  condition checks if the `testResultObs` has no `groupMembers`. If so, it verifies the `testResultObs` value to determine whether the lab order is done or pending.

I borrowed this implementation from [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/59934a3330a02eb0e41e79df4cabedacdda77f52/packages/esm-patient-orders-app/src/components/test-order.component.tsx#L51)  cc @brandones

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
mogoodrich: Oh interesting, I missed let, I will check out ticket and comment....

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
usamaidrsk: > Is the assumption here that the test result obs is going to be on the same encounter as the order?

Looking at a similar implementation from the [here](https://github.com/openmrs/openmrs-esm-patient-chart/blob/59934a3330a02eb0e41e79df4cabedacdda77f52/packages/esm-patient-orders-app/src/components/test-order.component.tsx#L28), looks like that is the case.

usamaidrsk: > I would be inclined to have the caller of this hook handle filtering orders down to only the ones placed on this visit. The filtered list of orders would be passed in as the argument. So here the change would go in ward-patient-pending-orders. @usamaidrsk can you make that change?

Will do @brandones. Thanks

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
mogoodrich: Okay @usamaidrsk , I updated the ticket, you are right, we can't display the TRANSER or DISCHARGE until we finish https://openmrs.atlassian.net/browse/O3-3648 , which I hope to start on shortly.

packages/esm-ward-app/src/hooks/usePatientPendingOrders.ts
usamaidrsk: > It also looked like according to the ticket this should be limited to orders placed on the current visit (which should hopefully simplify things)

Is it possible querying orders by a specific visit encounter? cc @mogoodrich @brandones 

Note: to fetch orders I used the [usePatientOrders](https://github.com/openmrs/openmrs-esm-patient-chart/blob/59934a3330a02eb0e41e79df4cabedacdda77f52/packages/esm-patient-common-lib/src/orders/useOrders.ts#L16) from the `esm-patient-common-lib`

mogoodrich: Looks like you might be correct about the test orders, let me start a discussion on Slack (or elsewhere)

brandones: The discussion on Slack: https://openmrs.slack.com/archives/CHP5QAE5R/p1721851387699259

brandones: @mogoodrich , do you have suggestions about implementation? Based on the Slack thread, it sounds like we may need to query obs by `order_id`?

mogoodrich: @mseaton  had suggested on the ticket (https://openmrs.atlassian.net/browse/O3-3223) moving this to the backend, which offhand made sense to me, I was planning on following up with him today and then one of us can update the ticket/PR.

packages/esm-ward-app/src/types/index.ts
```diff
@@ -141,7 +159,7 @@ export interface Encounter extends OpenmrsResourceStrict {
   voided?: boolean;
   visit?: Visit;
   encounterProviders?: Array<EncounterProvider>;
-  diagnoses?: any;
+  diagnoses?: Array<Diagnosis>;
```
mogoodrich, line 162: Is this diagnosis change tangential to the commit, or is this used somewhere within this commit, perhaps I'm missing it?

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-pending-orders.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useTranslation } from 'react-i18next';
+import { usePatientOrders } from '@openmrs/esm-patient-common-lib';
+import { SkeletonIcon } from '@carbon/react';
+import { useCountTestOrdersWithoutObs } from '../../hooks/usePatientPendingOrders';
+
+interface WardPatientPendingOrdersProps extends WardPatientCardProps {
+  orderType: string;
+}
+
+const WardPatientPendingOrdersItem: React.FC<WardPatientPendingOrdersProps> = ({ patient, orderType }) => {
```
mogoodrich, line 15: Let's just make this a dummy element now like the WardPatientTransfer

mogoodrich (changes requested):
Hey @usamaidrsk ... sorry for all the back-and-forth on this, and the lack of detail in the original ticket description, we probably should not have marked this ready for work until we had all the backend stuff figured out.

See my comments on the PR, but can we strip out all the pending order stuff (sorry for all the work you put in) and make this a place holder like the transfer stuff for now? 

I also had a question about the diagnosis changes that seem to be tangential? See my comment on the PR.

brandones:
As a note, we generally prefer to avoid adding "dummy" components. Just implement working components once it is possible to do so. This is for a few reasons:
- Having parts of the UI that look like they should do something but simply don't is confusing
- Having pieces of code that look like they do something real but actually do nothing meaningful is confusing
- Code is read 10x for every time it's written. Dummy components create more stuff to maintain
- If requirements change, it's not unlikely that the dummy gets forgotten about and either stays as a broken piece of UI or a piece of dead code that isn't used anywhere
- Even if requirements don't change, there's always a pretty good chance that the way you thought you were going to implement it won't actually work, once you find out how it needs to be wired up. The original dummy implementation just becomes a burden, additional cognitive load, something to be worked around or thrown out after some prevarication.
- A component that doesn't work is a bug. Shipping something which is lacking features is fine; shipping bugs is not fine.

mogoodrich:
That's all fair @brandones , I'm not a fan of dummy components either... I was just filling a little bad about having this ticket put up for work before it had proper tech design, and about having this PR hanging around for a long time before it's ready to be merged... @usamaidrsk are you okay with holding on this until we get the tech design for fetching the relevant data figured out?

usamaidrsk:
> . @usamaidrsk are you okay with holding on this until we get the tech design for fetching the relevant data figured out?

No problem @mogoodrich , we can hold.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
```diff
@@ -0,0 +1,35 @@
...
+  wardPatient: WardPatient;
+}
+
+const WardPatientTransfer: React.FC<WardPatientTransferProps> = ({ wardPatient }) => {
+  const { t } = useTranslation();
+
+  const { dispositionType } = wardPatient?.inpatientRequest;
+  const message = useMemo(() => {
+    if (dispositionType === 'TRANSFER') {
+      return t('transferToDeliveryWard', 'Transfer to Delivery ward');
```
mogoodrich, line 17: Sorry, we weren't clear but instead of being hardcoded to Delivery Ward, this should display wardPatient.inpatientRequest.dispositionLocation.name.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+    if (dispositionType === 'DISCHARGE') {
+      return t('transferToDeliveryWard', 'Transfer to Delivery ward');
```
mogoodrich, line 20: We will change this to "Pending Discharge" per our discussion today on standup

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
```diff
@@ -0,0 +1,35 @@
...
+    }
+    return '';
+  }, [dispositionType]);
+
+  if (dispositionType === 'ADMIT') return null;
```
mogoodrich, line 25: Just to be safe if we start adding other dispositions, let's make this display null for any dispositionType (or null dispositionType) other than TRANSFER and DISCHARGE.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx
```diff
@@ -32,6 +34,12 @@ const WardPatientCard: WardPatientCard = (wardPatient) => {
         ))}
         <ExtensionSlot name={headerExtensionSlotName} state={wardPatient} />
       </div>
+      {wardPatient?.inpatientRequest ? (
+        <div className={styles.wardPatientCardPendingItemsRow}>
+          <Hourglass className={styles.hourGlassIcon} size="24" />:
+          <WardPatientTransfer wardPatient={wardPatient} />
+        </div>
+      ) : null}
```
mogoodrich, line 42: @chibongho  thoughts on whether we can hard code this into the card or do we need/should we make this configurable?

mogoodrich (changes requested):
Generally looks good @usamaidrsk , just a few requests for changes, and a couple questions for @chibongho

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
usamaidrsk: If the `dispositionLocation` is null, what should be displayed? cc @mogoodrich

mogoodrich: Thanks @usamaidrsk !  Let's just display "Pending Transfer" in that case.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx
mogoodrich: Offhand, that seems reasonable @brandones

mogoodrich (approved):

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-transfer.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Movement } from '@carbon/react/icons';
+import styles from '../ward-patient-card.scss';
+import { useTranslation } from 'react-i18next';
+import { type WardPatient } from '../../types';
+
+export interface WardPatientTransferProps {
+  wardPatient: WardPatient;
+}
+
+const WardPatientTransfer: React.FC<WardPatientTransferProps> = ({ wardPatient }) => {
```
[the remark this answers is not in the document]
chibongho: `WardPatientPendingTransfer` sounds fine to me.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx
chibongho: Yeah, I'm ok with making it not configurable until we have a use case that requires it.

chibongho (approved):
LGTM

PR #1223, opened 2024-07-02
(fix) O3-3524: Add wrapping functions to write value in session storage

vasharma05:
## Requirements

## Summary
This PR adds a helper function `updateValueInSessionStorage` to prevent writing `null` or `undefined` in the session storage.

Previously, when calling `sessionStorage.setItem(key, value)`, if the value was `null`, it was saved in the session storage and when fetching the data from the storage, it was returning `'null'`, which passed the check `value !== null`, since `value = 'null'`.

## Screenshots
None

## Related Issue
https://issues.openmrs.org/browse/O3-3524

### DIFF
--- a/packages/esm-service-queues-app/src/helpers/helpers.ts
+++ b/packages/esm-service-queues-app/src/helpers/helpers.ts
@@ -15,14 +15,7 @@ export const getServiceCountByAppointmentType = (
 };
 
 /**
- * This function updates the value in session storage if the value is a valid string.
- *
- * In case the value is null or undefined, the key will be removed from session storage.
- *
  * This function is mainly useful for not writing null/ undefined in the session storage
- *
- * @param key
- * @param value
  */
 export function updateValueInSessionStorage(key: string, value: string) {
   if (value === undefined || value === null) {
@@ -34,8 +27,6 @@ export function updateValueInSessionStorage(key: string, value: string) {
 
 /**
  * This function fetches the value for the passed key from session storage
- * @param key
- * @returns
  */
 export function getValueFromSessionStorage(key: string): string | null {
   return sessionStorage.getItem(key);
--- a/packages/esm-service-queues-app/src/helpers/helpers.ts
+++ b/packages/esm-service-queues-app/src/helpers/helpers.ts
@@ -14,17 +14,36 @@ export const getServiceCountByAppointmentType = (
     .reduce((count, val) => count + val, 0);
 };
 
-const initialQueueLocationNameState = { queueLocationName: sessionStorage.getItem('queueLocationName') };
-const initialQueueLocationUuidState = { queueLocationUuid: sessionStorage.getItem('queueLocationUuid') };
-const initialServiceUuidState = {
-  serviceUuid: sessionStorage.getItem('queueServiceUuid'),
-  serviceDisplay: sessionStorage.getItem('queueServiceDisplay'),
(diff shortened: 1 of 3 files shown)


packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -25,7 +25,7 @@ function getInitialUrl(rep: string, searchCriteria?: QueueEntrySearchCriteria) {
 
   if (searchCriteria) {
     for (let [key, value] of Object.entries(searchCriteria)) {
-      if (value != null) {
+      if (value != null && value !== undefined) {
```
brandones, line 28: undefined is nully

```suggestion
      if (value != null) {
```

packages/esm-service-queues-app/src/helpers/helpers.ts
```diff
@@ -14,17 +14,51 @@ export const getServiceCountByAppointmentType = (
     .reduce((count, val) => count + val, 0);
 };
 
-const initialQueueLocationNameState = { queueLocationName: sessionStorage.getItem('queueLocationName') };
-const initialQueueLocationUuidState = { queueLocationUuid: sessionStorage.getItem('queueLocationUuid') };
+/**
+ * This function updates the value in session storage if the value is a valid string.
```
brandones, line 18:
```suggestion
 * This function updates the value in session storage if the value is not null or undefined.
```

packages/esm-service-queues-app/src/helpers/helpers.ts
```diff
@@ -14,17 +14,51 @@ export const getServiceCountByAppointmentType = (
...
+ *
+ * In case the value is null or undefined, the key will be removed from session storage.
+ *
+ * This function is mainly useful for not writing null/ undefined in the session storage
```
brandones, line 22:
```suggestion
```

Clear enough I think :)

packages/esm-service-queues-app/src/helpers/helpers.ts
```diff
@@ -14,17 +14,51 @@ export const getServiceCountByAppointmentType = (
...
+    sessionStorage.removeItem(key);
+  } else {
+    sessionStorage.setItem(key, value);
+  }
+}
+
+/**
+ * This function fetches the value for the passed key from session storage
+ * @param key
+ * @returns
```
brandones, line 38:
```suggestion
```

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
vasharma05: Ohh okay, I just realized, `undefined == null` is true, `undefined === true` is false. Thanks @brandones !

vasharma05:
> For reference, [here is the pattern](https://github.com/openmrs/openmrs-esm-core/blob/dad1df6fbfa5619c5d8b3b01011f50426db6edb1/packages/framework/esm-config/src/module-config/state.ts#L56-L66) I usually follow for handling local storage.

Actually, we are saving and getting the `string` type values only, and JSON will give an error on that.
But I take your suggestion for future works.
Thanks!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
brandones: And importantly, undefined == false is false, and null == false is false.

PR #1228, opened 2024-07-08
(fix): O3-3567  Fix `undefined, undefined` patient name on the patient card

usamaidrsk:
## Requirements

## Summary

Fix `undefined, undefined` patient name on the patient card by adding the `usePatient` hook and `getPatientName`

## Related Issue

[O3-3567](https://openmrs.atlassian.net/browse/O3-3567)

## Other

[O3-3567]: https://openmrs.atlassian.net/browse/O3-3567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


brandones:
@usamaidrsk Two things:

1. I'm confused about what went wrong. What caused this bug?

2. I think we want to avoid making a patient REST request for every card, if we can avoid it. Does the patient object from the ward endpoint really not have names?

usamaidrsk:
> @usamaidrsk Two things:
> 
> 1. I'm confused about what went wrong. What caused this bug?
> 2. I think we want to avoid making a patient REST request for every card, if we can avoid it. Does the patient object from the ward endpoint really not have names?

Yes @brandones , I agree that we do not need to make patient REST requests for every card. As mentioned [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/199906e93d3d4a09a0d505eb610d3173c9aa39bb/packages/esm-ward-app/src/ward-view/ward-view.component.tsx#L73), some data is not present in the patient object from the ward endpoint and the EMR API.

We might as well go ahead and close this PR. Upon further inspection, I found that the undefined issue was resolved in this [PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/1221) by @chibongho where he added [this fix](https://github.com/openmrs/openmrs-esm-patient-management/blob/199906e93d3d4a09a0d505eb610d3173c9aa39bb/packages/esm-ward-app/src/ward-view/ward-view.component.tsx#L78)

PR #1230, opened 2024-07-15
(feat) O3-3192: Harvest bed administration functionality from ugandaemr into patientmanagement

brandones:
## Requirements

## Summary

Copied the code from https://github.com/METS-Programme/esm-ugandaemr-bed-management-app to create a bed management app here. Code required no modification, other than to fix some imports. The `ErrorState` component now comes from esm-framework rather than esm-patient-common-lib, and the `CardHeader` component has been copied into this repo.

Since the code is all copy-pasted from https://github.com/METS-Programme/esm-ugandaemr-bed-management-app, I would not as a reviewer bother to review most of the files. If we want to improve things we should do that in follow-up PRs. Please do check the boilerplate stuff such as `package.json`.

## Screenshots

Seems to work. Here is bed management running from this PR.

https://github.com/user-attachments/assets/f080e7a7-06f8-4e79-ab4a-dd3e095c523e

## Related Issue

https://openmrs.atlassian.net/browse/O3-3192


vasharma05:
Hi @brandones!
Thanks for the PR!
Can you also verify if the configuration defaults are set to work with dev3?
Thanks!

packages/esm-bed-management-app/src/config-schema.ts
```diff
@@ -0,0 +1,10 @@
+import { Type } from "@openmrs/esm-framework";
+
+export const configSchema = {
+  admissionLocationTagUuid: {
+    _type: Type.UUID,
+    _description:
+      "UUID for the location tag of the `Admission Location`. Patients may only be admitted to inpatient care in a location with this tag",
+    _default: "1c783dca-fd54-4ea8-a0fc-2875374e9cb6",
```
vasharma05, line 8: No location tag `Admission Location` was found on the dev3
@ibacher, do we need to add a new tag in the Distro RefApp?
Thanks!

mseaton (changes requested):
@brandones - see a few comments.  It looks to me like the entire esm has been harvested over - not just the files related to bed administration, but also those related to patient-bed assignments, which should not be included.

packages/esm-bed-management-app/src/config-schema.ts
mseaton: It would be best if we could broaden this to support a `uuid or name` of a location tag, and then configure this by default to be `Admission Location`, which is what is expected throughout the backend modules, including EMR API and Bed Management.  I do not support hard-coding uuids as defaults in esm configs in general.

packages/esm-bed-management-app/src/assets/landing-page.png
ibacher, line 1: I'm also not sure why we need this in the package?

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -0,0 +1,185 @@
...
+  return {
+    data: data?.data?.results ?? [],
+    error,
+    isLoading,
+    isValidating,
+    mutate,
+  };
+};
+
+export const getBedsForLocation = (locationUuid: string) => {
```
ibacher, line 22: I sort of understand having a `use` hook and a `get` function for the same thing (though, not really), but `getBedsForLocation` and `findBedByLocation` are basically duplicative.

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  return {
+    bedData: mappedBedData,
+    isLoading,
+    error,
+  };
+};
+
+export const useLocationName = (locationUuid: string) => {
+  const apiUrl = `/ws/rest/v1/location/${locationUuid}`;
```
ibacher, line 55: This feels like a case where we should've already been able to get this data without making a new request.

```suggestion
  const apiUrl = `/ws/rest/v1/location/${locationUuid}?v=custom:(display)`;
```

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+
+export const findBedByLocation = (locationUuid: string) => {
+  const locationsUrl = `/ws/rest/v1/bed?locationUuid=${locationUuid}`;
+  return openmrsFetch(locationsUrl, {
+    method: 'GET',
+  });
+};
+
+export const useWards = (locationUuid: string) => {
```
ibacher, line 72:
```suggestion
export const useWards = (tag: string) => {
```

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -0,0 +1,185 @@
...
+};
+interface BedType {
+  name: string;
+  displayName: string;
+  description: string;
+}
+interface BedTag {
+  name: string;
+}
+export async function saveBedType({ bedPayload }): Promise<FetchResponse<BedType>> {
```
ibacher, line 153: The parameter accepted here should be typed.

packages/esm-bed-management-app/src/summary/summary.scss
```diff
@@ -0,0 +1,72 @@
+@use '@carbon/colors';
+@use '@carbon/layout';
+
+.cardContainer {
```
ibacher, line 4: Formatting here should be consistent. Did we run prettier on this at all?

ibacher (commented):
Most comments I made are placeholders for things that should be fixed, so I'll surface the two that are relevant:

1. I don't see why we need the PNG file at all
2. At least one of the SCSS files has very inconsistent spacing, which should be fixed.

packages/esm-bed-management-app/src/config-schema.ts
mogoodrich: Agreed, this property should be something like "admission location tag" instead of "admission location uuid" and default to "Admission Location"

brandones: Yeah I would love it if we could reference things by name in the configs, but I'm not aware of the location endpoint will query locations by location tag name. If that is possible, it's not documented in the [REST docs](https://rest.openmrs.org/#location). Though it should be noted that neither is getting locations by location tag UUID documented there. Not sure if there's some other source of documentation or people just read the backend Java code to figure out what the API is.

The other possibility is to support using location tag name in the config by making two requests. First we would get all the location tag data from `/locationtag`, then look up the UUID by the configured name, then use that UUID in the `/location` query.

ibacher: Well, the REST docs are incomplete on that point; we [support both](https://github.com/openmrs/openmrs-module-webservices.rest/blob/57b2e5ed9ce59201a351093ea68e62de408799ff/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/LocationSearchHandler.java#L62).

brandones: Oh I replaced the UUID with a tag name and it just works :laughing:  We should probably document this somewhere

brandones:
@ibacher @mseaton What's the deal with the `/admissionLocation` REST endpoint? Presumably this comes from [module-bedmanagement](https://github.com/openmrs/openmrs-module-bedmanagement)? Is it documented anywhere?

Right now there is a function called `useAdmissionLocations` that makes a request to that endpoint. Semantically I don't know why that should be a different thing than making a request to the `/location` endpoint with the "Admission Location" tag.

packages/esm-bed-management-app/src/summary/summary.resource.ts
brandones: `getBedsForLocation` is cruft. `findBedByLocation` is called many times by the [same component](https://github.com/openmrs/openmrs-esm-patient-management/blob/74a79e963fa561a900d3cc0d6616760e09a6bd33/packages/esm-bed-management-app/src/bed-administration/bed-administration-table.component.tsx#L87-L108) to fetch beds for an array of locations. This can probably be simplified (or all the logic moved into a hook, at least).

ibacher:
@brandones Yeah, it's poorly named class. `AdmissionLocation` essentially provides bed-management data about a location, e.g., the number of beds and the number of beds occupied. See [this class](https://github.com/openmrs/openmrs-module-bedmanagement/blob/master/api/src/main/java/org/openmrs/module/bedmanagement/AdmissionLocation.java). APIs for modules are generally much less documented than the APIs for REST, and APIs for modules that have largely not been a part of the RefApp prior to O3 are usually almost completely undocumented.

packages/esm-bed-management-app/src/summary/summary.resource.ts
ibacher: That's definitely something that should go in the "improve the API" queue.

packages/esm-bed-management-app/src/config-schema.ts
mseaton: TBH, in the future my hope is that we don't need this Admission Location tag at all.  I don't really see why the bedmanagement module should restrict which locations beds can be assigned to to only "Admission Location" locations.  There really is no need for that, and in fact it may be that we want to support assigning patients to beds without actually admitting them, as was indicated in a recent slack conversation...  But that's for a future ticket.

mseaton:
> Yeah, it's poorly named class

+1.  It may make sense to deprecate and rename this more sensibly in the future.

packages/esm-bed-management-app/src/config-schema.ts
mseaton: > If we remove the tag, how will we distinguish locations which are not wards? We still want a way to mark a location as admission location since not every location can admit patients

I'm not saying we should remove the tag from the back end, or from use in appropriate places.  But all we are aiming to harvest here is a UI for creating beds and assigning beds to a location, as well as managing bed types and tags.  Whether or not the locations we are assigning beds to are considered inpatient wards that can admit patients is not necessarily a factor.

mseaton: For now, I think Admission Location is used on the backend as well, so this should stay as-is.  I was just commenting on the fact that this might go away in the future.  A new location tag would be fine if we need to distinguish.  Personally, I see no harm in an admin UI for adding beds to locations broadly.  Then, the locations that users are allowed to assign patients to beds are those locations that have beds allocated to them...

packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
```diff
@@ -93,7 +93,7 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
     resolver: zodResolver(BedAdministrationSchema),
     defaultValues: {
       bedId: initialData.bedNumber || '',
-      description: initialData.description || '',
+      description: initialData.bedType.description || '',
```
mseaton, line 96: Is this really intended to be the bedType.description?  BedFormData does appear to have a description property.  Why was this changed?

packages/esm-bed-management-app/src/bed-admission/bed-layout/bed-layout.component.tsx
mseaton, line 1: Are you sure these bed-layout components should be deleted?  I can't see here what they were doing, but if these are responsible for creating a layout of beds within a location then that is part of bed administration we need to keep.

mseaton (commented):
A bit hard to review.  Looks like lots of good changes.  I made a few comments on a few things to check on.

packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
brandones: The `description` is an attribute of the bed type on the backend. AFAIK beds themselves do not have descriptions. The form data has a description field, which I think is why it is represented at the top level of that object.

packages/esm-bed-management-app/src/bed-admission/bed-layout/bed-layout.component.tsx
brandones: No, these are responsible for displaying the layout of beds when allocating patients to beds. See [<BedLayoutList>](https://github.com/METS-Programme/esm-ugandaemr-bed-management-app/blob/e094746e7ef12d86b885a0dca9b955be2b84265e/src/bed-admission/bed-layout/bed-layout-list.component.tsx#L72) which is only used in [`allocate-bed-workspace.component.tsx`](https://github.com/METS-Programme/esm-ugandaemr-bed-management-app/blob/main/src/workspace/allocate-bed-workspace.component.tsx#L106).

brandones:
Please excuse the force-push, wiping out previous comments. But I've made this much easier to review (hopefully) on a per-commit basis. Now all of the changes I've made are in the commit https://github.com/openmrs/openmrs-esm-patient-management/pull/1230/commits/db036dc7869d7d66e40c40102e002cf79b2bdd03 . The prior commits are just copying the files in and then prettier & lint fix.

mseaton (approved):
Great work @brandones .  LGTM, though I only gave it a rather cursory review this time around.  Would be great to get someone from the Uganda team to look it over and confirm they'd be comfortable using this shared code for bed administration.

brandones:
Tagging contributors to the https://github.com/METS-Programme/esm-ugandaemr-bed-management-app repo: @denniskigen @slubwama @gabriel090

PR #1235, opened 2024-07-17
(fix)O3-3590: Unable to change Provider and appointment duration when  editing appointment

Faithaigo:
## Requirements

## Summary
This PR verifies duplicate appointments for the same service and location only during the creation of an appointment. It removes the verification when an appointment is being edited.

## Screenshots
Before
https://github.com/user-attachments/assets/9586d893-5a5b-4352-b3b8-5394bad753a6

After
https://github.com/user-attachments/assets/db8aae51-be3f-458d-81ac-89a44c362c27

## Related Issue

[](https://openmrs.atlassian.net/browse/O3-3590)

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/form/appointments-form.component.tsx
@@ -275,21 +275,23 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
     const appointmentPayload = constructAppointmentPayload(data);
 
     // check if Duplicate Response Occurs
-    const response: FetchResponse = await checkAppointmentConflict(appointmentPayload);
-    let errorMessage = t('appointmentConflict', 'Appointment conflict');
-    if (response?.data?.hasOwnProperty('SERVICE_UNAVAILABLE')) {
-      errorMessage = t('serviceUnavailable', 'Appointment time is outside of service hours');
-    } else if (response?.data?.hasOwnProperty('PATIENT_DOUBLE_BOOKING')) {
-      errorMessage = t('patientDoubleBooking', 'Patient already booked for an appointment at this time');
-    }
-    if (response.status === 200) {
-      setIsSubmitting(false);
-      showSnackbar({
-        isLowContrast: true,
-        kind: 'error',
-        title: errorMessage,
-      });
-      return;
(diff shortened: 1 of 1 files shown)


packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -275,21 +275,23 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
...
-    }
-    if (response.status === 200) {
-      setIsSubmitting(false);
-      showSnackbar({
-        isLowContrast: true,
-        kind: 'error',
-        title: errorMessage,
-      });
-      return;
+    if (context !== 'editing') {
```
mogoodrich, line 278: The issue here is that now if when editing an appointment you *can* potentially change it to a date/time of another apopintment...  also, I would think we'd still want to run the "service unavailable"?

Ideal, I think, we'd simply not throw the "double booking" error if the uuid of the "double" appointment matches the 

Thanks @Faithaigo  I see two issue here:
1) We still want to check if the service is unavailable at the new time, I think?
2) We still want to make sure we don't change the date to the time when a patient has *another* appointment booked

Ideally, the "checkAppointmentConflict" endpoint would give us the uuid of the conflict appointment so that we could compare if the uuid == uuid of appointment conflict we are trying to edit, but it might not do this?

In that case, I would move the context !== 'editing' check to just determine whether or not to show the "PATIENT_DOUBLE_BOOKING" error, but we should also ticket or note the limitations of our fix.

mogoodrich (commented):
@Faithaigo  see my comments, hopefully they are clear, lmk if you have questions

packages/esm-appointments-app/src/form/appointments-form.component.tsx
Faithaigo: Thank @mogoodrich.  Let me work something out based on the last point

Faithaigo: Hello @mogoodrich, I've pushed an update. Could you please review it and provide your feedback?
Thank you!

mogoodrich (approved):
LGTM, thanks @Faithaigo , I will merge!

PR #1237, opened 2024-07-18
(feat) O3-3125: Ward App - show patient cards for patients admitted but not assigned a bed

mogoodrich:
https://openmrs.atlassian.net/browse/O3-3125

So it was relatively straightforward to get this added, but there’s inconsistencies in how the bed cards vs unassigned patients are displayed (note the two different formats of names in the screenshots) that will exist until we the EMR-API endpoint up-to-snuff. I will make sure that is scheduled for the next sprint

image

## Requirements

### DIFF
--- a/packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
+++ b/packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import WardPatientCard from '../ward-patient-card/ward-patient-card';
-import styles from './unassigned-bed.scss';
+import styles from './unassigned-patient.scss';
 import { type WardPatient } from '../types';
 
 export interface UnassignedPatientProps {
@@ -9,7 +9,7 @@ export interface UnassignedPatientProps {
 
 const UnassignedPatient: React.FC<UnassignedPatientProps> = ({ wardPatient }) => {
   return (
-    <div className={styles.unassignedBed}>
+    <div className={styles.unassignedPatient}>
       <div key={'unassigned-bed-pt-' + wardPatient.patient.uuid}>
         <WardPatientCard patient={wardPatient.patient} visit={wardPatient.visit} />
       </div>
--- a/packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
+++ b/packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import WardPatientCard from '../ward-patient-card/ward-patient-card';
+import styles from './unassigned-bed.scss';
+import { type WardPatient } from '../types';
+
+export interface UnassignedPatientProps {
+  wardPatient: WardPatient;
+}
+
+const UnassignedPatient: React.FC<UnassignedPatientProps> = ({ wardPatient }) => {
+  return (
+    <div className={styles.unassignedBed}>
+      <div key={'unassigned-bed-pt-' + wardPatient.patient.uuid}>
+        <WardPatientCard patient={wardPatient.patient} visit={wardPatient.visit} />
+      </div>
+    </div>
+  );
+};
+
+export default UnassignedPatient;
--- a/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
+++ b/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
@@ -6,7 +6,7 @@ const WardPatientName: WardPatientCardElement = ({ patient }) => {
   // TODO: BED-10
   // TODO: this is a hack to support both the patient name rep returned by the bed endpoint (given and family) and the display used by the emr api; this should be fixed once the emr api endpoint is updated
   const { givenName, familyName, display } = patient?.person?.preferredName || {};
-  const name = display ? `${display}` : `${givenName} ${familyName}`;
+  const name = display ?? `${givenName} ${familyName}`;
   return <div className={styles.wardPatientName}>{name}</div>;
 };
 
--- a/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
+++ b/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
@@ -4,9 +4,9 @@ import styles from '../ward-patient-card.scss';
 
 const WardPatientName: WardPatientCardElement = ({ patient }) => {
   // TODO: BED-10
-  // make server return patient.display and use that instead
-  const { givenName, familyName } = patient?.person?.preferredName;
-  const name = `${givenName} ${familyName}`;
+  // TODO: this is a hack to support both the patient name rep returned by the bed endpoint (given and family) and the display used by the emr api; this should be fixed once the emr api endpoint is updated
+  const { givenName, familyName, display } = patient?.person?.preferredName;
+  const name = display ? `${display}` : `${givenName} ${familyName}`;
   return <div className={styles.wardPatientName}>{name}</div>;
 };
 
--- a/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
+++ b/packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
@@ -5,7 +5,7 @@ import styles from '../ward-patient-card.scss';
 const WardPatientName: WardPatientCardElement = ({ patient }) => {
   // TODO: BED-10
   // TODO: this is a hack to support both the patient name rep returned by the bed endpoint (given and family) and the display used by the emr api; this should be fixed once the emr api endpoint is updated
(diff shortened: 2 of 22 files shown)


packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import WardPatientCard from '../ward-patient-card/ward-patient-card';
+import styles from './unassigned-patient.scss';
+import { type WardPatient } from '../types';
+
+export interface UnassignedPatientProps {
+  wardPatient: WardPatient;
+}
+
+const UnassignedPatient: React.FC<UnassignedPatientProps> = ({ wardPatient }) => {
```
mogoodrich, line 10: FWIW, "UnassignedPatient" seemed a better name for this component than "UnassignedBed".  It seems like it should still be grouped with the "EmptyBed" and "OccupiedBed" components, but though it does make it seem a little weird to be in the "beds" directory.

mogoodrich:
So it was relatively straightforward to get this added, but there’s consistencies in how the bed cards vs unassigned patients are displayed that will exist until we the EMR-API endpoint up-to-snuff.  I will make sure that is scheduled for the next sprint

packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
mseaton: I would think that `UnassignedBed` = `EmptyBed`.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -19,7 +20,6 @@ const WardView = () => {
   const { t } = useTranslation();
   const isBedManagementModuleInstalled = useFeatureFlag('bedmanagement-module');
 
-  //TODO:Display patients with admitted status (based on their observations) that have no beds assigned
   if (!isBedManagementModuleInstalled || isLoadingLocation) {
```
mseaton, line 23: Won't this still be an issue @mogoodrich ?  Can the code make it past this point?

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -70,6 +70,7 @@ const WardViewByLocation = () => {
         const admittedPatient = admittedPatientsByUuid.get(patient.uuid);
 
         if (admittedPatient) {
+          admittedPatientsByUuid.delete(patient.uuid); // remove the patient from the admittedPatientsByUuid map so at the end we have a list of patients that are admitted but have no beds assigned
```
mseaton, line 73: I don't love this.  I'd rather we either have a variable with a different name, or just maintain a Map of patients with beds assigned and then check if the patient is in one but not the other.  I don't love having a Map named `admittedPatientsByUuid` that doesn't actually contain admitted patients by uuid at all times.

mseaton (commented):
Mostly looks good I think, though a few questions / comments

mseaton:
@mogoodrich - can you update the PR description and add more of a description / screenshots?

packages/esm-ward-app/src/beds/unassigned-patient.component.tsx
mogoodrich: Right, agreed, I was just saying that the ticket says to call this element "UnassignedBed", which seemed wrong to me because as you said UnassignedBed implies EmptyBed to me, so I went with UnassignedPatient.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
mogoodrich: yeah, i wasn't crazy about it either, but wasn't sure if I was nit-picking, I can probably do some sort of filter.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
mogoodrich: Oh, good point, will fix... it gives me an answer to another question I had as well, didn't see we had a variable to determine if the module was installed or not.

mogoodrich:
Moved my description and screenshot from the comments to actual PR decsription.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -92,16 +92,27 @@ const WardViewByLocation = () => {
...
+        return (
+          <UnassignedPatient wardPatient={{ ...admittedPatient, admitted: true }} key={admittedPatient.patient.uuid} />
+        );
+      });
+
     return (
       <>
         {wardBeds}
-        {bedLayouts.length == 0 && (
+        {bedLayouts.length == 0 && isBedManagementModuleInstalled && (
```
mseaton, line 108: Not sure we really want or need this warning, but we can address in a future ticket if necessary.

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -92,16 +92,27 @@ const WardViewByLocation = () => {
       return <WardBed key={bed.uuid} bed={bed} wardPatients={wardPatients} />;
     });
 
+    const patientsInBedsUuids = bedLayouts.flatMap((bedLayout) => bedLayout.patients.map((patient) => patient.uuid));
```
mseaton, line 95: I assume bedLayouts is always non-null?

mseaton (approved):
LGTM

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
mogoodrich: I tested on a location with no beds, and this worked, but it's a good, bigger point that now that I've removed the check to render nothing if bed management isn't installed we probably need a lot more checks around this, let me take a look.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
```diff
@@ -4,9 +4,9 @@ import styles from '../ward-patient-card.scss';
 
 const WardPatientName: WardPatientCardElement = ({ patient }) => {
   // TODO: BED-10
-  // make server return patient.display and use that instead
-  const { givenName, familyName } = patient?.person?.preferredName;
-  const name = `${givenName} ${familyName}`;
+  // TODO: this is a hack to support both the patient name rep returned by the bed endpoint (given and family) and the display used by the emr api; this should be fixed once the emr api endpoint is updated
+  const { givenName, familyName, display } = patient?.person?.preferredName;
```
brandones, line 8: If `patient` or `patient.person` or `patient.person.preferredName` is nullish, this will produce a TypeError due to attempting to destructure null. It's a shame that TS doesn't catch that. Either get rid of the optional chaining or add a default object `|| {}` to the end of the line.

brandones (changes requested):

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
mogoodrich: makes sense, thanks @brandones !

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
mogoodrich: Yeah, agreed... this was an existing warning I preserved.

mogoodrich (commented):
Alright, sorry, another update... this breaks up the ward view with bed management from the one without bed management to prevent an error when loaded on a instance that does not have the bed management module installed.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
```diff
@@ -4,9 +4,9 @@ import styles from '../ward-patient-card.scss';
 
 const WardPatientName: WardPatientCardElement = ({ patient }) => {
   // TODO: BED-10
-  // make server return patient.display and use that instead
-  const { givenName, familyName } = patient?.person?.preferredName;
-  const name = `${givenName} ${familyName}`;
+  // TODO: this is a hack to support both the patient name rep returned by the bed endpoint (given and family) and the display used by the emr api; this should be fixed once the emr api endpoint is updated
+  const { givenName, familyName, display } = patient?.person?.preferredName || {};
```
mogoodrich, line 8: This should be fixed, thanks @brandones !

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-name.tsx
```diff
@@ -4,9 +4,9 @@ import styles from '../ward-patient-card.scss';
...
+  const name = display ? `${display}` : `${givenName} ${familyName}`;
```
brandones, line 9:
```suggestion
  const name = display ?? `${givenName} ${familyName}`;
```

brandones (approved):

PR #1250, opened 2024-07-25
O3-3648: Ward App - Update Ward View after updating the Inpatient Visit API

mogoodrich:
See : https://openmrs.atlassian.net/browse/O3-3648

This updates the Ward app to use the new EMR API endpoint for inpatient admissions added here:

https://openmrs.atlassian.net/browse/EA-190

This allows us to use this endpoint for all admitted ward patients.  (We still need to handle the edge case of patients that somehow are in a bed but aren't admitted)


__mocks__/patient.mock.ts
```diff
@@ -20,7 +19,14 @@ export const mockPatientAlice: Patient = {
...
     causeOfDeath: null,
-    preferredName: null,
+    preferredName: {
+      uuid: '00000000-0000-0001-0000-000000000000',
+      display: 'Alice Johnson',
+      givenName: 'Alice',
+      familyName: 'Johnson',
+      familyName2: null,
+      voided: false,
+    },
```
mogoodrich, line 29: Not quite sure if there was a reason we weren't adding the preferred name directly to the mock?

packages/esm-ward-app/src/beds/occupied-bed.test.tsx
```diff
@@ -5,33 +5,26 @@ import { bedLayoutToBed, filterBeds } from '../ward-view/ward-view.resource';
...
 jest.mocked(useConfig).mockReturnValue(defaultConfig);
 
 const mockBedLayouts = filterBeds(mockAdmissionLocation);
-
 const mockBedToUse = mockBedLayouts[0];
-jest.replaceProperty(mockBedToUse.patients[0].person, 'preferredName', {
-  uuid: '',
-  givenName: 'Alice',
-  familyName: 'Johnson',
-});
```
mogoodrich, line 21: See above, I just added the preferred name directly to the mock, and didn't seem to cause issues.

packages/esm-ward-app/src/hooks/useInpatientAdmission.ts
```diff
@@ -0,0 +1,22 @@
+import useSWR from 'swr';
+import { type FetchResponse, openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
+import { type InpatientAdmissionFetchResponse, InpatientRequestFetchResponse } from '../types';
+
+export function useInpatientAdmission(locationUuid: string) {
+  const customRepresentation =
```
mogoodrich, line 6: We can tweak the REP we request as needed

packages/esm-ward-app/src/types/index.ts
```diff
@@ -16,6 +16,11 @@ export interface WardPatientCardProps {
 
 export type WardPatientCardRow = React.FC<WardPatientCardProps>;
 export type WardPatientCardElement = React.FC<WardPatientCardProps>;
+export type WardPatient = {
+  patient: Patient;
+  visit?: Visit;
+  admitted: boolean;
+};
```
mogoodrich, line 25: so I tried to simplify this to singel WardPatient type, instead of having an AdmittedPatient and a InpatientRequest, hopefully this make sense to others.

Note that as we add additional functionality (like the time on ward functionality) we will need to add additional (optional) properties to WardPatient like "latestAdmissionOrTransferEncounter" (which will be a property returned by the inpatient admission endpoint, but not the bed management admissionLocation endpoint).

packages/esm-ward-app/src/types/index.ts
```diff
@@ -96,34 +98,14 @@ export interface InpatientRequest {
...
 }
 
-export type DispositionTypeOld = 'ADMISSION' | 'TRANSFER' | 'DISCHARGE';
-
-// AdmittedPatient[] returned by:
-// GET /rest/emrapi/inpatient/visits
-export interface AdmittedPatient {
+export interface InpatientAdmission {
   patient: Patient;
   visit: Visit;
```
mogoodrich, line 110: Will need to add properties like "latestAdmissionOrTransferEncounter" once we modify the REP to return them.

packages/esm-ward-app/src/types/index.ts
mseaton: I would have expected this to have an `inpatientAdmission?: InpatientAdmission` property (and also something like a `bedAssignment?: BedAssignment` property)

packages/esm-ward-app/src/types/index.ts
mseaton: I'm confused.  You are setting a custom rep already in this PR - why aren't you including that property?

__mocks__/wards.mock.ts
```diff
@@ -66,271 +66,9 @@ export const mockAdmissionLocation: AdmissionLocation = {
...
-          uri: 'http://localhost:8080/openmrs/ws/rest/v1/location/4d7e927d-6850-11ee-ab8d-0242ac120002?v=full',
-          resourceAlias: 'location',
-        },
-      ],
-      resourceVersion: '2.0',
-    },
-    timeSinceAdmissionInMinutes: 15,
-    timeAtInpatientLocationInMinutes: 15,
+    patient: mockPatientBrian,
+    visit: null,
```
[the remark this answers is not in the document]
mogoodrich: This will be null in the (admittedly edge) case when a patient is in a bed but somehow has not been admitted.

packages/esm-ward-app/src/hooks/useInpatientAdmission.ts
```diff
@@ -0,0 +1,22 @@
...
+    'custom:(visit,' +
+    'patient:(uuid,identifiers,voided,' +
+    'person:(uuid,display,gender,age,birthdate,birthtime,preferredName,preferredAddress,dead,deathDate,voided)))';
```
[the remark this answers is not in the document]
mogoodrich: Fair enough, I just copied the rep used for one of the other endpoints, but no need to include voided, you are correct, will update.

packages/esm-ward-app/src/hooks/useInpatientAdmission.ts
```diff
@@ -0,0 +1,22 @@
...
+  const apiUrl = `/ws/rest/emrapi/inpatient/admission?currentInpatientLocation=${locationUuid}`;
+  const { data, ...rest } = useSWR<FetchResponse<InpatientAdmissionFetchResponse>, Error>(
+    locationUuid
+      ? `${restBaseUrl}/emrapi/inpatient/admission?currentInpatientLocation=${locationUuid}&v=${customRepresentation}`
+      : null,
+    openmrsFetch,
+  );
```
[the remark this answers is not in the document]
mogoodrich: Makes sense, will ticket.

packages/esm-ward-app/src/types/index.ts
mogoodrich: This is following the existing pattern that had been started.   We may not always have an InpatientAdmission (edge case when a patient is in a bed but not admitted).  So we hvae a "Patient" property that we make sure we populate (either from patient/admission or the bed management endpoint) in all cases.

packages/esm-ward-app/src/types/index.ts
mogoodrich: I was just starting with the basics and noted that the new property will be needed in the follow-on ticket I created.

packages/esm-ward-app/src/types/index.ts
mseaton: I understand the goal of having a common patient property you can rely upon, but that doesn't mean we still shouldn't have an inpatientAdmission property that is available.  That's why I indicated above it should be nullable.  Why wouldn't we include this??

packages/esm-ward-app/src/types/index.ts
mseaton: Seems odd though to make a comment elsewhere in the PR that something needs to be added to the rep, when you are defining the rep.  Why not just include it in the first place and not have to make that comment?

__mocks__/wards.mock.ts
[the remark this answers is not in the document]
mseaton: Why?  I would expect everyone to have an active visit, whether it is associated with an admission or not.

brandones:
Looks like there's a lot of merge conflicts, mostly from https://github.com/openmrs/openmrs-esm-patient-management/pull/1253 .

mogoodrich:
Ah thanks @brandones .... looks like that commit when in this morning, I will fix.

mogoodrich (commented):
Added some comments, there are merge conflicts from another comment coming through this morning, I will resolve now and then likely merge this in before there are more (and this is blocking picking up other tickets)

mogoodrich:
@brandones merge conflict should be resolved

brandones (approved):
LGTM, thanks Mark. I imagine the types will continue to evolve.

PR #1255, opened 2024-07-26
(fix) O3-3661: Admission Requests List should include both Admission and Transfer requests

mogoodrich:
https://openmrs.atlassian.net/browse/O3-3661

The terminology is a little confusing, but according to @Fiona Anderson  the admissions request list should should those awaiting admission to the ward, and those awaiting transfer into the ward from another ward.

There isn't a currently a way to set up transfer questions on our demo server (that will come with https://openmrs.atlassian.net/issues/O3-3247?filter=10648, so this won't be fully testable until then) but this PR expands the admission request view to dsplay both admission and transfer requests for the current ward in the following view.


packages/esm-ward-app/src/hooks/useInpatientRequest.ts
```diff
@@ -6,7 +6,10 @@ import useWardLocation from './useWardLocation';
 const defaultRep =
   'custom:(dispositionLocation,dispositionType,disposition,dispositionEncounter:full,patient:default,dispositionObsGroup,visit)';
 
-export function useInpatientRequest(dispositionType: Array<DispositionType> = ['ADMIT'], rep: string = defaultRep) {
+export function useInpatientRequest(
+  dispositionType: Array<DispositionType> = ['ADMIT', 'TRANSFER'],
```
mogoodrich, line 10: technically I didn't need to change the default value, but if we are going to have a default value seemed to make sense to have it equal to the only value we are currently passing in.

packages/esm-ward-app/src/ward-workspace/admission-requests.workspace.tsx
```diff
@@ -3,15 +3,13 @@ import styles from './admission-requests-workspace.scss';
...
 import { useInpatientRequest } from '../hooks/useInpatientRequest';
-import useWardLocation from '../hooks/useWardLocation';
 import { Loading } from '@carbon/react';
 import { InlineNotification } from '@carbon/react';
 import { useTranslation } from 'react-i18next';
 
 const AdmissionRequestsWorkspace: React.FC<DefaultWorkspaceProps> = () => {
   const { t } = useTranslation();
-  const { inpatientRequests, isLoading, error } = useInpatientRequest();
-  const admissionRequests = inpatientRequests?.filter((request) => request.dispositionType == 'ADMIT');
```
mogoodrich, line 14: We shouldn't have to filter here if we are only fetching types we want?

brandones:
Hi @mogoodrich , please don't delete the PR template. Could you please add a screenshot or video? It's hard to evaluate the work without that.

mogoodrich:
Sorry about deleting the PR template @brandones .

RE: screenshot, I realized after I made the changes that this won't be fully testable (and therefore demoable) until we implement https://openmrs.atlassian.net/issues/O3-3247?filter=10648 , but I just added a screenshot to the PR showing what this PR should change.

packages/esm-ward-app/src/ward-workspace/admission-requests.workspace.tsx
```diff
@@ -29,7 +27,7 @@ const AdmissionRequestsWorkspace: React.FC<DefaultWorkspaceProps> = () => {
 
   return (
     <div className={styles.admissionRequestsWorkspace}>
-      {admissionRequests.map((admissionRequest, indx) => (
+      {inpatientRequests.map((admissionRequest, indx) => (
         <AdmissionRequestCard key={indx} patient={admissionRequest.patient} />
```
denniskigen, line 31: Is there something in the response we could use as a key here?

brandones: Or just like `inpatient-request-workspace-card-${indx}`.

@vasharma05 sorry we didn't catch this in the original PR with this change, but you can't just use an integer as a React key; it is very easy to wind up with duplicate keys that way. Duplicate keys can cause really bizarre and confusing bugs.

brandones:
```suggestion
        <AdmissionRequestCard key={`inpatient-request-workspace-card-${indx}`} patient={admissionRequest.patient} />
```

chibongho (approved):
LGTM

PR #1257, opened 2024-07-28
(Fix) UI Issue with Jumbled Search Results when PatientSearchBar Component is used in a workspace

amosmachora:
## Requirements

## Summary
In this PR I have fixed a small UI issue that showed jumbled search results whever the `PatientSearchBar` component is used in small windows for lack of a better term i.e workspaces.

The issue was that there is an element that hard coded this style 
```
[data-extension-id=patient-search-bar]>div button {
    height: 3rem;
}
```
so the above css rule forced the button to a height of `3rem` thereby throwing the UI into a complete catastrophe. Fixed it by adding a new rule with `height: auto` that overrides the original rule. Please check the attached screenshots.

### Before
Screenshot 2024-07-27 190914

### After
Screenshot 2024-07-27 190938


vasharma05 (commented):
Hi @amosmachora!
Please follow the [PR conventions](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#pull-requests) and mention the ticket in the PR title to move ahead.
Thanks!

vasharma05 (changes requested):
Hi @amosmachora!
Please follow the [PR conventions](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#pull-requests) and mention the ticket in the PR title to move ahead.
Thanks!

denniskigen:
Isn't [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/patient-search/patient-search.scss#L9) you ought to fix? Because it's not scoped under a class name, it applies to all instances of the `patient-search-bar` extension. That's not ideal. But yeah, the fix should probably happen there and not anywhere else.

amosmachora:
You are right I should avoid overriding the original rule. I fear I might break other places where the component is used but let me test to see if we can achieve the same functionality.

amosmachora:
Here is a screenshot.

Scoping [this ](https://github.com/openmrs/openmrs-esm-patient-management/blob/e0a71ec0c4dac6b73b08a8aeaba582e1630cd9d2/packages/esm-appointments-app/src/patient-search/patient-search.scss#L9) class to only apply to the `searchBarWrapper` makes wherever else the component is used the height becomes `auto` by default.

image

denniskigen (approved):

vasharma05 (approved):
Thanks @amosmachora !

PR #1261, opened 2024-07-31
(feat) O3-3632 - Add config for maxDate and minDate to DateObsField 

kajambiya:
## Requirements

## Summary
This PR allows users to pass max and min dates to the `DateObsField` via config

## Screenshots

https://github.com/user-attachments/assets/f4ae01a9-2e67-4910-91bd-ef175e9d910d


packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -159,15 +159,38 @@ interface DateObsFieldProps {
...
+  if (dateString === 'today') {
+    return new Date();
+  }
+
+  const date = new Date(dateString);
+  if (!isNaN(date.getTime())) {
+    return date;
+  }
+
+  return 'Invalid min or max date!';
```
ibacher, line 180: 1. This is a user-facing string so it must be localizable
2. Make the message friendlier by passing in as an argument whether this is the min date or max date
3. Instead of checking the return type, throw an error and use try...catch. It may be better to actually create the error message at that point.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -12,7 +12,8 @@ export interface FieldDefinition {
   label?: string;
   uuid: string;
   placeholder?: string;
```
ibacher, line 14: May as well remove placeholder here too? Also, these should be removed from the _actual config schema_ and not just the types.

ibacher (changes requested):
Thanks @kajambiya! A few things need to be cleaned up.

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -159,15 +159,38 @@ interface DateObsFieldProps {
...
+  }
+
+  if (dateString === 'today') {
+    return new Date();
+  }
+
+  const date = new Date(dateString);
+  if (!isNaN(date.getTime())) {
+    return date;
+  }
```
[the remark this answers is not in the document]
kajambiya: This was to check whether the string was successfully converted into a date but since you advised a try...catch, this is no longer needed.

brandones:
Ping @kajambiya . Is it clear to you how to move forward with this?

kajambiya:
> Ping @kajambiya . Is it clear to you how to move forward with this?

Yes it is @brandones . Was just busy with some other tasks and had packed this but going to wire a commit tomorrow.

kajambiya:
> Thanks @kajambiya! A few things need to be cleaned up.

Hi @ibacher 
I've addressed the comments.
Just a quick one, I'm logging the thrown error but I'm also going ahead to pass it as a field error. Is this necessary or the log should be enough?

ibacher:
@kajambiya As a general rule, the most important thing is that if there is an error and it blocks the user from taking an action, that error must be shown to the user in a way they can understand and (hopefully) correct. So the general answer to your question is that logging to the console is a nice-to-have, but not a sufficient implementation of handling an error.

brandones:
@ibacher Does that mean this is in good shape? It looks like the `configDateError` has been added to the display using the `invalidText` property of the datepicker.

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -159,14 +161,43 @@ interface DateObsFieldProps {
...
+    return null;
+  }
+  if (dateString === 'today') {
+    return dayjs(new Date()).toDate();
+  }
+
+  const parsedDate = dayjs(dateString);
+  if (!parsedDate.isValid()) {
+    throw new Error(
+      translateFrom(moduleName, 'invalidObsDateErrorMessage', `The ${dateType} date value provided is invalid!`),
```
ibacher, line 179: 1. Why are we using `translateFrom` here instead of `t`?
2. We can't interpolate `Min` and `Max` in here like this. It's not translatable. We can do something like `t('invalidObsDateMinValue', 'The minimum date value provided is invalid.')`

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -159,14 +161,43 @@ interface DateObsFieldProps {
...
   label: string;
   required?: boolean;
-  dateFormat?: string;
-  placeholder?: string;
+  minDate?: string;
+  maxDate?: string;
 }
 
-function DateObsField({ concept, label, required, placeholder }: DateObsFieldProps) {
+const evaluateConfigDate = (dateString: string, dateType: string): Date | null => {
```
ibacher, line 168: So basically, the real feature being supported by this PR is the ability to specify `today` as either the lowest selectable date or the highest selectable date?

ibacher: Basically, the question here is "why don't we just build that instead of this?"

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
```diff
@@ -159,14 +161,43 @@ interface DateObsFieldProps {
...
+  if (!dateString) {
+    return null;
+  }
+  if (dateString === 'today') {
+    return dayjs(new Date()).toDate();
```
ibacher, line 173:
```suggestion
    return new Date();
```

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
ibacher: It would, of course, be better to include some example of what is valid.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -11,8 +11,8 @@ export interface FieldDefinition {
   type: string;
   label?: string;
   uuid: string;
-  placeholder?: string;
-  dateFormat?: string;
+  maxDate?: string;
+  minDate?: string;
```
brandones, line 15: You modified the interface but did not change the actual config schema. Also why did you get rid of `placeholder`?

brandones (changes requested):

packages/esm-patient-registration-app/src/patient-registration/field/obs/obs-field.component.tsx
brandones: +1, I would support a config interface like, for example,

```
datepicker: {
  allowPastDates: boolean
  allowFutureDates: boolean
}
```

does what's actually needed, is easier to maintain and code, and easier to understand and use.

kajambiya: @ibacher & @brandones, my question here would be, don't we anticipate a scenario where someone would pass a fixed date e.g 2024-10-01 as either the `min` or `max` date. I don't have a use case per se but I just thought it would be possible.

brandones: Yeah I just can't really imagine a use case for that. It is good not to build things no one needs, or to design and build things with no use case. It is valuable to keep our code easy to understand, maintain, and use. If someday a use case comes up, we'll implement it then.

mseaton: @ibacher / @brandones / @kajambiya please keep in mind that "today" from the perspective of a field on a form often needs to be "as of the encounter date" _not_ "as of the entry date".  So it is not sufficient to say we are only going to support past and future as of a "new Date()" object.  So this might be an example of where a parent component (eg. a parent form with an encounterDate variable would want to set that as the minDate / maxDate / defaultDate on an obsField it contains).  Sorry if this is not relevant, but wanted to mention it in case it is.

ibacher: @mseaton Yeah, retrospective registration probably requires a redesign of the app. Here, though, we're just passing in a value from the configuration, so the issue is kind of orthogonal to that concern (i.e., we could implement this so that it interprets "today" relative to the user's current selected time for the registration to happen). The issue here is that there's isn't a lot of use for handling strings like '2020-01-05'.

brandones:
@kajambiya is it clear how to move forward?

kajambiya:
> @kajambiya is it clear how to move forward?

@brandones Here's how I think we should proceed. Let me know if you have any thoughts:

Remove the `minDate` and `maxDate`.
Introduce two flags: `allowFutureDates` and `allowPastDates`.
By default, both flags will be set to true, allowing users to select any date. Implementers can then configure whether to allow or restrict future or past dates as needed.

brandones:
Sounds good, thanks @kajambiya

kajambiya:
> Sounds good, thanks @kajambiya

@brandones @ibacher, I've updated the code and the video accordingly. Kindly review and revert.

brandones:
Sorry, I thought we were going to do `allowFutureDates` and `allowPastDates`? The reason I think this is more intuitive is because the dates are not being "enabled" or "disabled," their selection is being "allowed" or "disallowed." But "disallow" is uncommon in interfaces, so setting "allowFutureDates" to false is, I think, the intuitive interface for this.

kajambiya:
> Sorry, I thought we were going to do `allowFutureDates` and `allowPastDates`? The reason I think this is more intuitive is because the dates are not being "enabled" or "disabled," their selection is being "allowed" or "disallowed." But "disallow" is uncommon in interfaces, so setting "allowFutureDates" to false is, I think, the intuitive interface for this.

Hi @brandones Sorry for the late response. I've updated the code and the video accordingly

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -12,7 +12,8 @@ export interface FieldDefinition {
   label?: string;
   uuid: string;
   placeholder?: string;
-  dateFormat?: string;
+  allowFutureDates?: boolean;
+  allowPastDates?: boolean;
```
brandones, line 16: You have modified the type but you have not modified the actual config schema.

brandones (changes requested):

packages/esm-patient-registration-app/src/config-schema.ts
brandones: You must modify the actual config schema; there should be a console warning about that visible in the browser console logs. Please use the browser console.

brandones:
Ping @kajambiya

packages/esm-patient-registration-app/src/config-schema.ts
kajambiya: Done

brandones (approved):

PR #1265, opened 2024-08-01
(feat) O3-3213, O3-3214 - ward app - add elements to display time sin…

chibongho:
…ce admission and time on ward

## Requirements

## Summary

2 new elements for displaying time since admission and time on ward. 

This will probably need another commit to resolve merge conflicts from [this pending PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/1263). 

## Screenshots

image

### DIFF
--- a/packages/esm-ward-app/src/types/index.ts
+++ b/packages/esm-ward-app/src/types/index.ts
@@ -9,20 +9,17 @@ import type {
 } from '@openmrs/esm-framework';
 import type React from 'react';
 
-export interface WardPatientCardProps {
-  patient: Patient;
-  visit: Visit;
-  bed?: Bed;
-}
-
-export type WardPatientCardRow = React.FC<WardPatientCardProps>;
-export type WardPatientCardElement = React.FC<WardPatientCardProps>;
+export type WardPatientCardRow = React.FC<WardPatient>;
+export type WardPatientCardElement = React.FC<WardPatient>;
 
 // WardPatient is a patient admitted to a ward, and/or in a bed on a ward
 export type WardPatient = {
   patient: Patient;
-  visit?: Visit;
+  visit: Visit;
+  bed?: Bed;
   admitted: boolean;
+  encounterAssigningToCurrentInpatientLocation: Encounter;
+  firstAdmissionOrTransferEncounter: Encounter;
 };
 
 export const patientCardElementTypes = [
@@ -34,6 +31,8 @@ export const patientCardElementTypes = [
   'patient-coded-obs-tags',
   'admission-time',
   'patient-identifier',
+  'time-on-ward',
+  'time-since-admission',
 ] as const;
 export type PatientCardElementType = (typeof patientCardElementTypes)[number];
 
@@ -113,6 +112,18 @@ export interface InpatientAdmissionFetchResponse {
 export interface InpatientAdmission {
   patient: Patient;
   visit: Visit;
+
+  // the encounter of type "Admission" or "Transfer" that is responsible
+  // for assigning the patient to the current inpatient location. For example,
+  // if the patient has been admitted /transferred to multiple locations as follows:
+  // A -> B -> A
+  // then encounterAssigningToCurrentInpatientLocation
+  // would be the transfer encounter that lands the patient back to A.
+  encounterAssigningToCurrentInpatientLocation: Encounter;
+
+  // the first encounter of the visit that is of encounterType "Admission" or "Transfer",
+  // regardless of the admission location
+  firstAdmissionOrTransferEncounter: Encounter;
 }
 
 // TODO: Move these types to esm-core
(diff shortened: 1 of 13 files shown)


vasharma05:
Hi @chibongho!
I'll suggest you to please keep only 1 ticket in the PR title and list all tickets in the description.
Alongside, please link the tickets together in JIRA for better understanding.
Thanks!
CC: @ibacher @denniskigen

packages/esm-ward-app/src/types/index.ts
```diff
@@ -9,20 +9,17 @@ import type {
...
+export type WardPatientCardRow = React.FC<WardPatient>;
+export type WardPatientCardElement = React.FC<WardPatient>;
 
 // WardPatient is a patient admitted to a ward, and/or in a bed on a ward
 export type WardPatient = {
   patient: Patient;
-  visit?: Visit;
+  visit: Visit;
+  bed?: Bed;
   admitted: boolean;
```
mogoodrich, line 20: Tangential, but we could potentially remove admitted now and calculate that based on encounterAssigningToCurrentInpatientLocation being null or not.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.tsx
```diff
@@ -0,0 +1,16 @@
+import { age } from '@openmrs/esm-framework';
+import React from 'react';
+import { type WardPatientCardElement } from '../../types';
+import { useTranslation } from 'react-i18next';
+
+const WardPatientTimeOnWard: WardPatientCardElement = ({ encounterAssigningToCurrentInpatientLocation }) => {
+  const { t } = useTranslation();
+  if (encounterAssigningToCurrentInpatientLocation) {
+    const timeOnWard = age(encounterAssigningToCurrentInpatientLocation.encounterDatetime);
```
mogoodrich, line 9: oh, cool, does the age function handle what unit to display (minutes, hours, days, years?) based on the age?

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
```diff
@@ -61,10 +61,16 @@ const WardViewWithBedManagement = () => {
...
+          return { ...inpatientAdmission, admitted: true };
         } else {
           // for some reason this patient is in a bed but not in the list of admitted patients, so we need to use the patient data from the bed endpoint
-          return { patient: patient, visit: null, admitted: false };
+          return {
+            patient: patient,
+            visit: null,
+            admitted: false,
+            encounterAssigningToCurrentInpatientLocation: null,
+            firstAdmissionOrTransferEncounter: null, // TODO: populate this after O3-3704
```
mogoodrich, line 72: FWIW, we still won't be able to populate this after O3-3704... this is the case when for some reason the patient hasn't been admitted and only the bed management endpoint is returning the patient.

mogoodrich (approved):
LGTM, added a few comments, but all tangential, thanks @chibongho

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.tsx
mseaton: Yeah, when you and @ibacher were going back and forth about naming this function, I was going to weigh in that we might not want to use age-related terms, since the idea is to use it for things beyond age, not sure if we need to re-litigate that, but happy to see this used here.

chibongho: Yes! [Recent change](https://github.com/openmrs/openmrs-esm-core/pull/1100).

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
mseaton: firstAdmissionOrTransferEncounter will be null if the patient is not admitted.  There will be no TODO in this case, and the "timeSinceAdmission" should presumably be null and hidden (since they aren't admitted).

However, the EncounterAssigningToCurrentInpatientLocation could (and probably should) be taken from the encounter associated with the patient bed assignment that first put them on the ward, and then this could still drive the "Time on Ward" element accurately.

mseaton (approved):
Looks good to me.  Made a few comments, but they should likely result in new tickets, not issues with this particular feature.  If you can create those as needed, that would be great.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.tsx
ibacher: Well, the specific cut-off points may or may not make sense outside of displaying a patient age, which is why I thought it made sense to keep the age-specific function about age. For instance, age differences less than an hour are displayed in minutes, but over an hour in hours, whereas here it may be more appropriate to have, e.g., "6 hours, 20 minutes".

Basically, as an EHR where patient ages are important, it makes sense to have a first-class function for formatting ages with clinically reasonable defaults. We likely should also have a first-class function for formatting durations, just likely with different cut-off points (and maybe more options).

packages/esm-ward-app/src/ward-view/ward-view.component.tsx
chibongho: filed https://openmrs.atlassian.net/browse/BED-13

PR #1274, opened 2024-08-09
(feat) O3-3242 - ward app - add admission request detail to admission…

chibongho:
… request card

## Requirements

## Summary

This PR adds a row to Admission Request Cards to display the admission note. The row is added as an extension, while the actual element inside the row is the `WardPatientObs`,  which is already implemented (for displaying the Gravita value).

The default uuid for that row, "161011AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA", is a valid concept in both refapp and PIH EMR. However, it seems to be a generic "Free text comment" concept. I'm not sure if I need to go further to limit this concept to the admission note that's submitted as part of the triage form. If there are multiple "Free text comment" obs collected, all of them would be displayed.

## Screenshots

image


packages/esm-ward-app/src/config-schema-admission-request-note.ts
```diff
@@ -0,0 +1,9 @@
+import { Type } from '@openmrs/esm-framework';
+
+export const admissionRequestNoteRowConfigSchema = {
+  conceptUuid: {
+    _type: Type.UUID,
+    _description: 'Required. Identifies the concept for the admission request note.',
+    _default: '161011AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
```
mogoodrich, line 7: No sure what our best practices are for defaults, but I'm thinking we should only include a default if it's something in CIEL and included in the ref app.  (which may be true for the above concept).

mogoodrich (approved):
Generally LGTM, thanks @chibongho , see my one comment.

packages/esm-ward-app/src/config-schema-admission-request-note.ts
chibongho: Yeah, the bar for putting a default UUID seems to be whether it's included in the ref app. 

In this case, the UUID is both in CIEL and in the ref app.

PR #1275, opened 2024-08-11
(feat) O3-3782: Use styleguide page header component

Twiineenock:
## Requirements

## Summary

A reuseable dashboard header component was implemented in the styleguide [here](https://github.com/openmrs/openmrs-esm-core/pull/1084). Therefore this PR adapts all the `openmrs-esm-patient-management` apps to use this.

### Appointments
appointments-header

### Patient Lists
patient-lists-pageHeader

### Service queues
service-queues-pageHeader

## Related Issue

[https://openmrs.atlassian.net/browse/O3-3782](https://openmrs.atlassian.net/browse/O3-3782)


chibongho (approved):
Thanks and nice work!

denniskigen:
@mogoodrich @chibongho note that this removes two elements from the page header - the location span and the datepicker shown in the top right. Is that acceptable?

mogoodrich:
> @mogoodrich @chibongho note that this removes two elements from the page header - the location span and the datepicker shown in the top right. Is that acceptable?

Where are these @denniskigen ?  In the very top header?  (Maybe we were already hiding them in our app and I forgot they existed)

chibongho:
Internally we talked about those 2 elements not needing to be there for every app. So I think this standardized design change is a step in the right direction.

However... In the Appointments app, we definitely need a way to pick the calendar date. Is there another way to do that with this change? 

cc @ojwanganto

brandones:
@chibongho Yes, you can put anything on the right side of the header as the children of the header component. If they are in the designs, @Twiineenock , please fix the implementation so they are not removed. Otherwise this would be a regression.

Twiineenock:
> @chibongho Yes, you can put anything on the right side of the header as the children of the header component. If they are in the designs, @Twiineenock , please fix the implementation so they are not removed. Otherwise this would be a regression.

@brandones , did you mean we retain everything on the right in the patient management as below?

retain

chibongho:
> > @chibongho Yes, you can put anything on the right side of the header as the children of the header component. If they are in the designs, @Twiineenock , please fix the implementation so they are not removed. Otherwise this would be a regression.
> 
> @brandones , did you mean we retain everything on the right in the patient management as below?
> 
> retain

Just the calendar picker. I believe that is the only UI component in the appointments app to be able to change which date we want to view the appointments.

Twiineenock:
Hi @brandones , @chibongho ,

I've committed the changes to ensure that the Datepicker remains on the right side of the page-header in the UI.

[datePicker.webm](https://github.com/user-attachments/assets/4890c0df-5713-4d79-b292-5877466c75d6)

packages/esm-patient-list-management-app/src/lists-dashboard/lists-dashboard.component.tsx
```diff
@@ -42,6 +42,16 @@ const ListsDashboard: React.FC = () => {
...
   const { search } = useLocation();
+  const newCohortUrl = window.getOpenmrsSpaBase() + 'home/patient-lists?new_cohort=true';
+  const handleShowNewListOverlay = () => {
+    // URL navigation is in place to know either to open the create list overlay or not
+    // The url /patient-list?new_cohort=true is being used in the "Add patient to list" widget
+    // in the patient chart. The button in the above mentioned widget "Create new list", navigates
+    // to /patient-list?new_cohort=true to open the overlay directly.
+    navigate({
+      to: newCohortUrl,
+    });
```
brandones, line 53: This is very fragile. This is making in-page logic dependent on the URL remaining `home/patient-lists`. It will be very annoying if someone changes that URL and then can't figure out why adding a cohort isn't working.

I suggest that instead, you add a state `showCreatePatientList`, initialize that state using the query param, have the button modify that state, and use that state instead of `isCreatingPatientList` to determine whether to show `<CreateEditPatientList>`.

Does that make sense to you?

brandones (changes requested):

packages/esm-patient-list-management-app/src/lists-dashboard/lists-dashboard.component.tsx
brandones: We had a call and I explained it.

denniskigen (approved):
LGTM. Thanks, @Twiineenock.

PR #1283, opened 2024-08-15
(feat) O3-3404: Validate `date appointment issued` in the appointments form

lucyjemutai:
## Requirements

## Summary

https://openmrs.atlassian.net/browse/O3-3404
This PR does the following:

1. Moves the `Date appointment issued` at the bottom of the Appointment form,
2.  Ensure that the `Date appointment issued ` should never be after the appointment date. 

## Screenshots

[appointments.webm](https://github.com/user-attachments/assets/b1851cf4-cc44-43ab-b28d-dae890ff8956)

image

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.component.tsx
+++ b/packages/esm-appointments-app/src/form/appointments-form.component.tsx
@@ -192,6 +192,7 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
         message: 'Date appointment issued cannot be after the appointment date',
       },
     );
+
   type AppointmentFormData = z.infer<typeof appointmentsFormSchema>;
 
   const defaultDateAppointmentScheduled = appointment?.dateAppointmentScheduled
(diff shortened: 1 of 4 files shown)


hadijahkyampeire (approved):
LGTM, thanks @lucyjemutai

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -178,6 +178,19 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
...
+      (formValues) => {
+        const appointmentDate = formValues.appointmentDateTime?.startDate;
+        const dateAppointmentScheduled = formValues.dateAppointmentScheduled;
+
+        if (!appointmentDate || !dateAppointmentScheduled) return true;
+        return dateAppointmentScheduled < appointmentDate;
+      },
+      {
+        path: ['dateAppointmentScheduled'],
+        message: 'Date appointment issued cannot be after the appointment date',
```
mogoodrich, line 192: Can we make this localizable using t()? Thanks!

mogoodrich (changes requested):
Looks great to me, thanks @lucyjemutai , just one small request to to localize the error message.

It looks like whoever did the "A reoccurring appointment..." didn't localize that message as well, any chance you can fix that as well while you are at it? :)

packages/esm-appointments-app/src/form/appointments-form.component.tsx
lucyjemutai: Done.

Thanks @mogoodrich . Thanks @hadijahkyampeire

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -176,10 +176,22 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
       },
       {
         path: ['appointmentDateTime.recurringPatternEndDate'],
-        message: 'A recurring appointment should have an end date',
+        message: t('A recurring appointment should have an end date'),
```
mogoodrich, line 179: Thanks for following up @lucyjemutai  but I think we our best practice is to have a code (something without spaces) here followed by the default text, ie `"t('recurringAppointmentShouldHaveEndDate','A recurring appointment should have an end date').`

Am I correct @denniskigen ?

mogoodrich (changes requested):

packages/esm-appointments-app/src/form/appointments-form.component.tsx
lucyjemutai: Will we  aslo need to update the translations ? @mogoodrich

mogoodrich: I don't think so @lucyjemutai ... if you add the above` t('recurringAppointmentShouldHaveEndDate','A recurring appointment should have an end date') `I believe it will automatically add a translation mapping `'recurringAppointmentShouldHaveEndDate': ''A recurring appointment should have an end date"` to the en.json file.

mogoodrich (approved):
Looks great @lucyjemutai , merging this in!

packages/esm-appointments-app/src/form/appointments-form.component.tsx
```diff
@@ -176,10 +176,25 @@ const AppointmentsForm: React.FC<AppointmentsFormProps> = ({
...
       },
-    );
+    )
+    .refine(
+      (formValues) => {
+        const appointmentDate = formValues.appointmentDateTime?.startDate;
+        const dateAppointmentScheduled = formValues.dateAppointmentScheduled;
 
+        if (!appointmentDate || !dateAppointmentScheduled) return true;
+        return dateAppointmentScheduled < appointmentDate;
```
mogoodrich, line 188: I'm wondering if the issue with the E2E tests here when the date scheduled is the same day as the appointment date, but earlier in time?

For instance, perhaps you only schedule appointments by the day (not day and time).  So say a 10am on August 16, you try to schedule an appointment for "August 16th", and because there is no time component it says it is *before* the scheduled date.  Could we just compare the "date" (as opposed to "datetime" components of the date here)?

mogoodrich (commented):
Thanks @lucyjemutai  I was going to approve, but I just realized that one of the E2E tests is failing.  You should be able to see it in the summary on the PR?

Can you take a look?  I have a guess as to what is happened, which I noted on a comment, but I could be wrong, I didn't look into it in great detail.

mogoodrich (approved):
Thanks @lucyjemutai  for fixing the tests, merging this in!

PR #1284, opened 2024-08-16
(fix) O3-3244: Fix can not read from `undefiend` `uuid` on  inpatient Transfers

usamaidrsk:
## Requirements

## Summary

PR fixes can not read from `undefiend` `uuid` on inpatient Transfers

### DIFF
--- a/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
@@ -12,7 +12,7 @@ import { createEncounter } from '../../ward.resource';
 import useWardLocation from '../../hooks/useWardLocation';
 import type { ObsPayload, WardPatientWorkspaceProps } from '../../types';
 import { useInpatientRequest } from '../../hooks/useInpatientRequest';
-import { Form, ButtonSet, Button, TextArea, InlineNotification, RadioButtonGroup, RadioButton } from '@carbon/react';
+import { Button, ButtonSet, Form, InlineNotification, RadioButton, RadioButtonGroup, TextArea } from '@carbon/react';
 
 export default function PatientTransferForm({
   closeWorkspaceWithSavedChanges,
@@ -102,7 +102,7 @@ export default function PatientTransferForm({
       }
 
       createEncounter({
-        patient: patient.uuid,
+        patient: patient?.uuid,
         encounterType: emrConfiguration.visitNoteEncounterType.uuid,
         location: location.uuid,
         encounterProviders: [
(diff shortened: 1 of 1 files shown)


packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
```diff
@@ -102,7 +102,7 @@ export default function PatientTransferForm({
       }
 
       createEncounter({
-        patient: patient.uuid,
+        patient: patient?.uuid,
```
mogoodrich, line 105: What is the underlying issue here?  You should never be able to create an encounter without a patient, so if patient is null here there is a bigger problem?

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
```diff
@@ -229,7 +229,7 @@ export default function PatientTransferForm({
         <Button
           type="submit"
           size="xl"
-          disabled={isLoadingEmrConfiguration || isSubmitting || errorFetchingEmrConfiguration}>
+          disabled={isLoadingEmrConfiguration || isSubmitting || errorFetchingEmrConfiguration || !patient}>
```
mogoodrich, line 232: ah... is it just that we are waiting for the patient to load?  In that case, is just adding the "!patient" flag here enough?

mogoodrich (commented):
I added a couple questions, thanks @usamaidrsk

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
usamaidrsk: Right, so I have added that the save button is only enabled after the patient is loaded.

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
usamaidrsk: Yeah, the load meanwhile takes a few seconds, i.e waiting to read it from the store.

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-transfer-request-form.component.tsx
mogoodrich: Thanks @usamaidrsk .... what I was wondering was the "patient" variable would ever be null prior to "createEncounter" function being *called*.  But I think I understand now... since createEncounter is within a function that is being defined using "useCallback", it's actually trying to resolve patient.uuid *prior* to createEncounter actually being called?

mogoodrich (approved):
LGTM @usamaidrsk , thanks for the explanation. I'm going to go ahead and merge tihs in.

PR #1306, opened 2024-09-03
(fix) O3-3902 - make age() function (and its usage) handle null birth…

chibongho:
…days

## Requirements

## Summary

See [corresponding PR](https://github.com/openmrs/openmrs-esm-core/pull/1136) in core

Add a null check to prevent NPE in the age() function when a patient's birthdate is null. There are also places, in core and patient-management, where this is insufficient, when UI elements should just not render in that case. This PR adds null check in those places too.

## Related Issue

https://openmrs.atlassian.net/browse/O3-3902


denniskigen (approved):
LGTM, @chibongho. Thanks!

PR #1311, opened 2024-09-06
(fix) O3-3915: The buttons in the refine-search overlay for the patient search button are partially hidden in a desktop layout

mccarthyaaron:
## Requirements

## Summary
This PR addresses an issue in the refine search overlay component under the patient search. On desktop the buttons for submitting or discarding the refined search are partially hidden.
This seems to be caused by the fact that the `Overlay` component [here](https://github.com/mccarthyaaron/openmrs-esm-patient-management/blob/main/packages/esm-patient-search-app/src/ui-components/overlay/index.tsx) has a height of 100vh as well as a position from the top of 3rem. This causes it to immediately go below the viewport. The actual issue then seems to arise down the component hierarchy [here](https://github.com/mccarthyaaron/openmrs-esm-patient-management/blob/main/packages/esm-patient-search-app/src/patient-search-page/refine-search.scss#L70) with the refine seach dialog container. It is positioned `bottom: 0` (which bottom is below the screen). Couple that with the `justify-content: flex-end` and you then end up with the the containers content partially hidden below the viewport.

the suggested solution here is to remove the height property from the Overlay component and give it  `bottom: 0`. Such that it sticks to the bottom of the viewport.

## Screenshots

https://github.com/user-attachments/assets/919003ba-10ee-468a-9535-51fc454b3046

## Related Issue
https://openmrs.atlassian.net/browse/O3-3915

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-page/refine-search.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/refine-search.component.tsx
@@ -292,7 +292,7 @@ const RefineSearch: React.FC<RefineSearchProps> = ({ setFilters, inTabletOrOverl
   }
 
   return (
-    <form onSubmit={handleSubmit} className={styles.refineSeachContainer} data-openmrs-role="Refine Search">
+    <form onSubmit={handleSubmit} className={styles.refineSearchContainer} data-openmrs-role="Refine Search">
       <h2 className={styles.productiveHeading02}>{t('refineSearch', 'Refine search')}</h2>
       <div className={styles.field}>
         <div className={styles.labelText}>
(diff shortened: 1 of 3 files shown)


brandones (approved):
LGTM. Eventually we want to move the patient search over to using the workspace system. Thanks for the bug ticket & fix @mccarthyaaron .

mccarthyaaron:
> LGTM. Eventually we want to move the patient search over to using the workspace system. Thanks for the bug ticket & fix @mccarthyaaron .

You are welcome @brandones . Concerning moving it to using the workspace system is there any ongoing work or discussion about it? and if so where can one follow it from

brandones:
Great question @mccarthyaaron , I should have linked the ticket before. Here it is: https://openmrs.atlassian.net/browse/O3-3326

PR #1312, opened 2024-09-07
(feat) O3-3861 - ward app - add tooltip to obs to show encounter date

chibongho:
## Requirements

## Summary

The observations displayed in the ward patient cards current has no indication to show when the obs values were taken. If the values were mis-configured to show from previous visits then the values might not be accurate.

This PR introduces ToggleTips that allow user to click on the obs values and see the corresponding encounter date.

## Screenshots

image
image
image

## Related Issue

https://openmrs.atlassian.net/browse/O3-3861


packages/esm-ward-app/src/hooks/useMotherAndChildren.ts
```diff
@@ -44,7 +38,7 @@ export function useMotherAndChildren(
     requireChildBornDuringMothersActiveVisit?.toString() ?? 'false',
   );
   rep && url.searchParams.append('v', rep);
-  return useOpenmrsPagination<MotherAndChildren>(fetch ? url : null, 50);
+  return useOpenmrsFetchAll<MotherAndChildren>(fetch ? url : null);
```
chibongho, line 41: Sneaking this fix in. This shouldn't have use `useOpenmrsPagination`.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx
```diff
@@ -62,16 +68,6 @@ const WardPatientCard: WardPatientCard = (wardPatient) => {
         ))}
         <ExtensionSlot name={footerExtensionSlotName} state={wardPatient} />
       </div>
-      <button
```
chibongho, line 65: Previously, the patient card was made clickable by adding this button to overlay on top of the elements of the card. I'm not sure why it was implemented that way, but I had to change it in order for the ToggleTips to receive click events.

cc @brandones

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-obs.tsx
```diff
@@ -37,7 +37,19 @@ const WardPatientObs: React.FC<WardPatientObsProps> = ({ config, patient, visit
...
-      return <span key={o.uuid}> {display} </span>;
+      return (
+        <span key={o.uuid}>
+          {display}
+          <span onClick={(e) => e.stopPropagation()}>
+            <Toggletip className={styles.wardPatientObsIcon} onClick={(e) => e.stopPropagation()}>
+              <ToggletipButton>
+                <Information />
+              </ToggletipButton>
+              <ToggletipContent>{o.encounter?.display}</ToggletipContent>
```
mogoodrich, line 48: Same comment about display.

mogoodrich (commented):
Thanks @chibongho ! Looks good, though one comment on formatting the display information.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.component.tsx
brandones: It is [widely and strongly recommended against](https://theadminbar.com/accessibility-weekly/divs-are-not-buttons/) adding onClick to a div.

There are myriad articles about how to create a "clickable card" component, which is simply kind of a hard problem, but I really like [this one](https://css-tricks.com/block-links-the-search-for-a-perfect-solution/). I implemented [Method 3](https://css-tricks.com/block-links-the-search-for-a-perfect-solution/#aa-method-3-the-good-ol-before-pseudo-element) from it, which agrees with the recommendations of the widely-cited [Card guidelines](https://inclusive-components.design/cards/) from Inclusive Components.

Rather than using JS as recommended by Method 4 of that article, I would recommend just using the current approach and increasing the z-index of any links you need to be clickable on top of the card. The only reason Method 4 would be preferable would be to make the text selectable, and I don't think that's relevant to this interface.

As an aside, did this ToggleTip thing get designed? It makes the interface look a lot busier, appears to have the info button at the same level of visual hierarchy as the obs information itself, and seems like it would be easy to click the wrong thing. I would suggest making the tooltip appear on hover anywhere over the tag element (with a 500ms delay). You would only need JS to do touch handling on mobile, something like touch-and-hold shows the tooltip while tap clicks the card. You would still need the z-index fix.

Alternatively, since it seems like this is supposed to solve a problem not actually faced by users (but rather by devs & admins) you could add a boolean config option that makes all tags show some metadata. Then you would be able to debug just by switching it in the implementer tools.

chibongho: Ok, makes sense. I'll implement it using z-index. 

> As an aside, did this ToggleTip thing get designed?

No, I did this before I got a [reply from UX designers](https://openmrs.slack.com/archives/C06P2GVUGEP/p1725656280125159). I'll make the whole tag clickable for mobile and hoverable on desktop.

> Alternatively, since it seems like this is supposed to solve a problem not actually faced by users (but rather by devs & admins)

While it's mostly for debugging, I think it's also important to show this info so that mistakes in implementation / configuration are as apparent as possible to the end users.

mogoodrich (approved):
LGTM!

I'm not sure encounterType.display supports localization, but if it doesn't we should fix on the server side when we encounter the issue.

denniskigen:
From the screenshots, it looks like the Toggletips could benefit from some left margin.

chibongho:
> From the screenshots, it looks like the Toggletips could benefit from some left margin.

Oh... I should have mentioned. The screenshots are not up to date anymore. I'm making the entire tags be a tooltip on hover when on desktop and a toggletip when on mobile. It looks like this now:

image

image

PR #1313, opened 2024-09-09
(feat) HIE-9: Add MPI workflows to OpenMRS frontend

reagan-meant:
## Requirements

## Screenshots

https://github.com/user-attachments/assets/7e9fbee1-b41d-434c-851e-d7e79c6af842

## Related Issue

https://openmrs.atlassian.net/browse/HIE-9


packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,45 @@
+import capitalize from 'lodash-es/capitalize';
```
ibacher, line 1:
```suggestion
import { capitalize } from 'lodash-es';
```

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function inferModeFromSearchParams(searchParams: URLSearchParams) {
+  return searchParams?.get('mode')?.toLowerCase() == 'external' ? 'external' : 'internal';
```
ibacher, line 4:
```suggestion
  return searchParams.get('mode')?.toLowerCase() === 'external' ? 'external' : 'internal';
```

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+export function mapToOpenMRSPatient(fhirPatients: any): any {
```
ibacher, line 7: This could also benefit from explicit types on both sides. Presumable `fhirPatients` is a `fhir.Bundle` and each entry is a `fhir.Patient` or similar.

ibacher: I think there's also an OpenMRS patient type in core that can be the return type.

ibacher (changes requested):
A few more nit-picky comments. I can review more later.

packages/esm-patient-registration-app/src/patient-registration/mpi/use-mpi-patient.ts
```diff
@@ -0,0 +1,27 @@
+import { fhirBaseUrl, openmrsFetch } from '@openmrs/esm-framework';
+import { useEffect, useState } from 'react';
+
+export function useMPIPatient(patientId: string) {
+  const [patient, setPatient] = useState<fhir.Patient>(null);
+  const [isLoading, setIsLoading] = useState(false);
+
+  let url = `${fhirBaseUrl}/Patient/${patientId}/$cr`;
```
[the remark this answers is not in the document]
reagan-meant: Done

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+          cityVillage: address.city,
+          stateProvince: address.state,
+          country: address.country,
+          postalCode: address.postalCode,
+          preferred: false,
+          address1: '',
+        })),
+        age: null,
+        birthdate: fhirPatient.birthDate,
+        gender: capitalize(fhirPatient.gender),
```
ibacher, line 31: Rather than this, it would actually be better to do: `getCoreTranslations(fhirPatient.gender)`

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+        dead: !fhirPatient.active,
```
ibacher, line 32: This seems wrong or, at the very least, very specific. `dead` corresponds to FHIR's `deceasedBoolean` (and `deathDate` with `deceasedDateTime`). `active === false` just means "this patient record is no longer actively used in this system".

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+        deathDate: '',
+        personName: {
+          display: `${fhirPatient.name[0].family} ${fhirPatient.name[0].given[0]}`,
```
ibacher, line 35: I always dislike inferring what the display should be for a name, since this is very location-sensitive, e.g., how its written, how it's formatted etc. I'd prefer that we either pull from the `name[0].text` first.

packages/esm-patient-search-app/src/patient-search-button/patient-search-button.test.tsx
```diff
@@ -37,8 +38,13 @@ describe('PatientSearchButton', () => {
   it('displays overlay when button is clicked', async () => {
     const user = userEvent.setup();
 
-    render(<PatientSearchButton />);
-
+    render(
+      <BrowserRouter>
```
ibacher, line 42: It's probably better to use a MemoryBrowser in tests.

packages/esm-patient-search-app/src/patient-search-overlay/patient-search-overlay.component.tsx
```diff
@@ -5,6 +5,8 @@ import AdvancedPatientSearchComponent from '../patient-search-page/advanced-pati
 import Overlay from '../ui-components/overlay';
 import PatientSearchBar from '../patient-search-bar/patient-search-bar.component';
 import { type PatientSearchConfig } from '../config-schema';
+import { inferModeFromSearchParams } from '../mpi/utils';
+import { useSearchParams } from 'react-router-dom';
```
ibacher, line 9: Very minor, but this import should come before all relative imports.

packages/esm-patient-search-app/src/patient-search-page/patient-search-views.component.tsx
```diff
@@ -85,20 +91,54 @@ export const SearchResultsEmptyState: React.FC<CommonProps> = ({ inTabletOrOverl
...
+                  'trySearchFromClientRegistry',
+                  "You can try searching using the patient's unique ID number or search the external MPI",
+                )}
+              </p>
+            </div>
+            <Button
+              kind="ghost"
+              renderIcon={'Search'}
+              onClick={(e) => {
+                e.preventDefault();
```
ibacher, line 111: Do we need the `preventDefault()` call here?

ibacher (approved):
A few small things, but basically LGTM

bmamlin:
Do we expect users to know what "MPI" means? I would think "Check for national chart" or "Search health exchange" would be easier for providers/users to understand.

reagan-meant:
@bmamlin How about 'Search External Registry'

packages/esm-patient-search-app/src/index.ts
```diff
@@ -35,6 +36,8 @@ export const patientSearchBar = getSyncLifecycle(patientSearchBarComponent, opti
 export function startupApp() {
   defineConfigSchema(moduleName, configSchema);
 
+  registerFeatureFlag('mpiFlag', 'MPI Service', 'Enables the Master Patient Index workflows');
```
ibacher, line 39: Probably should've said this: our preferred means of registering feature flags is via `routes.json`. Basically add the same properties, but to a `featureFlag` definition there.

ibacher (approved):

denniskigen:
Could you move this along, @reagan-meant?

reagan-meant:
@denniskigen are you available to review?

brandones:
@reagan-meant , could you fix the conflicts? Looks like @ibacher has approved. Once you fix the conflicts I'll give it a once over and merge.

denniskigen:
Thanks for working on this, @reagan-meant. I’ll merge it in shortly. There are a few areas we need to tweak, but that can happen once this gets checked in. I’m curious to see how this evolves!

reagan-meant:
@denniskigen here is the relevant backend [module](https://github.com/openmrs/openmrs-module-clientregistry)

denniskigen:
Is the plan to include it in the distro reasonably soon? Also, because this functionality is contigent on the module, we should include it in the `optionalBackendDependency` key in the `routes.json` files of the Search and Registration frontend modules.

ibacher:
We'd like to get it in reasonably soon, but I'm not sure it's going to be enabled for, e.g., the dev3 / test3 / o3 stack. This is maybe something we should discuss on the engineering call. Basically, I'm thinking about spinning up an integrated MPI demo that has, e.g., a demo CR linked to it and I think that's a separate stack in the sense that it wouldn't make sense in the RefApp. However, I am concerned that having functionality not activated in dev3 would make it very easy for that functionality to break. That's probably something we can solve with e2e tests, but we'd need a strategy to include some kind of dummy CR in the e2e tests.

denniskigen:
+1 for discussing it on the eng leads call. Thanks.

donaldkibet:
@reagan-meant @denniskigen @ibacher would like to use this work on cross border integration, is there an agreement on how we can have this move this forward.

ibacher:
> My initial thought is that we probably shouldn't, which would mean restricting the external search to only the main search page.

In general, querying an MPI is a registration task and should probably be limited to that if that's feasible.

ibacher:
> @ibacher, do you think we could add an attribute or tag to the patient [resource](https://github.com/openmrs/openmrs-esm-patient-management/pull/1313/files#diff-815ba10fcc983ee4592625a7ddaccc43fe379f506529e3099a1b2121f404a46cR16) to enable conditional UI rendering, instead of relying on route parameters?

Say more...

ibacher:
@donaldkibet It's basically what Dennis said. We just need to ensure that we're not breaking search in other contexts. We may need to separate the "MPI search" out into it's own modal or something?

reagan-meant:
@ibacher @denniskigen here are my [suggestions](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397/commits/8919b2d8aff90d4019ef7fd637b640d1b3ade3e3)
1. Make [this](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397/commits/8919b2d8aff90d4019ef7fd637b640d1b3ade3e3#diff-5f427848433b934389cb9503c249345c05434ce369413f0ac8b3d2eea01eea3aR85) path configurable and use it to determine if we are on the main search page
2. Check if [router](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397/commits/8919b2d8aff90d4019ef7fd637b640d1b3ade3e3#diff-cd60883bce1cf46f14c5ba8d52bc6ebbaf775016bb538cdfe634b79bd88ecaadR37-R43) is accessible and determine if external query or not.
These changes fix the issue.

ibacher:
For:

1. A better solution is just `window.location.pathname === `${getOpenmrsSpaBase()}search`
2. `try... catch` around hooks is not a pattern I'd want to encourage using. I think the better solution here is to write a `useSearchParams()` hook that doesn't depend on React-Router and add it to react-utils in the framework.

ibacher:
Re: 2 something like:

```ts
export const useSearchParams() {
    const [searchParams, setSearchParams] = useState(new URLSearchParams(window.location.search));

    useEffect(() => {
       const updateSearchParams = () => setSearchParams(new URLSearchParams(window.location.search));
       window.addEventListener('popstate', updateSearchParams);
       return () => window.removeEventListener('popstate', updateSearchParams);
    }, []);

    return searchParams;
}
```

ojwanganto:
Ping @donaldkibet. I know you have been looking into this as well.

PR #1317, opened 2024-09-17
(feat) Add text-to-speech voice call support for ticket numbers in queue screen

FelixKiprotich350:
…queues

## Requirements

## Summary
This feature aims at voice calling ticket number in Service Queue by using Text-to-Speech web api 

## Screenshots
https://drive.google.com/file/d/1i_mW3YTR6XcOHsdrGbUiQ0bLMz8M8Ju5/view?usp=sharing

## Related Issue
https://openmrs.atlassian.net/browse/O3-3995


packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
+        message.onend = resolve;
+        speaker.speak(message);
+      });
+    } else {
+      return Promise.resolve();
+    }
+  }
+
+  useEffect(() => {
+    const readableTickets = locationFilteredTickets.filter((item) => item.status == 'calling');
```
ojwanganto, line 49: readableTickets may mean some are not readable. We can adopt a name i.e. ticketsToCallOut or something else

FelixKiprotich350: resolved

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
 
 interface QueueScreenProps {}
 
 const QueueScreen: React.FC<QueueScreenProps> = () => {
   const { t } = useTranslation();
-  const { activeTickets, isLoading, error } = useActiveTickets();
+  const { activeTickets, isLoading, error, mutate } = useActiveTickets();
+  const speaker = window.speechSynthesis;
+  const [isSpeaking, setIsSpeaking] = useState(false);
+  const selectedLocation = useSelectedQueueLocationUuid();
```
denniskigen, line 16: Looks like we're not using this anywhere

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
+      message.pitch = 1;
+      message.text = utterance;
+      return new Promise((resolve) => {
+        message.onend = resolve;
+        speaker.speak(message);
+      });
+    } else {
+      return Promise.resolve();
+    }
+  }
```
denniskigen, line 46: This could be moved into the component function body and memoized with a `useCallback`:

```tsx
  const readTicket = useCallback((queue) => {
    if ('speechSynthesis' in window) {
      const message = new SpeechSynthesisUtterance();
      const [prefix, suffix] = queue.ticketNumber.split('-');
      const utterance = `Ticket Number: ${prefix.split('')}, - ${suffix.split('')}, Please Proceed To Room, ${queue.room}`;
      message.rate = 1;
      message.pitch = 1;
      message.text = utterance;
      return new Promise<void>((resolve) => {
        message.onend = resolve;
        speaker.speak(message);
      });
    }
    return Promise.resolve();
  }, [speaker]);
```

Also, if the `SpeechSynthesisUtterance` API is locale-aware (it probably is), we should consider internationalising `utterance` using something like:

```tsx
	 const utterance = t(
          'ticketAnnouncement',
          'Ticket number: {{prefix}}, - {{suffix}}, please proceed to room {{room}}',
          {
            prefix: prefix.split('').join(' '),
            suffix: suffix.split('').join(' '),
            room: queue.room,
          },
        );
```

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
+
+  useEffect(() => {
+    const ticketsToCallOut = locationFilteredTickets.filter((item) => item.status == 'calling');
```
denniskigen, line 49:
```suggestion
    const ticketsToCallOut = locationFilteredTickets.filter((item) => item.status === 'calling');
```

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
+    if (ticketsToCallOut.length > 0 && !isSpeaking) {
+      setIsSpeaking(true);
+      const readTickets = async () => {
+        for (const ticket of ticketsToCallOut) {
+          await readTicket(ticket, speaker);
+        }
+        setIsSpeaking(false);
+        if (typeof mutate === 'function') {
+          mutate();
+        }
```
denniskigen, line 59:
```suggestion
        mutate?.();
```

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -19,13 +73,6 @@ const QueueScreen: React.FC<QueueScreenProps> = () => {
     return <div>Error</div>;
```
denniskigen, line 73: We need a better error state here. Consider using the `ErrorState` component from the framework:

```tsx
 return <ErrorState error={error} headerTitle={t('queueScreenError', 'Queue screen error')} />;
```

denniskigen:
Thanks for taking a stab at this, @FelixKiprotich350. I've left a few suggestions around optimising the code, handling internationalisation, and error states.

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
denniskigen, line 16: Looks like we're not using this variable. If so, we should remove it.

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
+  const [isSpeaking, setIsSpeaking] = useState(false);
+  const selectedLocation = useSelectedQueueLocationUuid();
+  const locationFilteredTickets = activeTickets;
+
+  const rowData = locationFilteredTickets.map((ticket, index) => ({
+    id: `${index}`,
+    room: ticket.room,
+    ticketNumber: ticket.ticketNumber,
+    status: ticket.status,
+  }));
```
denniskigen, line 24: Consider memoizing `rowData`:

```tsx
  const rowData = useMemo(() => 
    activeTickets.map((ticket, index) => ({
      id: `${index}`,
      room: ticket.room,
      ticketNumber: ticket.ticketNumber,
      status: ticket.status,
    })),
  [activeTickets]);
```

Also, is there something unique in the response that we could use as the `id` instead of `index`? Maybe a UUID?

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,15 +1,69 @@
...
 interface QueueScreenProps {}
 
 const QueueScreen: React.FC<QueueScreenProps> = () => {
   const { t } = useTranslation();
-  const { activeTickets, isLoading, error } = useActiveTickets();
+  const { activeTickets, isLoading, error, mutate } = useActiveTickets();
+  const speaker = window.speechSynthesis;
+  const [isSpeaking, setIsSpeaking] = useState(false);
+  const selectedLocation = useSelectedQueueLocationUuid();
+  const locationFilteredTickets = activeTickets;
```
denniskigen, line 17: Did you mean to filter `activeTickets` by location here? If not, then we probably don't need the alias and we should use `activeTickets` directly.

FelixKiprotich350: i did this because ill be making a follow up later to filter by location, am waiting for  someone to include the locaion uuid in the response so that i can utilize that,,,

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
FelixKiprotich350: @denniskigen currently the response does not have uuid thats why i opted to use index

ibacher:
This option needs to be behind some kind of configuration. I don't think it should be enabled by default. I'm also slightly suspicious that embedding this in a frontend module is the right approach. While it works for a small clinic, I don't think it is a great idea for larger clinics. Basically, this is the type of feature that needs some community discussion before it's mergeable, but I like it!

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
denniskigen: Best to not include it until then

ojwanganto:
> This option needs to be behind some kind of configuration. I don't think it should be enabled by default. I'm also slightly suspicious that embedding this in a frontend module is the right approach. While it works for a small clinic, I don't think it is a great idea for larger clinics. Basically, this is the type of feature that needs some community discussion before it's mergeable, but I like it!

This will be helpful.

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
FelixKiprotich350: well gotten

FelixKiprotich350:
Yea sure ,,, with the community discussions i believe we will have good ideas on how best we can implement this and configure dynamically

FelixKiprotich350:
> This option needs to be behind some kind of configuration. I don't think it should be enabled by default. I'm also slightly suspicious that embedding this in a frontend module is the right approach. While it works for a small clinic, I don't think it is a great idea for larger clinics. Basically, this is the type of feature that needs some community discussion before it's mergeable, but I like it!

this is true @ibacher , but how do you see it given that the feature can be used dynamically with the help of some configurations. Given an example of a large clinic/facility with multiple queue locations and service points, where every point or Queue has a screen to call ticketes specifically for that point?

ibacher:
@FelixKiprotich350 If I had exact thoughts on how this kind of feature should work, I would've included them. That's why I think we need a community discussion on Talk, because I'm not exactly sure what the best behaviour is here. My guess is that there's a lot of additional layers that are needed, i.e., for a user viewing the queue to select to be a system that automatically calls out tickets so that, e.g., a clinic manager or clinician viewing the queue in an office doesn't need to be bothered. Similarly, with a larger clinic, there may be two or more, e.g., nurse stations working on the same queue, in which case only one should probably be announcing.

But this requires more thought than is ideal to document on a PR in comments, hence, a Talk post would be ideal.

FelixKiprotich350:
Agree..... You can guide on how to go about it

mogoodrich:
I tend to agree with @ibacher , thanks... fyi @mseaton  @chibongho

denniskigen:
Could we wire up some test coverage for this new functionality?

FelixKiprotich350:
> Could we wire up some test coverage for this new functionality?

yes sure i can do that

packages/esm-service-queues-app/src/queue-screen/queue-screen.component.tsx
```diff
@@ -1,31 +1,88 @@
...
+        status: ticket.status,
+      })),
+    [activeTickets],
+  );
+  const readTicket = useCallback(
+    (queue) => {
+      if ('speechSynthesis' in window) {
+        const message = new SpeechSynthesisUtterance();
+        const [prefix, suffix] = queue.ticketNumber.split('-');
+        // const utterance = `Ticket Number: ${prefix.split('')}, - ${suffix.split('')}, Please Proceed To Room, ${queue.room}`;
```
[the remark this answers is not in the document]
FelixKiprotich350: noted

brandones:
Agree with @ibacher that more discussion is needed. Since there's no Talk link posted here I assume there's no Talk thread and will share my thoughts here.

My suggestion would be that there should just be a page for this functionality. If someone wants their computer to make announcements for the queue, they just leave a tab open on that page.

Eventually, and this would need design, the full vision might be a "queue display" page which has a toggle for whether to make audio announcements and which has a button that will open a separate window that is intended to be displayed full-screen on a monitor facing the patients in the waiting room. Like Google Slides "presenter view" feature.

brandones:
Since no one is responding and there have been no updates I'm going to close this for now. Feel free to re-open once things get going again.

PR #1321, opened 2024-09-19
(fix) O3-4011 - Ward App - make styling for patient card's patient id…

chibongho:
…entifier consistent

## Requirements

## Summary

Styling of the patient identifier in the patient card's header row vs the mother child row were inconsistent. This PR fixes it.

I ended up having to change the mother-child row to not hard-code the elements (and instead define them in config) so that the same element config can be used in both the header row and mother -child row.

I also made it so that if the label for the patient identifier type is empty, we display the patient identifier (instead of the patient identifier type) in a carbon Tag. This should make the identifier fairly prominent and easy to find.

## Screenshots

Before:
image

After:
image

### DIFF
--- a/packages/esm-ward-app/src/config-schema.ts
+++ b/packages/esm-ward-app/src/config-schema.ts
@@ -1,7 +1,5 @@
 import { type ConfigSchema, Type, validators } from '@openmrs/esm-framework';
 
-export const builtInPatientCardElements = ['patient-age', 'time-on-ward', 'time-since-admission', 'patient-location'];
-
 export const addressFields = [
   'cityVillage',
   'stateProvince',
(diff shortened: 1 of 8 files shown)


mogoodrich (approved):

chibongho:
I ended up changing the implementation so that the patient identifier is rendered from a `esm-styleguide` component instead. This makes it so the configurations there (namely, to specify what patient identifier types are excluded) are applicable to the patinet card too.

New changes depends on this PR:
https://github.com/openmrs/openmrs-esm-core/pull/1159

Converted this one to draft since the yarn.lock file depends on local changes not changed into esm-core yet.

mogoodrich (approved):
Works for me (assuming we resolve the local lock stuff first)

PR #1339, opened 2024-10-10
(feat) O3-4078 - Ward App - Simplify patient card configurations

chibongho:
## Requirements

## Summary

See [O3-4078](https://openmrs.atlassian.net/browse/O3-4078) and relevant [slack thread](https://openmrs.slack.com/archives/C06P2GVUGEP/p1727443589822479?thread_ts=1727382096.520819&cid=C06P2GVUGEP).

This refactoring touches a lot of files (sorry), but there is almost no functional change besides how we configure the patient cards. 
- The ward app now only has 1 extension. It is the ward view (containing the metrics header and the beds) at the top level. Right now, we implemented a "default ward" and a "maternal ward", and which ward type we want can be configured by location.
  - Consequently, what used to be extension rows in the patient card (like mother-child-row), are now elements, hence a lot of renaming and moving around their configurations.
- `DefaultWardView` and `MaternalWardView` both define their own data fetching and card rendering. What elements appear in those cards are hard-coded (thus not configurable). If another implementation wants a different looking ward, they will need to make their own WardView. (Currently, they can only make that within the esm ward app). 
  - This should be fine in the short term. Longer term, we should have a way for different implementations to create their own esm defining their ward view using the elements / components defined in this esm.
- Elements in those respective cards can be individually configurable. Each element has an unique id hard-coded in the `DefaultWardView` or `MaternalWardView`, and the element configuration requires specifying the id. This means whoever is configuring the elements will need to know what id's are hard-coded into the WardView they want to use.

Most relevant files to review:
- config-schema.tsx
- default-ward-view.component.tsx and corresponding default-ward-specific components (DefaultWardBeds, DefaultWardUnassignedPatients, DefaultWardPendingPatients)
- maternal-ward-view.component.tsx and corresponding maternal-ward-specific components (MaternalWardBeds, MaternalWardUnassignedPatients, MaternalWardPendingPatients)

With this change, the configuration for maternal wards in MCOE is simpler:
```
"patientCardElements": {
      "obs": [
        {
          "id": "gravida",
          "conceptUuid": "3cee82de-26fe-102b-80cb-0017a47871b2",
          "onlyWithinCurrentVisit": true,
          "orderBy": "descending",
          "limit": 1
        },
        {
          "id": "admission-reason",
          "conceptUuid": "86a2cf11-1ea5-4b8a-9e4b-08f4cdbe1346",
          "onlyWithinCurrentVisit": true,
          "orderBy": "descending",
          "label": "REASON",
          "limit": 1
        }
      ],
      "patientIdentifier": [
        {
          "id": "patient-identifier",
          "identifierTypeUuid": "c09a1d24-7162-11eb-8aa6-0242ac110002",
          "label": ""
        }    
      ],
      "patientAddress": [
        {
          "id": "patient-address",
          "fields": [
            "countyDistrict"
          ]
        }
      ],
      "coloredObsTags": [
        {
          "id": "pregnancy-complications",
          "conceptUuid": "447f14be-4f34-4cba-9f80-8e091e23c6e3",
          "summaryLabel": "Pregnancy Complication(s)",
          "tags": [
            {
              "color": "red",
              "appliedToConceptSets": ["447f14be-4f34-4cba-9f80-8e091e23c6e3"]
            }
          ]
        }
      ],
      "admissionRequestNote": [
        {
          "id": "admission-request-note",
          "conceptUuid": "3cd9d956-26fe-102b-80cb-0017a47871b2"
        }
      ],
      "pendingItems": [
        {
          "id": "pending-items",
          "orders": {
            "orderTypes": [
              {
                "label": "Labs", 
                "uuid": "52a447d3-a64a-11e3-9aeb-50e549534c5e"
              }
            ]
          },
          "showPendingItems": true
        }
      ]
    },
    "wards": [
      {
        "id": "maternal-ward"
      }
    ]
```

## Screenshots

image
image

## Other

I introduced a bug with the baby row appearing twice within a mother's patient card sometimes (highlighted in screenshot). I will address that in [this ticket](https://openmrs.atlassian.net/browse/O3-4022).

[O3-4078]: https://openmrs.atlassian.net/browse/O3-4078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-ward-app/src/config-schema.ts
```diff
@@ -132,66 +168,115 @@ export const configSchema: ConfigSchema = {
...
-  wardPatientCards: WardPatientCardsConfig;
-}
-
-export interface WardPatientCardsConfig {
-  obsElementDefinitions: Array<ObsElementDefinition>;
-  pendingItemsDefinitions: Array<PendingItemsDefinition>;
-  identifierElementDefinitions: Array<IdentifierElementDefinition>;
-  addressElementDefinitions: Array<AddressElementDefinition>;
-  cardDefinitions: Array<WardPatientCardDefinition>;
+  patientCardElements: {
```
mogoodrich, line 268: Potential rabbit-hole... I guess I was thinking that each WardView would have it's own config object?  When we get to the point where a custom WardView can be defined outside the Ward App ESM, is there a way for the developers of new Ward Views to define new config parameters without adding to the Ward View config?  (For discussion, let's not let this be a blocker for merging)

mogoodrich (approved):
Thanks for slogging through this, @chibongho !  Yeah, I didn't make it 100% through this, but generally looks good and I agreed with the strategy based on our discussion earlier this week about this.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -79,9 +73,51 @@ export const configSchema: ConfigSchema = {
         },
       },
     },
-    identifierElementDefinitions: {
+    pendingItems: {
+      _type: Type.Array,
+      _description: 'Configures pending orders and transfers to display.',
+      _default: [
+        {
+          id: 'pending-items',
```
usamaidrsk, line 81: I think the id here would be pending-orders

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -37,11 +30,12 @@ export const addressFields = [
 type AddressField = keyof typeof addressFields;
 
 export const configSchema: ConfigSchema = {
-  wardPatientCards: {
-    _description: 'Configure the display of ward patient cards',
-    obsElementDefinitions: {
+  patientCardElements: {
+    _description:
+      'Configuration of various patient card elements. Each configured element must have an unique id, defined in the ward React component being used.',
```
usamaidrsk, line 35: Fixing some small grammar:
 `Configuration of various patient card elements. Each configured element must have a unique id, defined in the ward React component being used.`

packages/esm-ward-app/src/config-schema.ts
usamaidrsk, line 81: I think the `id` here would be `pending-orders`

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -79,9 +73,51 @@ export const configSchema: ConfigSchema = {
...
+          _description:
+            'Optional. If true, pending items (e.g., number of pending orders) will be displayed on the patient card.',
+        },
+      },
+    },
+    patientIdentifier: {
       _type: Type.Array,
-      _description: `Defines patient identifier elements that can be included in the card header or footer. The default element 'patient-identifier' displays the preferred identifier.`,
+      _description: `Configures patient identifier to display. An unconfigured element displays the preferred identifier.`,
       _default: [
```
usamaidrsk, line 121: looks like the `_default` should be an empty array, provided the default object passed is missing some keys i.e `identifierTypeUuid`

packages/esm-ward-app/src/config-schema.ts
usamaidrsk: If we could have a way to automatically generate the keys, it would significantly reduce the risk of human error.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -132,66 +168,115 @@ export const configSchema: ConfigSchema = {
...
+    patientIdentifier: Array<IdentifierElementConfig>;
+    patientAddress: Array<PatientAddressElementConfig>;
+    coloredObsTags: Array<ColoredObsTagsElementConfig>;
+    admissionRequestNote: Array<AdmissionRequestNoteElementConfig>;
+  };
+  wards: Array<WardDefinition>;
 }
 
-export interface PendingItemsDefinition {
+export interface PendingItemsElementConfig {
```
usamaidrsk, line 279: I think you missed the `id` here.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -210,25 +295,68 @@ export interface ObsElementDefinition {
...
 }
 
-export interface AddressElementDefinition {
+export interface PatientAddressElementConfig {
   id: string;
   fields: Array<AddressField>;
 }
 
-export interface WardPatientCardDefinition {
+export interface AdmissionRequestNoteElementConfig {
```
usamaidrsk, line 309: Same here, `id` was missed.

packages/esm-ward-app/src/config-schema.ts
chibongho: > If we could have a way to automatically generate the keys, it would significantly reduce the risk of human error.

This might be difficult, with how the ward view can be an extension outside the ward app.

packages/esm-ward-app/src/config-schema.ts
chibongho: For a custom WardView outside this ESM, it can have patient cards with a mixture of elements defined in this ESM and outside it. I think ideally we'd have the elements configurations be tied to whichever ESM it's defined in. We'll want to be careful as what `useConfig()` returns is dependent on which ESM / extension the component is in. I haven't done something like this before, but it feels doable.

chibongho:
> Thanks @chibongho , I have a few comments. All in all great work, and thanks for the tests!

Thanks for the comments. I was getting a bit sloppy with the config documentations.

I'm going to merge this in. Post-hoc comments welcome.

packages/esm-ward-app/src/config-schema.ts
ibacher: Probably the _best_ solution here is for the ward app to consume whatever config options it needs to consume and pass those as props to the elements or something along those lines...

PR #1341, opened 2024-10-11
(fix) O3-3787 : Refresh workspace title when clicked on new patient

kb019:
## Requirements

## Summary
I have mentioned detailed solution in this [PR](https://github.com/openmrs/openmrs-esm-core/pull/1177). For now this might be like a temporary solution.
The workspace title which was set dynamically does not updated when clicked on other patients.Putting the props which are passed to the title node as a dependency in useEffect works.
[OpenMRS - Personal - Microsoft_ Edge 2024-10-10 21-26-09.webm](https://github.com/user-attachments/assets/109330d6-b3c3-46e7-b7d0-5725a3505eda)

## Related Issue
https://openmrs.atlassian.net/browse/O3-3787


chibongho (approved):
Wow, that's it? Nice work!

kb019:
> Wow, that's it? Nice work!

Thanks @chibongho .

brandones:
I think this is the correct way to solve the problem.

denniskigen:
https://react.dev/learn/preserving-and-resetting-state#option-2-resetting-state-with-a-key

PR #1345, opened 2024-10-14
(feat) Restore recently searched patients functionality in compact search

denniskigen:
## Requirements

## Summary

This PR restores the recently searched patients functionality in the compact patient search. It also introduces some refactoring of the code to make it more modular and easier to maintain. These changes include:

- Tweaking the configuration schema to improve the names and descriptions of the config properties. This also includes renaming the `patientResultUrl` config property to `patientChartUrl`, which is more descriptive of what it is used for. Implementers will need to update their configurations to use the new key.
- Cleaning up the logic in the `useInfinitePatientSearch`, `useRecentlyViewedPatients` and `useRestPatients` hooks to make them more modular and easier to maintain. This includes removing rendering logic from the hooks. We can now drop the `.tsx` extension for the patient search resource file as it no longer contains any JSX.
- Annotating invocations of the `useConfig` hook with the `PatientSearchConfig` type for better type inference.
- Adding some error handling logic to the `addViewedPatientAndCloseSearchResults` function.
- Preloading user properties from the REST API when the user hovers over the search icon in the navbar using the SWR preload API.
- Preserving recently searched patients when revalidating the recently viewed patients list by enabling the `keepPreviousData` option in the `useInfinitePatientSearch` hook for better UX. Without this, the UI would show a loading state when revalidating the recently viewed patients list which could be potentially jarring for the user.

## Screenshots

https://github.com/user-attachments/assets/95bf661d-4de3-4fce-b023-939c4ae2cef7

## Other

This implementation is a slight departure from what was envisioned in the [original design](https://zpl.io/JEnOvox). It's also a departure from a similar feature in O2:

CleanShot 2024-10-14 at 9  09 08@2x

The key difference is that both of those implementations would add any recently opened charts to the top of the list (regardless of whether they were recently searched or not). This implementation does not do that. Instead, it adds recently searched patients to the top of the list, but it does not add recently viewed charts. So, for example, if you clicked on patient in the Active Visits list and launched their chart, they would not show up at the top of the recently searched list. They would show up in the list if you went back and searched for them, but that's it. The reason for this is that in O2, every "chart" page calls EMR API's `ApplicationEvent#patientViewed()` service, which pushes the patient's ID into the list of recently viewed patients. The challenge is that none of this logic is exposed via the REST API. To get around this, we'd have to expose this mechanism by extending the REST API. In lieu of this, we'll have to make do with this approach for now.


packages/esm-patient-search-app/src/compact-patient-search-extension/index.tsx
```diff
@@ -21,48 +22,40 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
...
   const handlePatientSelection = useCallback(
-    (evt, index: number) => {
-      evt.preventDefault();
+    (event, index: number) => {
+      event.preventDefault();
       if (selectPatientAction) {
         selectPatientAction(patients[index].uuid);
       } else {
         navigate({
-          to: `${interpolateString(config.search.patientResultUrl, {
```
denniskigen, line 53: `interpolateString` returns a string, so there's no need for the additional interpolation.

packages/esm-patient-search-app/src/compact-patient-search-extension/index.tsx
```diff
@@ -21,48 +22,40 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
...
-  }, []);
-
-  const handleClear = useCallback(() => {
-    setSearchTerm('');
-  }, [setSearchTerm]);
+  const handleChange = useCallback((val) => setSearchTerm(val), [setSearchTerm]);
 
-  const handleReset = useCallback(() => {
-    setSearchTerm('');
-  }, [setSearchTerm]);
```
denniskigen, line 41: `handleReset` and `handleClear` do the same thing.

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-search.component.tsx
```diff
@@ -95,6 +118,24 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
     }
   }, [focusedResult, bannerContainerRef, handleFocusToInput]);
 
+  useEffect(() => {
+    if (fetchError) {
```
denniskigen, line 122: I've moved these here so that the resource file doesn't have to deal with any rendering concerns. A post-merge commit should change the resource file extension from `.tsx` to `.ts`.

packages/esm-patient-search-app/src/compact-patient-search/recently-searched-patients.component.tsx
```diff
@@ -74,9 +74,16 @@ const RecentlySearchedPatients = React.forwardRef<HTMLDivElement, RecentPatientS
...
+              <span className={styles.resultsTextCount}>
+                {t('recentSearchResultsCount', '{{count}} recent search result', {
+                  count: searchResults.length,
+                })}
+              </span>
+              {isValidating && (
+                <span className={styles.validationIcon}>
+                  <InlineLoading className={styles.spinner} />
+                </span>
+              )}
```
denniskigen, line 86: This will render a loading spinner when we're revalidating recently searched patients. Additionally, because we've specified the [keepPreviousData](https://swr.vercel.app/docs/advanced/understanding.en-US#return-previous-data-for-better-ux) option in the `useInfinitePatientSearch` hook, the previous data will be returned while the new data is being fetched.

packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -2,17 +2,18 @@ import { Type, validators } from '@openmrs/esm-framework';
 
 export const configSchema = {
   search: {
-    // TODO: Rename this config property
-    patientResultUrl: {
+    patientChartUrl: {
```
denniskigen, line 5: This config property key change is a **BREAKING change** for implementers' config schemas. It will be communicated appropriately in the release notes when the next release it cut.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -93,68 +138,81 @@ export function useRecentlyViewedPatients(showRecentlySearchedPatients: boolean
...
+  );
+
+  const shouldFetch = isSearching && patientUuids !== null && patientUuids.length > 0;
 
-  const results = useMemo(
+  const { data, isLoading, isValidating, setSize, error, size } = useSWRInfinite<FetchResponse<SearchedPatient>, Error>(
+    shouldFetch ? getPatientUrl : null,
+    openmrsFetch,
+    {
+      keepPreviousData: true,
```
denniskigen, line 198: [keepPreviousData docs](https://swr.vercel.app/docs/advanced/understanding.en-US#return-previous-data-for-better-ux)

donaldkibet (approved):
Thanks @denniskigen LGTM

packages/esm-patient-search-app/src/compact-patient-search-extension/index.tsx
```diff
@@ -86,7 +79,7 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
 
   return (
     <div className={styles.patientSearchBar}>
-      <form onSubmit={handleSubmit} className={styles.searchArea}>
+      <form onSubmit={(event) => event.preventDefault} className={styles.searchArea}>
```
ibacher, line 82:
```suggestion
      <form onSubmit={(event) => event.preventDefault()} className={styles.searchArea}>
```

ibacher (approved):
Impressive! Just one small thing that should be fixed up.

PR #1326, opened 2024-09-24
(fix) - O3-4015 - Ward App - patient should get unassigned from bed at old l…

chibongho:
…ocation upon transfer to new location

## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

This PR fixes the following scenario:
- Patient in Ward location A with bed X assigned
- Patient gets transferred to Ward location B without being assigned a bed.

The patient should have bed X unassigned.

PR #1349, opened 2024-10-21
(fix) O3-4015 - Ward App - unassign bed from patient when they are tr…

chibongho:
…ansferred

## Requirements

## Summary

When a patient with a bed in location A is transferred to a new location B, their current bed should be unassigned. Previous [PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/1326/files) did not correctly fix this issue. It was trying to look for the assigned bed in location B instead of A. This PR fixes it by adding and using a new hook to get the bed assigned to the patient.

Adjusted the unit tests to reflect this change.


mseaton (approved):
Looks good to me, though for the future we may want to look into whether we should be handling this on the back-end transactionally vs. on the front end as a separate request...

PR #1351, opened 2024-10-23
(fix) Fix React missing key warning for QueueTable table filters

denniskigen:
## Requirements

## Summary

This PR fixes a missing React key warning for the tableFilter prop in QueueTable. `tableFilter` is an array of React nodes that get rendered as children of the `TableToolbarContent` component. I've renamed the prop to `tableFilters` to better describe that it's an array of filters. Second, I've updated the rendering logic to use the `map` function to iterate over the `tableFilters` array and provide a unique `key` prop to each child. I'm using a `React.Fragment` to wrap each child in the array to avoid unnecessary additional DOM elements. Each of these fragments uses the index of the array as the key. This is not a perfect solution in the sense that keys should be stable and unique identifiers for elements in the array. However, in this case, the array contents are dynamic and may change in size, so a static index identifier works for now. I've also added keys to the `QueueTableByStatusMenu` and `QueueTableMetricsCard` components.

I've also fixed the spacing of the filters in the `QueueTable` component as well as removed some unnecessary space from the filter label text.


packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -133,7 +133,7 @@ function DefaultQueueTable() {
               queueUuid={null}
               statusUuid={null}
               ExpandedRow={QueueTableExpandedRow}
-              tableFilter={[
+              tableFilters={[
```
chibongho, line 136: I've encountered a similar scenario in the ward app and decided that it's better to just type `tableFilters` as `ReactNode` (instead of `ReactNode[]`) and then do 

```
<>
  <QueueDropdownFilter />
  <StatusDropdownFilter />
  ...
</>
```

Then we don't have do deal with keys.

chibongho (approved):
Thanks for this!

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
denniskigen: Should we switch to that here instead?

chibongho: sure

chibongho (approved):
Thanks!

PR #1357, opened 2024-11-01
(fix) O3-3996 - Ward App - adjust pagination size of admilocation / t…

chibongho:
…ransfer location selector

## Requirements

## Summary

Increase the pagination page size of the admission / transfer location picker. I'm using 15 as the page size default because that's also the number UX designers quoted for the max number of beds we display using radio buttons instead of dropdown in the bed selector UI. It's roughly the number of radio button options that can be fit into one screen.

## Screenshots

image


mogoodrich (approved):
Approved, *if* the idea is that we still do want pagination if there are more than 15 items (and I'm understanding correctly)?

chibongho:
> Approved, _if_ the idea is that we still do want pagination if there are more than 15 items (and I'm understanding correctly)?

Yeah, I checked with @fanderson on that.

PR #1376, opened 2024-11-15
(fix) O3-4116 use correct colours for service queue priority

flosell:
## Requirements

## Summary

_(First PR here so please let me know if I'm doing something wrong)_

O3-4116 notes that priority tags in the service queue view don't have any color. 

Implementers were already able to configure this, however this was probably not used frequently. So I'm guessing the intended behaviour is a default configuration that can be overwritten if necessary. This is what this PR adds. 

The PR also adds support for the orange color intended for the 'urgent' priority that was still missing. 

## Screenshots
Screenshot 2024-11-15 at 14 59 11

## Related Issue

https://openmrs.atlassian.net/browse/O3-4116

## Other

I'm not very happy the way the orange color is handled: Seems like the carbon design system does not have a color 'orange' so I implemented it as a special case similar to how the [tags for the header in the patient-chart are implemented](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/visit-header/visit-header.component.tsx#L108). I'm also hardcoding the [same color](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/visit-header/visit-header.scss#L97). This seems to be slightly smelly. I'm guessing this component in this color is used in a few different places so there should be a generic solution for this problem but I'm not sure how to approach it... Suggestions welcome! 

Also, the issue also asks for fixing capitalization for 'Not Urgent' - I haven't looked into this yet so might be another PR

### DIFF
--- a/packages/esm-service-queues-app/src/config-schema.ts
+++ b/packages/esm-service-queues-app/src/config-schema.ts
@@ -24,7 +24,7 @@ const statusIcons = ['Group', 'InProgress'] as const;
 type StatusIcon = (typeof statusIcons)[number];
 
 // Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview plus orange for priority tags
-const carbonTagColors = [
+const priorityTagColors = [
   'red',
   'magenta',
   'purple',
@@ -39,7 +39,7 @@ const carbonTagColors = [
   'high-contrast',
   'outline',
 ] as const;
-type CarbonTagColor = (typeof carbonTagColors)[number];
+type PriorityTagColor = (typeof priorityTagColors)[number];
 
 const tagStyles = ['bold'] as const;
 type TagStyle = (typeof tagStyles)[number];
@@ -282,7 +282,7 @@ export const configSchema = {
                 _type: Type.String,
                 _description:
                   'The color of the tag. This is based on the "type" field of the Carbon Design System "Tag" component.',
-                _validators: [validators.oneOf(carbonTagColors)],
+                _validators: [validators.oneOf(priorityTagColors)],
                 _default: 'gray',
               },
               style: {
@@ -467,7 +467,7 @@ export interface PatientIdentifierColumnConfig {
 }
 export interface PriorityConfig {
   conceptUuid: string;
-  color: CarbonTagColor;
+  color: PriorityTagColor;
   style: TagStyle;
 }
 
--- a/packages/esm-service-queues-app/src/config-schema.ts
+++ b/packages/esm-service-queues-app/src/config-schema.ts
@@ -23,7 +23,7 @@ type ColumnType = (typeof columnTypes)[number];
 const statusIcons = ['Group', 'InProgress'] as const;
 type StatusIcon = (typeof statusIcons)[number];
 
-// Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview
+// Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview plus orange for priority tags
 const carbonTagColors = [
   'red',
   'magenta',
@@ -32,6 +32,7 @@ const carbonTagColors = [
   'teal',
   'cyan',
   'gray',
+  'orange',
   'green',
   'warm-gray',
   'cool-gray',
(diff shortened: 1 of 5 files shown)


chibongho:
> I'm guessing this component in this color is used in a few different places so there should be a generic solution for this problem but I'm not sure how to approach it... Suggestions welcome!

Yeah, the patient chart ought to be using the same component as the services queues app to render those priority tags. It's usually done through extensions (https://o3-docs.openmrs.org/docs/extension-system). I suggest having the PriorityTag be its own component so it's closer to what we want.

> Also, the issue also asks for fixing capitalization for 'Not Urgent' - I haven't looked into this yet so might be another PR
I *think* the string is defined here (I don't have access to change it). https://app.openconceptlab.org/#/orgs/openmrs/sources/DemoQueueConcepts/concepts/DemoQueue-Status/

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -23,7 +23,7 @@ type ColumnType = (typeof columnTypes)[number];
 const statusIcons = ['Group', 'InProgress'] as const;
 type StatusIcon = (typeof statusIcons)[number];
 
-// Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview
+// Options from https://react.carbondesignsystem.com/?path=/docs/components-tag--overview plus orange for priority tags
 const carbonTagColors = [
```
chibongho, line 27: Maybe rename this to `priorityTagColors` since it's no longer a list of carbon tag colors?

flosell:
> Yeah, the patient chart ought to be using the same component as the services queues app to render those priority tags. It's usually done through extensions (https://o3-docs.openmrs.org/docs/extension-system). I suggest having the PriorityTag be its own component so it's closer to what we want.

Hi @chibongho - thanks for the input! I thought the extension system is more for cases where we might want to swap out or swap in various extensions into a slot. 
In this case, it's a fairly well-defined component that would be plugged into the services queues and the patient chart. So might this be more a case similar to [patient-photo](https://github.com/openmrs/openmrs-esm-core/blob/841f0d6b997fec1e0e639aa04a913322906408f8/packages/framework/esm-styleguide/src/patient-photo/patient-photo.component.tsx), a simple react component that can be used in multiple places? 

I also looked at the patient chart implementation a bit more, seems there's also a bit different behavior, e.g. what's 'Urgent' on the service queue is 'Priority' in the patient chart. 
All in all, sounds like a bit of a bigger change - should that refactoring (and aligning the behavior) be tackled in a separate issue?

chibongho:
> I thought the extension system is more for cases where we might want to swap out or swap in various extensions into a slot.

Yeah, you're right. (In fact, the extension system has often been misused for other purposes.) I would not be opposed to putting the priority tag in a common library. I guess one reason for using an extension is that it is possible for certain implementation to have the patient chart but not service queues functionality (although the current patient chart code isn't respecting that either.)

As for which common library to put it in, `esm-styleguide` historically has a fairly [high bar](https://openmrs.atlassian.net/browse/O3-1422) for admitting components. `esm-patient-common-lib` was originally meant only for packages within `openmrs-esm-patient-chart`, but is now depended on by `openmrs-esm-patient-management` as well. That seems like a good place to put it. What do you think @denniskigen @ibacher ?

> I also looked at the patient chart implementation a bit more, seems there's also a bit different behavior, e.g. what's 'Urgent' on the service queue is 'Priority' in the patient chart. All in all, sounds like a bit of a bigger change - should that refactoring (and aligning the behavior) be tackled in a separate issue?

TBH, the behavior in patient chart is probably wrong. But yeah, let's not worry about that for now. I was suggesting that we have the priority tag be its own component, mostly to avoid the big `<Tag>` component being there twice.

ibacher:
I’d prefer to do things in the styleguide than in the patient-common-lib. As much as possible, we want to avoid adding things there that are used outside the patient chart (and we’ve been trying to remove the stuff that makes things like patient management depend on it).

It’s not clear to me what this would add as a generic component over and above the Carbon Tag element (in particular, things that need configuration are a poor fit for the styleguide or the patient common lib).

flosell:
@ibacher - I guess the main addition on top of the Carbon Tag element would be the orange color for Priority / Urgent. 
But I sense maybe it's not enough to warrant a larger refactoring...

@chibongho I've pulled the tag into a variable to avoid the duplication - if we pulled it out into its own component there isn't really much left in QueuePriority so I kept the two things together

packages/esm-service-queues-app/src/queue-entry-table-components/queue-priority.scss
```diff
@@ -9,3 +9,7 @@
 .bold {
   font-weight: bold;
 }
+
+.orange {
+  background-color: #ffc9a3;
```
ibacher, line 14: So, in general, we want to prefer Carbon's color tokens wherever possible rather than hard-coded hex values. In this case, maybe `layout.$orange-20`, which is not identical, but pretty close.

ibacher:
> fixing capitalization

The capitalization here is a non-code issue, so I wouldn't worry about it.

packages/esm-service-queues-app/src/queue-entry-table-components/queue-priority.scss
flosell: fixed

ibacher (approved):

PR #1381, opened 2024-11-19
(refactor) O3-4199: Bed management should use the modal system

amosmachora:
## Requirements

## Summary
This PR refactors the current modal functionality in the bed management app to use the openmrs modal system as described [here](https://o3-docs.openmrs.org/docs/modal-system). When I was working on this PR I was careful only to refactor the modals, separate edit-add functionalities _(beds, bed tags bed types)_, and not change any underlying implementation. Any change made can be considered absolutely necessary.

I also improved the UX when submitting the modals by adding loading states on the button and showing a spinner. Nothing major.

### Short video verifying that nothing breaks after this change
https://github.com/user-attachments/assets/5e25e243-f2d7-4f17-915d-0b786401b138

## Related Issue

https://openmrs.atlassian.net/browse/O3-4199
## Other

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1209837472253519

### DIFF
--- a/packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
+++ b/packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
@@ -37,8 +37,7 @@ interface BedAdministrationFormProps {
   headerTitle: string;
   initialData: BedWithLocation;
   occupancyStatuses: string[];
-  onModalChange: (showModal: boolean) => void;
-  showModal: boolean;
+  closeModal: () => void;
 }
 
 interface ErrorType {
@@ -86,8 +85,7 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
   headerTitle,
   initialData,
   occupancyStatuses,
-  onModalChange,
-  showModal,
+  closeModal,
 }) => {
   const { t } = useTranslation();
   const [occupancyStatus, setOccupancyStatus] = useState(capitalize(initialData.status));
@@ -129,9 +127,8 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
   };
 
   return (
-    // TODO: Port this over to the modal system or create individual modals for each form
-    <ComposedModal open={showModal} onClose={() => onModalChange(false)} preventCloseOnClickOutside>
-      <ModalHeader title={headerTitle} />
+    <React.Fragment>
+      <ModalHeader title={headerTitle} closeModal={closeModal} />
       <ModalBody hasScrollingContent>
         <Form>
           <Stack gap={3}>
@@ -283,14 +280,14 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
         </Form>
       </ModalBody>
       <ModalFooter>
-        <Button onClick={() => onModalChange(false)} kind="secondary">
+        <Button onClick={closeModal} kind="secondary">
           {getCoreTranslation('cancel', 'Cancel')}
         </Button>
         <Button disabled={!isDirty} onClick={handleSubmit(onSubmit, onError)}>
           <span>{t('save', 'Save')}</span>
         </Button>
       </ModalFooter>
-    </ComposedModal>
+    </React.Fragment>
   );
 };
 
--- a/packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
+++ b/packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
@@ -6,7 +6,6 @@ import { zodResolver } from '@hookform/resolvers/zod';
 import {
   Button,
   ComboBox,
-  ComposedModal,
   Form,
   FormGroup,
   InlineNotification,
@@ -196,17 +195,7 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
                     items={allLocations}
                     itemToString={(location) => location?.display ?? ''}
                     label={t('location', 'Location')}
-                    /*
-                      TODO: onBlur shall be refactored to onBlur={onBlur} if esm-core has @carbon/react version 1.72+
-                      (ComboBox bug does not trigger onChange below mentioned version in production build - see https://github.com/carbon-design-system/carbon/issues/18145#issuecomment-2521936772)
-                    */
-                    onBlur={(event) => {
-                      const selectedLocation = allLocations.find((element) => element.display === event.target.value);
-                      if (selectedLocation)
-                        setValue('location', { display: selectedLocation.display, uuid: selectedLocation.uuid });
-                      else setValue('location', { display: '', uuid: '' });
-                      onBlur();
-                    }}
+                    onBlur={onBlur}
                     onChange={({ selectedItem }) => onChange(selectedItem)}
                     placeholder={t('selectBedLocation', 'Select a bed location')}
                     ref={ref}
--- a/packages/esm-bed-management-app/src/bed-administration/bed-administration-table.component.tsx
+++ b/packages/esm-bed-management-app/src/bed-administration/bed-administration-table.component.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   Button,
@@ -18,13 +18,17 @@ import {
   Tile,
 } from '@carbon/react';
 import { Add, Edit } from '@carbon/react/icons';
-import { ErrorState, isDesktop as desktopLayout, useLayoutType, usePagination } from '@openmrs/esm-framework';
-import type { BedWithLocation } from '../types';
+import {
+  ErrorState,
+  isDesktop as desktopLayout,
+  showModal,
+  useLayoutType,
+  usePagination,
+} from '@openmrs/esm-framework';
 import { useBedsGroupedByLocation } from '../summary/summary.resource';
 import CardHeader from '../card-header/card-header.component';
-import EditBedForm from './edit-bed-form.component';
 import Header from '../header/header.component';
-import NewBedForm from './new-bed-form.component';
+import type { BedWithLocation } from '../types';
 import styles from './bed-administration-table.scss';
 
 const BedAdministrationTable: React.FC = () => {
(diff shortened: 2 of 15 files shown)


packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  const onSubmit = (formData: BedTagData) => {
+    const result = BedTagAdministrationSchema.safeParse(formData);
+    if (result.success) {
+      setShowErrorNotification(false);
+      handleUpdateBedTag(formData);
+    }
+  };
+
+  const onError = (error: { [key: string]: ErrorType }) => {
+    setFormStateError(Object.entries(error)[0][1].message);
```
ibacher, line 128: So we only support a single error? That seems wrong...

In principle, each field should display an error message related to that field and we only need a bigger message if there's an error... bigger than a field.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+        <Form>
+          <Stack gap={3}>
+            <FormGroup legendText={''}>
+              <Controller
+                name="name"
+                control={control}
+                render={({ field, fieldState }) => (
+                  <>
+                    <TextInput
+                      id="bedTag"
```
ibacher, line 145: Is `id` required here?

ibacher (changes requested):
A couple of quick thoughts. While its a good idea to use the new modal system, I don't like  the idea of having to maintain 6 forms instead of the previous 3.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+const BedTagAdministrationSchema = z.object({
+  name: z.string().max(255),
+});
+
+interface ErrorType {
+  message: string;
+}
+
+const EditBedTagForm: React.FC<EditBedTagFormProps> = ({ editData, mutate, closeModal }) => {
```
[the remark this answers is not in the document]
ojwanganto: Ideally, we should maintain a single for adding/editing a tag.

amosmachora:
> A couple of quick thoughts. While its a good idea to use the new modal system, I don't like the idea of having to maintain 6 forms instead of the previous 3.

Hello @denniskigen is at a better position to answer this question. Also the new 'architecture' is a personal preference of mine too as it allows for separation of concerns. I think it looks like more work from the onset but it simplifies work later IMO.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
amosmachora: @ibacher the aim of this PR was to just separate out the forms. As I said in the PR description I consciously made the decision to leave everything as is to make it easier for review. If more changes like this are required I am more than happy to work on them but that might bloat the PR and make it more challenging to review.

Also as I was working on this I made the assumption that since I was working on top of already merged changes there is not much need to concentrate on the details.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
```diff
@@ -0,0 +1,191 @@
...
+  } = useForm<BedTagData>({
+    mode: 'all',
+    resolver: zodResolver(BedTagAdministrationSchema),
+    defaultValues: {
+      name: editData.name || '',
+    },
+  });
+
+  const onSubmit = (formData: BedTagData) => {
+    const result = BedTagAdministrationSchema.safeParse(formData);
```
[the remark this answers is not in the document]
amosmachora: The original code was written by @brandones in my understanding so I feel he is at a better position to answer this.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
ibacher: Because of the way these changes are done, it's very hard for me to tell what's "new" in this PR and what's just "moved" from somewhere else.

ibacher:
> I think it looks like more work from the onset but it simplifies work later IMO.

See, I see this in exactly the opposite way. Previously, we had one form to add or edit each of the three things (bed tags, etc.). Now, we have a separate form for each. That means if we decide that we need to, I don't know, collect a date the change should be effective on, we now need to add that field to both the add form and the edit form. With the previous setup, we just needed to modify the code in a single place.

While, in general, I'm in favour of breaking things up a bit, breaking things up in _this_ way makes on-going maintenance harder, precisely because you need to know that there are two files that need to be updated for each change.

The objection here isn't _necessarily_ to having a separate `add-tag-form` modal and `edit-tag-form` modal, but these should _share_ a single component that actually renders the form.

amosmachora:
> > I think it looks like more work from the onset but it simplifies work later IMO.
> 
> See, I see this in exactly the opposite way. Previously, we had one form to add or edit each of the three things (bed tags, etc.). Now, we have a separate form for each. That means if we decide that we need to, I don't know, collect a date the change should be effective on, we now need to add that field to both the add form and the edit form. With the previous setup, we just needed to modify the code in a single place.
> 
> While, in general, I'm in favour of breaking things up a bit, breaking things up in _this_ way makes on-going maintenance harder, precisely because you need to know that there are two files that need to be updated for each change.
> 
> The objection here isn't _necessarily_ to having a separate `add-tag-form` modal and `edit-tag-form` modal, but these should _share_ a single component that actually renders the form.

Great points @ibacher this absolutely makes sense if you explain it that way. I guess its just a personal preference of mine and the ticket did leave room for some creativity.  What do you suggest we do in that case then? should we just refactor the modals to use the modal system and leave everything as is?

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
amosmachora: I see. Think of it this way. the `new tag form` and `the edit tag form` both used an abstracted `tag component`. what I did was duplicate the tag component in both forms. So the `new tag form` know owns its own functionality that was previously in the `tag component` same as the edit one.

amosmachora: so nothing is new in short

amosmachora:
> > I think it looks like more work from the onset but it simplifies work later IMO.
> 
> See, I see this in exactly the opposite way. Previously, we had one form to add or edit each of the three things (bed tags, etc.). Now, we have a separate form for each. That means if we decide that we need to, I don't know, collect a date the change should be effective on, we now need to add that field to both the add form and the edit form. With the previous setup, we just needed to modify the code in a single place.
> 
> While, in general, I'm in favour of breaking things up a bit, breaking things up in _this_ way makes on-going maintenance harder, precisely because you need to know that there are two files that need to be updated for each change.
> 
> The objection here isn't _necessarily_ to having a separate `add-tag-form` modal and `edit-tag-form` modal, but these should _share_ a single component that actually renders the form.

While we are on the topic, just for fun, how about when you only need to refactor one functionality, say, the edit functionality? Isn`t there a chance that you might end up breaking two features instead of one or making an unnecessary change in the other while you only aimed at changing one thing?

ojwanganto:
> > > I think it looks like more work from the onset but it simplifies work later IMO.
> > 
> > 
> > See, I see this in exactly the opposite way. Previously, we had one form to add or edit each of the three things (bed tags, etc.). Now, we have a separate form for each. That means if we decide that we need to, I don't know, collect a date the change should be effective on, we now need to add that field to both the add form and the edit form. With the previous setup, we just needed to modify the code in a single place.
> > While, in general, I'm in favour of breaking things up a bit, breaking things up in _this_ way makes on-going maintenance harder, precisely because you need to know that there are two files that need to be updated for each change.
> > The objection here isn't _necessarily_ to having a separate `add-tag-form` modal and `edit-tag-form` modal, but these should _share_ a single component that actually renders the form.
> 
> Great points @ibacher this absolutely makes sense if you explain it that way. I guess its just a personal preference of mine and the ticket did leave room for some creativity. What do you suggest we do in that case then? should we just refactor the modals to use the modal system and leave everything as is?

In addition to @ibacher's response, it is also a case of convention and standards which also guide code reviews. If you look at many forms in the system, the convention is to use a single form for add/edit and unit tests ensure the forms work as expected. I suggest that you only refactor the original code to use the modals. That will also align with the PR title for ease of reviews

amosmachora:
> > > > I think it looks like more work from the onset but it simplifies work later IMO.
> > > 
> > > 
> > > See, I see this in exactly the opposite way. Previously, we had one form to add or edit each of the three things (bed tags, etc.). Now, we have a separate form for each. That means if we decide that we need to, I don't know, collect a date the change should be effective on, we now need to add that field to both the add form and the edit form. With the previous setup, we just needed to modify the code in a single place.
> > > While, in general, I'm in favour of breaking things up a bit, breaking things up in _this_ way makes on-going maintenance harder, precisely because you need to know that there are two files that need to be updated for each change.
> > > The objection here isn't _necessarily_ to having a separate `add-tag-form` modal and `edit-tag-form` modal, but these should _share_ a single component that actually renders the form.
> > 
> > 
> > Great points @ibacher this absolutely makes sense if you explain it that way. I guess its just a personal preference of mine and the ticket did leave room for some creativity. What do you suggest we do in that case then? should we just refactor the modals to use the modal system and leave everything as is?
> 
> In addition to @ibacher's response, it is also a case of convention and standards which also guide code reviews. If you look at many forms in the system, the convention is to use a single form for add/edit and unit tests ensure the forms work as expected. I suggest that you only refactor the original code to use the modals. That will also align with the PR title for ease of reviews

Perfect. Makes sense if that is the ongoing standard then there is probably no reason for this one to be different.

packages/esm-bed-management-app/src/bed-tag/edit-tag-form.modal.tsx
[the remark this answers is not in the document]
brandones: I don't think I wrote this; if I appear in the blame I probably moved the code in from somewhere else. Anyway it looks like this is outdated, so I will assume the question is resolved.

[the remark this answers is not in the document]
ojwanganto: @amosmachora please take note

NethmiRodrigo:
Thanks @amosmachora!

NethmiRodrigo (approved):

ibacher (approved):

PR #1386, opened 2024-11-25
(fix) O3-4218: Auto generated identifiers with manual entry are not saved

ynurmahomed:
## Requirements

## Summary
Auto-generated identifiers with manual entry are not being saved.
It looks like the initial value is undefined, which causes them to be filtered out before saving the registration form.

## Related Issue
https://openmrs.atlassian.net/browse/O3-4218


ynurmahomed:
I would like to implement an e2e test for this feature, if someone could point me to how to add the metadata required.

ibacher (changes requested):
I don't really understand the use-case for this. I think it's quite correct that we are filtering out the "manual entry" field _unless the user puts text in the field_. If you can demonstrate (via a test) that the user _cannot enter a value and have that saved_ that's a bug. This, i.e., assuming that if the user did not enter a value and did not set it to be autogenerated, that we want to save a blank identifier for a patient, strikes me as buggy behaviour.

ynurmahomed:
Hi @ibacher. 

I'm not trying to save an empty identifier. When the value for this type of identifier is empty, it will be handled in form-manager:

https://github.com/openmrs/openmrs-esm-patient-management/blob/d75bb3a72b4205a22da4841e44f10ca55acb137d/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts#L244-L252

The use case is when you have an identifier that can be [both auto-generated and entered manually](https://dev3.openmrs.org/openmrs/module/idgen/editAutoGenerationOption.form?autoGenerationOption=2).

If you can point me to how to add the necessary metadata (identifier types) I can write an a2a test.
For now, this is the current behaviour in dev3:

https://github.com/user-attachments/assets/4c6f4a66-612b-46f7-8b36-ed84f6f8f16d

ynurmahomed:
@ibacher If I'm not mistaken the filtering happens inside [patient-registration.component](https://github.com/openmrs/openmrs-esm-patient-management/blob/de0c9c70c72353f078ba50bdcfe0aabe24622753/packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx#L74) before calling [savePatientIdentifiers](https://github.com/openmrs/openmrs-esm-patient-management/blob/de0c9c70c72353f078ba50bdcfe0aabe24622753/packages/esm-patient-registration-app/src/patient-registration/form-manager.ts#L218).

Should I fix the filtering logic there?

ibacher:
Yes, I think so. I think that will result in code that's clearer (and thus harder to regress).

ynurmahomed:
@ibacher done.

ibacher (approved):

ibacher:
Great work! And thanks!

PR #1394, opened 2024-12-02
(fix) O3-3806: Use OpenMRS Fetch All for service queue entries API call

kb019:
## Requirements

## Summary
To use hooks already present in esm-core to fetch all the queue entries.

[388370799-867ef32b-fb36-4a01-a49b-c5fbf7d3b0ff.webm](https://github.com/user-attachments/assets/69cac106-ddb8-4053-a2fa-fc4bf86f6eaa)

## Related Issue
https://openmrs.atlassian.net/browse/O3-3806


packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
```diff
@@ -27,11 +28,17 @@ function ClinicMetrics() {
   const [initialSelectedItem, setInitialSelectItem] = useState(() => {
     return !currentService?.serviceDisplay || !currentService?.serviceUuid;
   });
-  const { totalCount } = useQueueEntries({
-    service: currentService?.serviceUuid,
-    location: currentQueueLocation,
-    isEnded: false,
-  });
+
+  const { data } = useSWR<QueueEntryResponse>(
```
brandones, line 32: You are no longer using `useQueueEntries` at all. I can't imagine that this works in cases where the number of entries exceeds the page size. I did my development and testing of this feature by creating 11 service queue entries and setting the page size to 3. I recommend you do something similar. There are a lot of edge cases that we need to make sure will continue to work.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -69,104 +53,29 @@ export function useMutateQueueEntries() {
...
-  const { data: pageData, isValidating, error: pageError } = useSWR<QueueEntryResponse, Error>(pageUrl, openmrsFetch);
+    setPageUrl(getInitialUrl(rep, searchCriteria));
+  }, [searchCriteria, rep]);
 
   useEffect(() => {
-    const nextUrl = getNextUrlFromResponse(pageData);
-    const stillWaitingForMutate = waitingForMutate && !isValidating;
-    if (waitingForMutate && isValidating) {
-      setWaitingForMutate(false);
+    if (initialMount) {
```
brandones, line 69: First off, this `initialMount` is a ref, so it is always true. You want `initialMount.current`.

But moreover, if `pageUrl` changes, then `useOpenmrsFetchAll` will get called with the new value of `pageUrl` as a parameter. Why would you call `mutate` at the same time?

brandones (changes requested):
It's a great idea but your implementation doesn't make any sense (yet).

packages/esm-service-queues-app/src/patient-queue-metrics/clinic-metrics.component.tsx
brandones: @kb019 wrote me on Slack with the following explanation:

> Before this pr , useQueueEntries was used in two places , one to show metrics and other one in the queue Table.  The useSwrInfinite used in useOpenmrsFetchAll Hook does not seem to cache results and leads to calling the url twice for each pageIndex. So i thought that inorder to show metrics we dont need to wait until all results are fetched.We can just get the metadata from the first url and show that.So i am not using useQueueEntries in clinic metrics component and just relying on useSWR.

Which makes sense!

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -69,104 +53,20 @@ export function useMutateQueueEntries() {
...
-      setError(pageError);
-    }
-  }, [pageError]);
+    setPageUrl(getInitialUrl(rep, searchCriteria));
+  }, [searchCriteria, rep]);
 
-  const queueUpdateListener = useCallback(() => {
-    refetchAllData();
-  }, [refetchAllData]);
+ const queueUpdateListener = useCallback(() => {
```
brandones, line 67: I think we simplify this further. I think we can get rid of all the `queueUpdateListener` and `queue-entry-updated` stuff, and whatever function it is that triggers the `queue-entry-updated` event. We can simply expose the mutate function and use that.

brandones (changes requested):
Great work @kb019 , just one suggestion for further simplification

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -23,30 +23,31 @@ export function getInitialUrl(rep: string, searchCriteria?: QueueEntrySearchCrit
...
         searchParam.append(key, value?.toString());
       }
     }
   }
- 
+
   return `${queueEntryBaseUrl}?${searchParam.toString()}`;
 }
 
+let queueEntryMutates: ReturnType<typeof useQueueEntries>['mutate'][] = [];
```
brandones, line 38: Do we still need `useMutateQueueEntries` at all, or can we just have `useQueueEntries` return the `mutate` returned by `useOpenmrsFetchAll`?

brandones (changes requested):

jayasanka-sack:
We're going to close this PR due to inactivity and as part of our repository cleanup. Thanks for the contribution. If you'd like to pick this up again in the future, feel free to reopen or create a new PR. Happy to help. 🙌

PR #1397, opened 2024-12-04
(feat) HIE-9: Add MPI workflows to OpenMRS frontend

reagan-meant:
## Requirements

## Summary

https://openmrs.atlassian.net/browse/HIE-9
## Screenshots

https://github.com/user-attachments/assets/646b72b4-f97f-4633-95d6-88ccfacc485b

## Related Issue

See background conversation [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/1313)


packages/esm-patient-registration-app/src/patient-registration/patient-registration-hooks.ts
```diff
@@ -39,14 +40,20 @@ interface DeathInfoResults {
...
 }
-
-export function useInitialFormValues(patientUuid: string): [FormValues, Dispatch<FormValues>] {
-  const { freeTextFieldConceptUuid } = useConfig<RegistrationConfig>();
-  const { isLoading: isLoadingPatientToEdit, patient: patientToEdit } = usePatient(patientUuid);
-  const { data: deathInfo, isLoading: isLoadingDeathInfo } = useInitialPersonDeathInfo(patientUuid);
-  const { data: attributes, isLoading: isLoadingAttributes } = useInitialPersonAttributes(patientUuid);
-  const { data: identifiers, isLoading: isLoadingIdentifiers } = useInitialPatientIdentifiers(patientUuid);
-  const { data: relationships, isLoading: isLoadingRelationships } = useInitialPatientRelationships(patientUuid);
+import { useMpiPatient } from './mpi/mpi-patient.resource';
```
ibacher, line 43: Can we move this import to be with the others?

packages/esm-patient-registration-app/src/patient-registration/patient-registration-hooks.ts
```diff
@@ -108,7 +115,42 @@ export function useInitialFormValues(patientUuid: string): [FormValues, Dispatch
...
+    isLoadingPatientToEdit,
+    patientToEdit,
+    patientUuid,
+    fieldConfigurations.phone.personAttributeUuid,
+    isLocal,
+  ]);
+
+  useEffect(() => {
+    const fetchValues = async () => {
+      //check that patient is not dead
```
ibacher, line 129: ???

packages/esm-patient-registration-app/src/patient-registration/patient-registration-hooks.ts
```diff
@@ -108,7 +115,42 @@ export function useInitialFormValues(patientUuid: string): [FormValues, Dispatch
...
+  }, [
+    initialFormValues,
+    isLoadingPatientToEdit,
+    patientToEdit,
+    patientUuid,
+    fieldConfigurations.phone.personAttributeUuid,
+    isLocal,
+  ]);
+
+  useEffect(() => {
```
ibacher, line 127: This pattern feels off. This feels like it should be split into it's own hook or something an ultimately `useSWR()` if at all possible.

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
-export function getPhonePersonAttributeValueFromFhirPatient(patient: fhir.Patient) {
+export async function getIdentifierFieldValuesFromFhirPatient(
+  patient: fhir.Patient,
+  identifierConfig,
+): Promise<{ [identifierFieldName: string]: PatientIdentifierValue }> {
+  const identifiers: FormValues['identifiers'] = {};
+
+  for (const identifier of patient.identifier) {
+    for (const config of identifierConfig) {
+      const identifierConfig = config.identifierTypeSystem === identifier.system ? config : null;
```
ibacher, line 203: This feels much cleaner:

```suggestion
      if (config.identifierTypeSystem !== identifier.system) {
         continue;
      }
```

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+
+      if (identifierConfig) {
+        let identifierTypeName;
+
+        const url = `${restBaseUrl}/patientidentifiertype/${identifierConfig.identifierTypeUuid}`;
+        await openmrsFetch(url).then((response) => {
+          if (response.status == 200 && response.data) {
+            identifierTypeName = response.data.name;
+          }
```
ibacher, line 212: There's no need to check status. When checking status from fetch, it's usually better to prefer `ok`, but `openmrsFetch()` already does this for you and rejects if the promise rejects.

```suggestion
          if (response.data) {
            identifierTypeName = response.data.name;
          }
```

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+        });
```
ibacher, line 213: We should probably have a `catch()` hook here to log the error.

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+
+  for (const identifier of patient.identifier) {
+    for (const config of identifierConfig) {
+      const identifierConfig = config.identifierTypeSystem === identifier.system ? config : null;
+
+      if (identifierConfig) {
+        let identifierTypeName;
+
+        const url = `${restBaseUrl}/patientidentifiertype/${identifierConfig.identifierTypeUuid}`;
+        await openmrsFetch(url).then((response) => {
```
ibacher, line 209: `await` here is a bad idea. It makes looping through each identifier make one request to the backend and then wait until processing the next one. It would be better to do something like:

```ts
// before for loop
const promises = [];

// around here
promises.push(openmrsFetch(url).then(response => {
   if (!response.data?.name) {
      return;
   }
   
   identifiers[response.data.name] = {
      // blah, blah
   }
}));

File: packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts [hunk 8]
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+          if (response.status == 200 && response.data) {
+            identifierTypeName = response.data.name;
+          }
+        });
+
+        identifiers[identifierTypeName] = {
+          identifierUuid: null,
+          preferred: false, // consider identifier.use === 'official' ?? by default autogen is preferred
```
ibacher, line 217: I'm of the opinion that no MPI identifiers should be marked as preferred, regardless of use.

ibacher: The default preferred identifier is part of the identifier configuration; it has nothing to do with autogeneration.

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,10 +192,48 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+  }
+
+  return identifiers;
+}
+
+export function getPhonePersonAttributeValueFromFhirPatient(patient: fhir.Patient, phoneUuid) {
   const result = {};
   if (patient.telecom) {
-    result['phone'] = patient.telecom[0].value;
+    result[phoneUuid] = patient.telecom[0].value;
```
ibacher, line 236: Since we're fixing this, we should find the _first_ telecom value with `system === 'phone'`. That's a better implementation than this.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -39,10 +39,14 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
   const config = useConfig() as RegistrationConfig;
   const [target, setTarget] = useState<undefined | string>();
   const { patientUuid: uuidOfPatientToEdit } = useParams();
+  const sourcePatientId = new URLSearchParams(search).get('sourceRecord');
```
ibacher, line 42: I don't think this kind of data should be passed through the URL. It's a SPA, we can pass it through some kind of state variable (including the AppContext or something if necessary).

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
+import { capitalize } from 'lodash-es';
+import { type SearchedPatient } from '../types';
+import { getCoreTranslation } from '@openmrs/esm-framework';
+export function inferModeFromSearchParams(searchParams: URLSearchParams) {
+  return searchParams.get('mode')?.toLowerCase() === 'external' ? 'external' : 'internal';
+}
+
+export function mapToOpenMRSPatient(fhirPatients: Array<any>): Array<SearchedPatient> {
```
ibacher, line 8: This should have a stronger type than `any`. We are definitely assuming that certain properties exist and that should be reflected in the type here.

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+      uuid: null,
+      identifiers: null,
+      person: {
+        addresses: fhirPatient?.address?.map((address) => ({
+          cityVillage: address.city,
+          stateProvince: address.state,
+          country: address.country,
+          postalCode: address.postalCode,
+          preferred: false,
+          address1: '',
```
ibacher, line 28: We should at least map `address.line[0]` here.

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+        })),
+        age: null,
```
ibacher, line 30: We have a birthdate, so we should be able to calculate the age. There's a helper for that in the framework.

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+        birthdate: fhirPatient.birthDate,
+        gender: getCoreTranslation(fhirPatient.gender),
+        dead: fhirPatient.deceasedBoolean,
+        deathDate: fhirPatient.deceasedDateTime,
```
ibacher, line 34: In FHIR, only one of these two properties will be set at any given time, so we need to be responsive to that. If `deceasedBoolean` is set, `deceasedDateTime` will be `null` and vice-versa. At the very least, if `deceasedDateTime` is non-null we can infer that the patient is deceased if the datetime is now or in the past.

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
...
+        personName: {
+          display: fhirPatient.name[0].text
+            ? fhirPatient.name[0].text
+            : `${fhirPatient.name[0].family} ${fhirPatient.name[0].given[0]}`,
```
ibacher, line 38: The name format here probably needs to be templateable.

```suggestion
          display: fhirPatient.name[0].text ??
            `${fhirPatient.name[0].given[0]} ${fhirPatient.name[0].family}`,
```

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,49 @@
+import { capitalize } from 'lodash-es';
+import { type SearchedPatient } from '../types';
+import { getCoreTranslation } from '@openmrs/esm-framework';
+export function inferModeFromSearchParams(searchParams: URLSearchParams) {
+  return searchParams.get('mode')?.toLowerCase() === 'external' ? 'external' : 'internal';
```
ibacher, line 5: These should probably be `mpi` and just no value, which can be interpretted as `local` or whatever.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -43,6 +44,7 @@ const patientSearchCustomRepresentation = `custom:(${patientProperties.join(',')
  */
 export function useInfinitePatientSearch(
   searchQuery: string,
+  searchMode: string,
```
ibacher, line 47:
```suggestion
  searchMode: 'mpi' | null | undefined,
```

ibacher (changes requested):
Thanks @reagan-meant! A few comments on this overall.

packages/esm-patient-search-app/src/mpi/utils.ts
reagan-meant: @ibacher this [helper](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-utils/src/age-helpers.ts#L17) does not return a number.
Incompatible with our patient interface

ibacher: I see... Can we omit it?

packages/esm-patient-registration-app/src/patient-registration/patient-registration.test.tsx
```diff
@@ -407,6 +517,186 @@ describe('Registering a new patient', () => {
   });
 });
 
+describe('Import an MPI patient record', () => {
```
reagan-meant, line 520: @ibacher Unfortunately, this test isn't providing much value at the moment. The issue seems to be that I'm unable to update the initially mocked values that are supposed to be updated via the useEffect [hook](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397/files#diff-25dfbd3829e19bcd1e08cd0d0e2a572e26c7ace2f93409574bd182379c947497R70).

samuelmale (commented):
Nice Regan! I've left a few comments.

packages/esm-patient-search-app/src/patient-search-page/patient-search-views.component.tsx
```diff
@@ -1,18 +1,23 @@
...
 interface CommonProps {
   inTabletOrOverlay: boolean;
+  searchMode?: 'mpi' | null | undefined;
+  searchTerm: string;
 }
 
 interface PatientSearchResultsProps {
   searchResults: SearchedPatient[];
+  searchTerm: string;
+  searchMode?: 'mpi' | null | undefined;
```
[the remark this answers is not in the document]
reagan-meant: @ibacher advised we go this route. See [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397#discussion_r1884259812)
The default is local

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
samuelmale: @ibacher While we could use the [AppContext API](https://github.com/openmrs/openmrs-esm-core/blob/a61deb8b00d527f6d7386d096385bf5d1737a87f/packages/framework/esm-context/src/context.ts#L29), wouldn’t a query parameter be a reasonable fit for passing a patient ID? Or is the concern about exposing identifiers, particularly those from external registries? If that’s not the case, creating a context for this might feel overkill. What do you think?

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -53,6 +61,13 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
   const fieldDefinition = config?.fieldDefinitions?.filter((def) => def.type === 'address');
   const validationSchema = getValidationSchema(config);
 
+  useEffect(() => {
+    if (initialMPIFormValues) {
+      setInitialFormValues(initialMPIFormValues);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [initialMPIFormValues, setInitialMPIFormValues]);
```
samuelmale, line 69: Is `setInitialMPIFormValues` a necessary dependency?

packages/esm-patient-registration-app/src/patient-registration/patient-registration.test.tsx
samuelmale: Is there a plan to make the test suite more useful?

packages/esm-patient-search-app/src/mpi/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+export function inferModeFromSearchParams(searchParams: URLSearchParams): 'mpi' | null {
+  return searchParams.get('mode')?.toLowerCase() === 'mpi' ? 'mpi' : null;
+}
+
+export function mapToOpenMRSPatient(fhirPatients: fhir.Bundle, nameTemplate: string): Array<SearchedPatient> {
+  if (fhirPatients.total < 1) {
+    return [];
+  }
+  //Consider patient // https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-api/src/types/patient-resource.ts
+  const pts: Array<SearchedPatient> = [];
```
samuelmale, line 13:
```suggestion
  const omrsPatients: Array<SearchedPatient> = [];
```

samuelmale (commented):
@reagan-meant, any idea why the E2E Tests job is failing?

reagan-meant:
@samuelmale using the appcontext was failing [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/1397#discussion_r1913245361) and so either way we need to be using this for the time being

packages/esm-patient-registration-app/src/patient-registration/patient-registration-hooks.ts
```diff
@@ -180,6 +188,64 @@ export function useInitialFormValues(patientUuid: string): [FormValues, Dispatch
...
+          attributes: getPhonePersonAttributeValueFromFhirPatient(
+            mpiPatient.data,
+            fieldConfigurations.phone.personAttributeUuid,
+          ),
+        };
+        setInitialMPIFormValues(values);
+      }
+    })();
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
ibacher: Normally in situations where we use this, it's a good idea to add a comment explaining what dependencies we're explicitly excluding.

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,11 +192,59 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
   });
 }
 
-export function getPhonePersonAttributeValueFromFhirPatient(patient: fhir.Patient) {
+export async function getIdentifierFieldValuesFromFhirPatient(
+  patient: fhir.Patient,
+  identifierConfig,
```
ibacher, line 197: We should have a type here of some sort right?

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,11 +192,59 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+): Promise<{ [identifierFieldName: string]: PatientIdentifierValue }> {
+  const identifiers: FormValues['identifiers'] = {};
+  const promises: Promise<void>[] = [];
```
ibacher, line 200: nit:

```suggestion
  const promises: Array<Promise<void>> = [];
```

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,11 +192,59 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+
+  for (const identifier of patient.identifier) {
+    for (const config of identifierConfig) {
```
ibacher, line 203: Rather than two nested for loops, it seems like it would make sense to loop through the `identifierConfig` once and create an object that keyed from `fhirIdentifierSystem` to the `config` object associated with it. That way, we loop through it once and the lookup from identifier -> identifierConfig is an O(1) operation.

packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -192,11 +192,59 @@ export function getPatientIdentifiersFromFhirPatient(patient: fhir.Patient): Arr
...
+
+      const url = `${restBaseUrl}/patientidentifiertype/${config.openmrsIdentifierTypeUuid}`;
+
+      promises.push(
+        openmrsFetch(url)
+          .then((response) => {
+            if (!response.data?.name) {
+              return;
+            }
+            identifiers[response.data.name] = {
```
ibacher, line 216: I think the assumptions we're making here about the OpenMRS identifiers MPI identifiers correspond to need to be documented, i.e., I'm not always sure that `preferred : false` is the correct behaviour.

packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
ibacher: @samuelmale I'm more concerned with the user winding up in a weird state when they, e.g., submit a patient registration and then click the Back button or something similar. But maybe that's silly?

packages/esm-patient-search-app/src/patient-search-page/patient-search-views.component.tsx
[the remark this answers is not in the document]
ibacher: I'm fine with either, but `'local'` requires changing more things.

samuelmale:
> I'm more concerned with the user winding up in a weird state when they, e.g., submit a patient registration and then click the Back button or something similar. But maybe that's silly?

@ibacher that makes a lot of sense!

gracepotma:
Hi guys, I'm wondering - are we planning to get this into a release at any point? How about 3.5? (We've missed the cut-off for 3.4 last week.)

reagan-meant:
Config example
``` 
 "@openmrs/esm-patient-registration-app": {
    "identifierMappings": [
      {
        "fhirIdentifierSystem": "urn:oid:1.2.36.146.595.217.0.1",
        "openmrsIdentifierTypeUuid": "a71403f3-8584-4289-ab41-2b4e5570bd45"
      }
    ]
  }
````

PR #1437, opened 2025-01-12
(fix) O3-4356: Fix ward location selection for bed creation and occupancy strings

jbravo94:
## Requirements

## Summary
- The occupation status was shown in the UI with the same text which was fixed.
- Beds could not be created because ComboBox onChange does with current version of @carbon/react < 1.72 not work because of an existing bug (https://github.com/carbon-design-system/carbon/issues/18145#issuecomment-2521936772) - an inline workaround was added which can be removed later when new version is available in core.

## Screenshots
image

## Related Issues
https://openmrs.atlassian.net/browse/O3-4356
https://openmrs.atlassian.net/browse/O3-4355

## Other
- Result was manually tested
- Other translations files would need updates of the keys as well - feedback required.


mogoodrich:
See: https://talk.openmrs.org/t/bed-management-in-refapp/31942/6

packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
```diff
@@ -232,8 +241,8 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
                     {occupancyStatuses.map((occupancyStatus, index) => (
                       <SelectItem
                         key={`occupancyStatus-${index}`}
-                        text={t('occupancyStatus', `${occupancyStatus}`)}
-                        value={t('occupancyStatus', `${occupancyStatus}`)}
+                        text={t(`occupancyStatus${occupancyStatus}`, `${occupancyStatus}`)}
+                        value={t(`occupancyStatus${occupancyStatus}`, `${occupancyStatus}`)}
```
mseaton, line 245: The value of the item should not be translated, should it?  Isn't this what is sent to the server?  Only the displayed value should be translated.

packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
```diff
@@ -203,7 +204,15 @@ const BedAdministrationForm: React.FC<BedAdministrationFormProps> = ({
                     items={allLocations}
                     itemToString={(location) => location?.display ?? ''}
                     label={t('location', 'Location')}
-                    onBlur={onBlur}
+                    /*
+                      TODO: ComboBox bug does not trigger onChange below @carbon/react version 1.72 (https://github.com/carbon-design-system/carbon/issues/18145#issuecomment-2521936772)
+                      onBlur shall be refactored to onBlur={onBlur}
```
mseaton, line 209: What exactly is the TODO here?

mseaton (changes requested):
See comments.  Also, all PRs should have an associated JIRA ticket that describes the change that can be tracked.  Please create one and refer to it in the commit title and comment.

NethmiRodrigo:
This is a bit weird, but while I can recreate the underlying issue this PR fixes on dev3, it doesn't happen locally on main. 
Also, I added a console log to check and onChange does get fired

https://github.com/user-attachments/assets/756888d2-04e3-4593-bc18-bdbed187b500

packages/esm-bed-management-app/src/bed-administration/bed-administration-form.component.tsx
jbravo94: The todo means that the original onChange method should be used if the core has a more updated version for @carbon/react which does not have the but there - but this would required bumping and testing packages in core which i thought is not that simple. But it can be left out as well if you prefer.

jbravo94:
I checked only in the distro not with the local build. My assumption is that the current used version of @carbon/react in production mode has exactly this issue: https://github.com/carbon-design-system/carbon/issues/18145#issuecomment-2521936772

onChange was not fired when I tested in the bundled distro

mseaton (approved):
LGTM - thanks!

NethmiRodrigo:
> I checked only in the distro not with the local build.

> onChange was not fired when I tested in the bundled distro

@jbravo94 what do you mean when you say you tested in the bundled distro?

NethmiRodrigo:
Nvm, please ignore my previous comment, you're right about it coming up in production mode @jbravo94. Thanks!

NethmiRodrigo (approved):

NethmiRodrigo:
Merging this in and updating the tickets since the original tickets were made for the backend module. For future reference @jbravo94 here's the epic for the bed management frontend - https://openmrs.atlassian.net/browse/O3-3720 :)

jbravo94:
@NethmiRodrigo Thank you!

PR #1449, opened 2025-01-17
(feat) O3-4335: Enable multiple filters on the appointments app dashboard

Samstar10:
## Requirements

## Summary
This PR introduces a new feature that allows a user to filter appointments by multiple service types at once in the Appointments app dashboard.

## Screenshots
Screenshot 2025-01-17 at 12 59 58

## Related Issue
https://openmrs.atlassian.net/browse/O3-4335


paulsonder:
Hi @Samstar10 - this looks good to me based on the screenshot.

Samstar10:
> Hi @Samstar10 - this looks good to me based on the screenshot.
> 
> Do the numbers update in the metrics cards? E.g. 'Scheduled appointments'?

Yes they do @paulsonder

denniskigen:
@paulsonder @ciaranduffy I'm not entirely sure that the [MenuButton](https://carbondesignsystem.com/components/menu-buttons/usage/) is the correct Carbon component to use for this table filter.
On the one hand, it makes sense to use it if the goal is to hide extra actions in limited screen space:
CleanShot 2025-01-28 at 1  14 37@2x

But it doesn't make sense if the goal is to display a list of filters:
CleanShot 2025-01-28 at 1  16 30@2x

The [MultiSelect](https://carbondesignsystem.com/components/dropdown/usage/#multiselect) component is better suited for displaying a list of filters, but its not ideal for situations where space is limited.

So, I'm unsure which component is the best fit. To add another consideration, this problem has been solved in a different way in the service queues page where a list of filters are displayed in the table header.
CleanShot 2025-01-28 at 1  21 30@2x

One specific aspect I like about the approach there is the proximity of the filters to the table for the following reasons:

- It gives users immediate context - filters are placed directly above the table, creating a clear visual hierarchy. Users instinctively understand that these controls affect the data below.
- It reduces cognitive load - no need to search for filters in the page header. Everything related to the table data is grouped together.
- It enables faster iteration - users can adjust filters and immediately see the results without scrolling or navigating to another part of the UI.

Do you think we should consider adopting this approach for the appointments page?

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -10,14 +17,51 @@ import styles from './appointments-header.scss';
...
+  
+      setSelectedItems(updatedSelectedItems);
+    }
+  };
+  
+  
+
+  useEffect(() => {
+    onChange?.('');
+  }, [])
```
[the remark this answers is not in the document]
Samstar10: @denniskigen if we initialise the onChange function using an empty array the table becomes reset to an empty state rather than showing all categories on initialisation as is intended.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -10,14 +17,51 @@ import styles from './appointments-header.scss';
...
+  appointmentServiceType?: string[];
   onChange?: (evt) => void;
 }
 
 const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointmentServiceType, onChange }) => {
   const { t } = useTranslation();
   const { selectedDate, setSelectedDate } = useContext(SelectedDateContext);
   const { serviceTypes } = useAppointmentServices();
+  
+  const items = [{ name: 'All', uuid: '' }, ...serviceTypes];
```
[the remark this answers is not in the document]
Samstar10: @denniskigen when the "All" option is selected, it resets all other filters selected and also when a filter is selected the "All" is automatically unselected. This is convenient for when a user selects multiple filters and wants to go back to seeing all the categories, they don't have to go back and unselect all the selected categories but instead they can just click "All" and have everything reset. So when a filter is selected, the All is automatically unselected and vice versa.

Samstar10:
> @paulsonder @ciaranduffy I'm not entirely sure that the [MenuButton](https://carbondesignsystem.com/components/menu-buttons/usage/) is the correct Carbon component to use for this table filter.
> 
> On the one hand, it makes sense to use it if the goal is to hide extra actions in limited screen space:
> 
> CleanShot 2025-01-28 at 1  14 37@2x
> 
> But it doesn't make sense if the goal is to display a list of filters:
> 
> CleanShot 2025-01-28 at 1  16 30@2x
> 
> The [MultiSelect](https://carbondesignsystem.com/components/dropdown/usage/#multiselect) component is better suited for displaying a list of filters, but its not ideal for situations where space is limited.
> 
> So, I'm unsure which component is the best fit. To add another consideration, this problem has been solved in a different way in the service queues page where a list of filters are displayed in the table header.
> 
> CleanShot 2025-01-28 at 1  21 30@2x
> 
> One specific aspect I like about the approach there is the proximity of the filters to the table for the following reasons:
> 
> * It gives users immediate context - filters are placed directly above the table, creating a clear visual hierarchy. Users instinctively understand that these controls affect the data below.
> * It reduces cognitive load - no need to search for filters in the page header. Everything related to the table data is grouped together.
> * It enables faster iteration - users can adjust filters and immediately see the results without scrolling or navigating to another part of the UI.
> 
> Do you think we should consider adopting this approach for the appointments page?

@denniskigen should I wait for response on this before I push the requested changes?

ciaranduffy:
Thanks for your patience while we got around to this @Samstar10 and thank you @denniskigen for catching this point. You're absolutely correct to say that the MenuButton isn't the correct component here and the component you've pointed to in the Service Queues would in fact be more suitable. That component is https://carbondesignsystem.com/components/dropdown/usage/ and it's the `inline` variant. 

Hope that's helpful and helps you move this forward

Samstar10:
> Thanks for your patience while we got around to this @Samstar10 and thank you @denniskigen for catching this point. You're absolutely correct to say that the MenuButton isn't the correct component here and the component you've pointed to in the Service Queues would in fact be more suitable. That component is https://carbondesignsystem.com/components/dropdown/usage/ and it's the `inline` variant.
> 
> Hope that's helpful and helps you move this forward

Hello @ciaranduffy, the dropdown component does not allow selection of multiple items and wouldn't be able to meet the feature requirements.

ciaranduffy:
@Samstar10 wouldn't you just use the multi-select variant of the same component then? 

I think this pattern could be improved though. Carbon's multi-select component doesn't do a great job of showing you what filters are applied, only that two (or more) filters are applied, but that's something that could be improved in a future iteration, I think.

Samstar10:
> @Samstar10 wouldn't you just use the multi-select variant of the same component then?
> 
> I think this pattern could be improved though. Carbon's multi-select component doesn't do a great job of showing you what filters are applied, only that two (or more) filters are applied, but that's something that could be improved in a future iteration, I think.

Alright @ciaranduffy, that sounds goods to me. I'll get it done. Thank you.

Samstar10:
Hello @denniskigen, I have implemented the requested changes.
Screenshot 2025-02-04 at 12 31 04

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -19,6 +19,26 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
   const { selectedDate, setSelectedDate } = useContext(SelectedDateContext);
   const { serviceTypes } = useAppointmentServices();
 
+  const items = [...serviceTypes];
```
denniskigen, line 22:
```suggestion
```

`items` seems unnecessary to me. We should be able to use `serviceTypes` directly.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -19,6 +19,26 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
+        setSelectedItems(selectedUuids);
+        onChange?.(selectedUuids);
+      }
+    },
+    [onChange],
+  );
+
+  useEffect(() => {
+    onChange?.('');
+  }, [onChange]);
```
denniskigen, line 40:
```suggestion
```

I think this useEffect should be removed in favor of the suggested logic in the event handler above.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -19,6 +19,26 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
+
+  const handleMultiSelectChange = useCallback(
+    ({ selectedItems }) => {
+      const selectedUuids = selectedItems.map((item) => item.id);
+      if (selectedUuids.length === 0) {
+        setSelectedItems([]);
+      } else {
+        setSelectedItems(selectedUuids);
+        onChange?.(selectedUuids);
+      }
```
denniskigen, line 33:
```suggestion
      setSelectedItems(selectedUuids);
      onChange?.(selectedUuids);
```

We should call the `onChange` callback regardless of whether there are selected items or not. Not doing so results in unexpected results when you clear the checkboxes.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -19,6 +19,26 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
+    },
+    [onChange],
+  );
+
+  useEffect(() => {
+    onChange?.('');
+  }, [onChange]);
+
```
denniskigen, line 41:
```suggestion
  const serviceTypeOptions = useMemo(
    () => serviceTypes?.map((item) => ({ id: item.uuid, label: item.name })) ?? [],
    [serviceTypes],
  );
```

Let's memoize the `serviceTypes` mapping.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -37,20 +57,13 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
-            onChange={({ selectedItem }) => onChange(selectedItem?.uuid)}
-            selectedItem={
-              serviceTypes.find((service) => service.uuid === appointmentServiceType) || { name: 'All', uuid: '' }
-            }
-            size="sm"
-            titleText={t('view', 'View')}
+          <MultiSelect
+            id="serviceTypeMultiSelect"
+            label={t('filterByServiceType', 'Filter by service type')}
+            items={items.map((item) => ({ id: item.uuid, label: item.name }))}
```
denniskigen, line 63:
```suggestion
            items={serviceTypeOptions}
```

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -37,20 +57,13 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
-            label={t('selectServiceType', 'Select service type')}
-            onChange={({ selectedItem }) => onChange(selectedItem?.uuid)}
-            selectedItem={
-              serviceTypes.find((service) => service.uuid === appointmentServiceType) || { name: 'All', uuid: '' }
-            }
-            size="sm"
-            titleText={t('view', 'View')}
+          <MultiSelect
+            id="serviceTypeMultiSelect"
+            label={t('filterByServiceType', 'Filter by service type')}
```
denniskigen, line 62:
```suggestion
            label={t('filterAppointmentsByServiceType', 'Filter appointments by service type')}
```

I propose changing the MultiSelect label for two reasons:

1. The suggested label is longer which means the dropdown menu is larger, giving us more room for the filter labels.
2. The suggested label is arguably clearer about communicating what this filter does to the user.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -37,20 +57,13 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
...
+            items={items.map((item) => ({ id: item.uuid, label: item.name }))}
+            itemToString={(item) => (item ? item.label : '')}
+            onChange={handleMultiSelectChange}
+            initialSelectedItems={items.length > 0 ? [items[0]] : []}
```
denniskigen, line 66:
```suggestion
            initialSelectedItems={serviceTypeOptions.length > 0 ? [serviceTypeOptions[0].id] : []}
```

packages/esm-appointments-app/src/header/appointments-header.scss
```diff
@@ -22,14 +22,9 @@
...
-    overflow: hidden;
-  }
-
-  :global(.cds--dropdown--inline .cds--list-box__menu) {
-    left: -(layout.$spacing-10);
+.menuButton {
+  :global(.cds--btn--ghost) {
+    color: colors.$gray-80;
   }
 }
```
denniskigen, line 29: I suppose we can remove this class.

packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
```diff
@@ -4,7 +4,7 @@ import { useAppointmentList } from '../../hooks/useAppointmentList';
 import AppointmentsTable from '../common-components/appointments-table.component';
```
denniskigen, line 5: We need to distinguish between two empty states: 

1. Initial empty state with no appointments 
2. The empty state that may result after applying filters with no matching appointments 

To address the second empty state, we might need to pass a prop down to the `AppointmentsTable` that determines when to show this filter empty state. We could do something like this:

```tsx
return (
    <AppointmentsTable
      appointments={activeAppointments}
      isLoading={isLoading}
      tableHeading={title}
      hasActiveFilters={appointmentServiceType?.length > 0 || filterCancelled}
    />
  );
```

And then we could add `hasActiveFilters` to `AppointmentsTable` as an optional boolean prop. We could then use the prop like so:

```tsx
  if (hasActiveFilters && !appointments?.length) {
    return (
      <div className={styles.filterEmptyState}>
        <Layer level={0}>
          <Tile className={styles.filterEmptyStateTile}>
            <p className={styles.filterEmptyStateContent}>
              {t('noMatchingAppointments', 'No matching appointments found')}
            </p>
            <p className={styles.filterEmptyStateHelper}>{t('checkFilters', 'Check the filters above')}</p>
          </Tile>
        </Layer>
      </div>
    );
  }
```

The accompanying CSS could look like this:

```scss
.filterEmptyState {
  display: flex;
  justify-content: center;
  align-items: center;
  margin: layout.$spacing-05 0;
  padding: layout.$spacing-08 0;
  text-align: center;
  border: 1px solid $ui-03;
}

.filterEmptyStateTile {
  margin: auto;
  width: fit-content;
}

.filterEmptyStateContent {
  @include type.type-style('heading-compact-02');
  color: $text-02;
  margin-bottom: layout.$spacing-03;
}

.filterEmptyStateHelper {
  @include type.type-style('body-compact-01');
  color: $text-02;
}
```

With that, you'd end up with a filter empty state that looks like this:

CleanShot 2025-02-05 at 10  42 09@2x

In this case, the only Checked In appointment is for the `Outpatient Department` service whereas the selected filter is `Rehabilitation Service`. So instead of rendering the EmptyState component and mistakenly telling the user that there are no checked in appointments, this filter empty state communicates clearly that there are no matching appointments for the selected filter.

denniskigen:
Can we also modify the condition in the `filterByServiceType` function's return to check for the `appointmentServceType` array's length instead of its existence? Arrays are objects. Even if an array is empty, its still an object. Objects are truthy so an empty array would evaluate to true in a boolean context. Array length is falsy, so we should use that instead:

```tsx
export function filterByServiceType(appointmentList: any[], appointmentServiceType: string[]) {
  return appointmentServiceType?.length > 0
    ? appointmentList.filter(({ service }) => appointmentServiceType.includes(service.uuid))
    : appointmentList;
}
```

Samstar10:
Hello @denniskigen I have implemented the requested changes. Please review.

denniskigen (approved):
LGTM. Thanks, @Samstar10!

PR #1461, opened 2025-01-22
(fix) O3-4373 :  Default Queue Priority When Moving to Next Service 

Muppasanipraneeth:
## Requirements

## Summary

# issue
- When moving a patient to the next service, the queue priority should default to whatever its set to
- Example The patient is in “Emergency“ the default here [Move patient to the next service] should be “Emergency“ and vice versa. 

## Screenshots

# Before

https://github.com/user-attachments/assets/5cae9e2b-b1cc-44ad-9e38-30c1394c33c5

# After

https://github.com/user-attachments/assets/18da9882-1e74-4ea8-abce-2188ede15479

https://github.com/user-attachments/assets/eff52ded-1aa0-487a-b7b8-78c78d56e010


packages/esm-service-queues-app/src/active-visits/change-status-dialog.component.tsx
```diff
@@ -37,6 +37,7 @@ const ChangeStatus: React.FC<ChangeStatusDialogProps> = ({ queueEntry, closeModa
   const { t } = useTranslation();
   const { concepts } = useConfig<ConfigObject>();
   const { allowedPriorities, allowedStatuses } = queueEntry.queue ?? {};
+  const userPriority = queueEntry.priority.toString();
```
chibongho, line 40: - Generally, use uuid instead of display string  to identify the thing we want to select.
- `user` is not accurate. That would be the clinician / provider. maybe `currentPriority`?

packages/esm-service-queues-app/src/active-visits/change-status-dialog.component.tsx
```diff
@@ -57,7 +58,7 @@ const ChangeStatus: React.FC<ChangeStatusDialogProps> = ({ queueEntry, closeModa
     formState: { isSubmitting, errors },
     getValues,
   } = useForm<ChangeStatusForm>({
-    defaultValues: { priority: allowedPriorities[1]?.uuid },
+    defaultValues: { priority: userPriority },
```
chibongho, line 61: Does this still work after changing from uuid to display string? (Has this been tested?)

packages/esm-service-queues-app/src/active-visits/change-status-dialog.component.tsx
```diff
@@ -57,7 +58,7 @@ const ChangeStatus: React.FC<ChangeStatusDialogProps> = ({ queueEntry, closeModa
     formState: { isSubmitting, errors },
     getValues,
   } = useForm<ChangeStatusForm>({
-    defaultValues: { priority: allowedPriorities[1]?.uuid },
+    defaultValues: { priority: currentPriority || allowedPriorities[1]?.uuid },
```
chibongho, line 61: Let's not do the `allowedPriorities[1]` thing at all. It's just wrong. 

I don't think `currentPriority` would ever be nullish, but if so, we can use `concepts.defaultPriorityConceptUuid`. Github won't let me comment on that line, but we can clean up line 71-72 while we're at it.

packages/esm-service-queues-app/src/types/index.ts
```diff
@@ -318,6 +318,7 @@ export interface MappedQueueEntry {
   patientUuid: string;
   queue: Queue;
   priority: Concept;
+  priorityUuid: string;
```
chibongho, line 321: This field doesn't exist. Adding it to the interface won't help. See `mapVisitQueueEntryProperties`, the actual function to convert from `QueueEntry` to `MappedQueuedEntry`.

Better to just stick to using `priority.uuid`

packages/esm-service-queues-app/src/active-visits/change-status-dialog.component.tsx
```diff
@@ -37,6 +37,7 @@ const ChangeStatus: React.FC<ChangeStatusDialogProps> = ({ queueEntry, closeModa
   const { t } = useTranslation();
   const { concepts } = useConfig<ConfigObject>();
   const { allowedPriorities, allowedStatuses } = queueEntry.queue ?? {};
+  const currentPriority = queueEntry.priorityUuid;
```
chibongho, line 40: `queueEntry.priority.uuid`

chibongho (changes requested):

chibongho (approved):

PR #1464, opened 2025-01-23
(fix) O3-4387: Fetch active visits from parent location when login location is not a visit location

NethmiRodrigo:
## Requirements

## Summary

This PR updates the `useActiveVisits` url to have the `includeParentLocation` search param to account for when a user starts an active visit from a login location that is not a visit location. In situations like such, we should show the active visits from the parent location which is a visit location.

## Related Issue

https://issues.openmrs.org/browse/O3-4387

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,12 @@ You could provide `yarn start` with as many `sources` arguments as you require. 
 yarn start --sources 'packages/esm-patient-search-app' --sources 'packages/esm-patient-registration-app'
 ```
 
+To run an app locally with the same [configuration that is set in the reference application](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/main/frontend/config-core_demo.json), use:
+
+```bash
+yarn start --config-url /openmrs/spa/config-core_demo.json --sources 'packages/esm-<insert-package-name>-app'
+```
+
 ## Troubleshooting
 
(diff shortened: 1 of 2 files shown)


NethmiRodrigo:
A weird thing this brought up is that, I used the login location `Outpatient Clinic` to test. This is a login location but **not a visit location**. It is a child of `Ubuntu Hospital` which is a visit location. On dev3, when you are logged in to `Outpatient clinic` and start a visit for a patient, for the visit location, `Ubuntu Hospital` is selected by default, and `Outpatient clinic` is not an option (this is expected behavior). 

But locally, the behavior is very different. When you try to start a visit, the default selected location is `Outpatient clinic`, and `Ubuntu Hospital` is not even an option. This is very weird because the location hierarchy is the same in both, and also renders this fix pretty much useless locally.
Screenshot 2025-01-23 at 7 05 28 PM
Screenshot 2025-01-23 at 7 07 01 PM

denniskigen (approved):
LGTM. Thanks, @NethmiRodrigo. We should really find a way to document this somewhere. Maybe in the [Patient Chart app's README](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/README.md)?

PR #1468, opened 2025-01-24
(fix) Improve code splitting and dependencies

ibacher:
## Requirements

## Summary

This PR does two things:

* It moves "shared" dependencies into the apps actual dependency list
* It reworks how the appointments app does code splits to reduce the overall size of the app

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,20 +15,10 @@
     "test-e2e": "playwright test",
     "postinstall": "husky install"
   },
-  "dependencies": {
-    "@hookform/resolvers": "^3.3.1",
-    "@internationalized/date": "^3.5.4",
-    "classnames": "^2.3.2",
-    "react-hook-form": "^7.46.2",
-    "swr": "^2.0.1",
-    "xlsx": "^0.18.5",
-    "zod": "^3.22.2"
-  },
   "devDependencies": {
     "@babel/core": "^7.11.6",
     "@carbon/react": "^1.71.0",
     "@openmrs/esm-framework": "next",
-    "@openmrs/esm-patient-common-lib": "next",
     "@playwright/test": "1.49.0",
     "@swc/core": "^1.2.165",
     "@swc/jest": "^0.2.29",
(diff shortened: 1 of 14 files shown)


denniskigen (approved):
Excellent! Thanks, @ibacher!

denniskigen:
@ibacher what's the rule of the thumb now re: `getAsyncLifecycle` and `getSyncLifecycle`. In a past iteration, the prevailing thought seemed to be to prefer using `getSyncLifecycle` for pages and dashboard links, and `getAsyncLifecycle` for everything else.

Has this changed? Should we be using `getAsyncLifecycle` for everything except dashboard links?

ibacher:
So this PR originated when I was playing around with caching and noticed the appointments app was downloading a 350kB chunk of code, which seemed large (most chunks are <50kB). When I ran the analyzer there were a few minor things (it was embedding react-dom and xlsx was probably larger than it needed to be), but I noticed that Carbon was being loaded in both the "main" chunk and the 350kB one. At its heart, that's what this is fixing.

So my current approximation for a rule here is: don't load something that makes heavy use of Carbon synchronously, especially "heavier" components like data tables or other things where each component uses multiple sub-components). I'm also thinking that most apps shouldn't have more than 2-3 synchronously loaded components, since that seems to be where we wind up.

Maybe that can be glossed as "just load things in the sidebar synchronously"?

denniskigen:
Yeah, thanks for the context. I ask because I'd gone on to file a separate PR inspired by this one, but I feel it goes too far in the other direction as we end up with approximately one or two max synchronously loaded components per app.

ibacher:
> Does that sound too far off the mark?

No that sounds about right. I think previously, I thought that front-loading as much of an app as we could would yield better performance, but I'm no longer sure that's the case (though it probably gets you better performance on initial load).

denniskigen:
Ok, here's the PR https://github.com/openmrs/openmrs-esm-patient-management/pull/1472

PR #1472, opened 2025-01-27
(refactor) Improve code splits further across apps

denniskigen:
## Requirements

## Summary

Related: #1468 

This PR makes changes to several `index.ts` files across apps converting synchronous component imports to asynchronous ones using dynamic imports. This refactoring improves performance by reducing the initial bundle size and improving the page load time. With the additional code splitting, components are only loaded when they are actually needed. Key changes include:

- Replacing `getSyncLifecycle` with `getAsyncLifecycle` and dynamic imports.
- Removing direct component imports from `index.ts` files.

### DIFF
--- a/packages/esm-active-visits-app/src/index.ts
+++ b/packages/esm-active-visits-app/src/index.ts
@@ -1,7 +1,5 @@
 import { defineConfigSchema, getAsyncLifecycle, getSyncLifecycle } from '@openmrs/esm-framework';
 import { configSchema } from './config-schema';
-import activeVisitsComponent from './active-visits-widget/active-visits.component';
-import visitDetailComponent from './visits-summary/visit-detail.component';
 
 const moduleName = '@openmrs/esm-active-visits-app';
 
@@ -16,9 +14,9 @@ export function startupApp() {
   defineConfigSchema(moduleName, configSchema);
 }
 
-export const activeVisits = getSyncLifecycle(activeVisitsComponent, options);
+export const activeVisits = getAsyncLifecycle(() => import('./active-visits-widget/active-visits.component'), options);
 
-export const visitDetail = getSyncLifecycle(visitDetailComponent, options);
+export const visitDetail = getAsyncLifecycle(() => import('./visits-summary/visit-detail.component'), options);
 
 export const homeActiveVisitsTile = getAsyncLifecycle(
   () => import('./home-page-tiles/active-visits-metric-tile/active-visits-tile.component'),
--- a/packages/esm-bed-management-app/src/index.ts
+++ b/packages/esm-bed-management-app/src/index.ts
@@ -1,4 +1,4 @@
-import { getAsyncLifecycle, defineConfigSchema, getSyncLifecycle } from '@openmrs/esm-framework';
+import { defineConfigSchema, getAsyncLifecycle, getSyncLifecycle } from '@openmrs/esm-framework';
 import { configSchema } from './config-schema';
 import { createLeftPanelLink } from './left-panel-link.component';
 
--- a/packages/esm-patient-list-management-app/src/index.ts
+++ b/packages/esm-patient-list-management-app/src/index.ts
@@ -3,9 +3,6 @@ import { configSchema } from './config-schema';
 import { createDashboardLink } from './createDashboardLink.component';
 import { dashboardMeta } from './dashboard.meta';
 import { setupOffline } from './offline';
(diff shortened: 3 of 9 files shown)


samuelmale (approved):
LGTM; thanks @denniskigen!

denniskigen:
Thanks, @samuelmale. Please don't merge this yet. Keen to discuss the potential tradeoffs some more. Also keen to develop a clear rubric that we can document about where the sweet spot is for this potential performance optimization.

samuelmale:
What are the bundle size and performance implications of not lazy-loading the root component(s)? (I understand that some root components heavily ref on Carbon components)

chibongho:
Are there prior discussions about performance that might provide context to what / how to optimize? Reducing the bundle size is good, but don't we aggressively cache the JS files in the browser anyway? If I understand correctly, this should improve the load time when we first time we load O3 after a deploy, but the performance impact would not be that significant for subsequent loads.

ibacher:
> What are the bundle size and performance implications of not lazy-loading the root component(s)?

It's hard to know in detail. This is a balancing act, not an exact science. The microfrontend architecture and especially the extension system make ensuring that quite challenging, so a lot of this is going to be trial-and-error.

High-level, we'd want to get to a place where:

* We're limited to as close to, say, no more than 20 blocking requests to load a single "page"
* Over a "good" connection the app can load < 1 second and over a laggy connection it takes no more than 30 seconds
* With all JS and openmrs.css cached, most pages load < 300kB of data and all pages load < 1 MB of data

How we get there, though, is what we need to experiment with. Essentially, adding code splits trade a network request for a smaller volume of data-over-the-wire in each case, but it can go wrong in two ways:

### Chunks that are too small

Currently, the leading edge of this is the translations for the home page app, which has only a single translation key, `home`. What gets sent over the wire is this:

```js
"use strict";(globalThis.webpackChunk_openmrs_esm_home_app=globalThis.webpackChunk_openmrs_esm_home_app||[]).push([[300],{4300:e=>{e.exports={home:"Home"}}}]);
```

Which is 160 bytes of data to load 13 bytes of "usable" data (the JS object `{home:"Home"}`. Worse still because the headers with the CSP take up 622 bytes, we actually send 782 bytes over the wire for 13 usable bytes. This is a little bit silly and I don't think it's a real problem, but we don't want to end up in a state where we have too many such cases.

### Chunks that are too large

On the flip side of things, we also end up with chunks that are probably larger than they should be. Currently on dev3, the largest JS file is a chunk from the billing app consisting of this:

Billing App

However, it compresses pretty well (a lot of that is embedded CSS, which likely has a large amount of repetition, so it compresses very well), so the most space consuming thing sent is currently this:

ESM Core

Which is Carbon packaged as a peer dependency (although if we start using Carbon as a peer dependency everywhere, this chunk only needs to be loaded once and we can replace all the places that Carbon gets embedded in the app, so maybe we do want this).

Even these "large" chunks aren't _huge_, under 300 kB when gzipped, but they can often point to places where code has either been split inefficiently (e.g., why are all the SCSS files for the whole billing app in a single chunk instead of paired with the code for each extension?) or where there may be unnecessary duplication of libraries (which is what #1468 was ultimately about).

But I don't have any really rote approach to how we determine these, we just need to keep looking at:

* The number of files being requested
* The size of those files
* The load time over those files with a "slow" connection

Because, ultimately, the goal here—more than any hard numbers—is that it the app feels responsive to users.

To be clear, I don't mean the above examples to be cases of things we should get away with. It probably makes sense to load translations the way we currently do because while the 782 bytes to 13 useful bytes ratio sounds atrocious, it also means we can use a single loading pathway that supports both the home app and an app with 120 different translation keys. Similarly, the Carbon chunk is huge relative to the app, but if we only load it a single time, it will likely reduce the overall amount of data transferred and, since it's loaded fairly early, doing so might actually make the app "feel" more responsive (since we paid the loading cost before the login page fully rendered).

> Are there prior discussions about performance that might provide context to what / how to optimize?

Yes, but they are mostly captured in PRs like this. I think we've synthesized some of them into the developer docs though, where we can.

> Reducing the bundle size is good, but don't we aggressively cache the JS files in the browser anyway?

Yeah, but we run into the reality that many long-term users of OpenMRS-based systems have been taught to empty the cache an reload pages when things in the EMR don't work. And, in any case, as we see larger facilities deploy OpenMRS and more cloud-based deployments, any bandwidth use we can shave helps.

PR #1487, opened 2025-02-17
(test) Use UI flow instead of direct API calls to test removing a patient from a list

NethmiRodrigo:
## Requirements

## Summary

The e2e test for managing a patient list heavily relies on directly making API  calls rather than interacting with the user interface, which defeats the purpose of it existing. This updates just the delete patient to hopefully get more info as to why it fails in [this](https://github.com/openmrs/openmrs-esm-core/pull/1280) PR.

### DIFF
--- a/e2e/specs/patient-list.spec.ts
+++ b/e2e/specs/patient-list.spec.ts
@@ -73,17 +73,28 @@ test('Manage patients in a list', async ({ api, page }) => {
     await patientListPage.goto(cohort.uuid);
   });
 
-  await test.step('Then I should be able to add and remove patients from that list', async () => {
+  await test.step('Then I should be able to add a patient to the list', async () => {
     // Add a patient to the list
     cohortMember = await addPatientToCohort(api, cohort.uuid, patient.uuid);
     await patientListPage.goto(cohort.uuid);
     await expect(patientListPage.patientListHeader()).toHaveText(/1 patients/);
     await expect(patientListPage.patientsTable()).toHaveText(new RegExp(patient.person.display));
+  });
 
-    // Remove a patient from the list
-    await removePatientFromCohort(api, cohortMember.uuid);
-    await patientListPage.goto(cohort.uuid);
-    await expect(patientListPage.patientListHeader()).toHaveText(/0 patients/);
+  await test.step('And when I click on the bin icon to remove the patient from the list', async () => {
+    await page.getByRole('button', { name: /remove from list/i }).click();
+  });
+
+  await test.step("Then I click the 'Remove from list' in the delete modal to confirm the action", async () => {
+    await page.getByRole('button', { name: 'danger Remove from list' }).click();
+  });
+
+  await test.step('And then I should see a success message', async () => {
+    await expect(page.getByText(/patient removed from list/i)).toBeVisible();
+  });
+
+  await test.step('And then the patient count should update to zero', async () => {
+    await expect(page.getByText(/0 patients/)).toBeVisible();
     cohortMember = null;
   });
 });


e2e/specs/patient-list.spec.ts
```diff
@@ -73,16 +73,24 @@ test('Manage patients in a list', async ({ api, page }) => {
...
 
-    // Remove a patient from the list
-    await removePatientFromCohort(api, cohortMember.uuid);
-    await patientListPage.goto(cohort.uuid);
+  await test.step('And then I should be able to remove a patient from the list', async () => {
+    await page.getByLabel(/remove from list/i).click();
+  });
+
+  await test.step("And then I click the 'Remove from list' to confirm", async () => {
+    await page.getByRole('button', { name: 'danger Remove from list' }).click();
```
vasharma05, line 89: > danger Remove from list?

This looks weird 🤔

NethmiRodrigo: That's how its done in elsewhere -https://github.com/openmrs/openmrs-esm-patient-management/blob/9201ed389bd8f811ff2c1f265c4ccbf60d5c3079/e2e/specs/appointments.spec.ts#L120 + what Playwright locator suggests

e2e/specs/patient-list.spec.ts
```diff
@@ -73,16 +73,24 @@ test('Manage patients in a list', async ({ api, page }) => {
...
     await expect(patientListPage.patientListHeader()).toHaveText(/1 patients/);
     await expect(patientListPage.patientsTable()).toHaveText(new RegExp(patient.person.display));
+  });
 
-    // Remove a patient from the list
-    await removePatientFromCohort(api, cohortMember.uuid);
-    await patientListPage.goto(cohort.uuid);
+  await test.step('And then I should be able to remove a patient from the list', async () => {
+    await page.getByLabel(/remove from list/i).click();
+  });
```
denniskigen, line 86:
```suggestion
  await test.step('And when I click on the bin icon to remove the patient from the list', async () => {
    await page.getByLabel(/remove from list/i).click();
  });
```

e2e/specs/patient-list.spec.ts
```diff
@@ -73,16 +73,24 @@ test('Manage patients in a list', async ({ api, page }) => {
...
+
+  await test.step("And then I click the 'Remove from list' to confirm", async () => {
```
denniskigen, line 88:
```suggestion
  await test.step("Then I click the 'Remove from list' in the delete modal to confirm the action", async () => {
```

e2e/specs/patient-list.spec.ts
```diff
@@ -73,16 +73,24 @@ test('Manage patients in a list', async ({ api, page }) => {
...
     await patientListPage.goto(cohort.uuid);
     await expect(patientListPage.patientListHeader()).toHaveText(/1 patients/);
     await expect(patientListPage.patientsTable()).toHaveText(new RegExp(patient.person.display));
+  });
 
-    // Remove a patient from the list
-    await removePatientFromCohort(api, cohortMember.uuid);
-    await patientListPage.goto(cohort.uuid);
+  await test.step('And when I click on the bin icon to remove the patient from the list', async () => {
+    await page.getByLabel(/remove from list/i).click();
```
denniskigen, line 85:
```suggestion
    await page.getByRole('button', { name: /remove from list/i }).click();
```

NethmiRodrigo:
The problem here is that the patient lists header count is supposed to update and show `0 patients` instead of `1 patients` when a patient gets deleted. Except it doesn't

NethmiRodrigo:
To add more context, this PR tries to update the e2e test that deletes a patient from a list by using the user interface. Up until now, this test did this by having a function to directly call the API. In these two PRs - https://github.com/openmrs/openmrs-esm-core/pull/1280, https://github.com/openmrs/openmrs-esm-core/pull/1290, this test started failing because the API call failed for some reason. Which is why I tried to resort to using the UI, because we can and we should anyway.

NethmiRodrigo:
@jayasanka-sack would appreciate your input on this

samuelmale:
The issue could be related to time zones (client vs backend); @NethmiRodrigo try voiding the membership explicitly [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-list-management-app/src/api/api-remote.ts#L183-L185):

```ts
return postData(`${cohortUrl}/cohortmember/${cohortMembershipUuid}`, {
    endDate: new Date(),
    voided: true
  });
```

NethmiRodrigo:
> The issue could be related to time zones (client vs backend); @NethmiRodrigo try voiding the membership explicitly [here](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-list-management-app/src/api/api-remote.ts#L183-L185):
> 
> ```ts
> return postData(`${cohortUrl}/cohortmember/${cohortMembershipUuid}`, {
>     endDate: new Date(),
>     voided: true
>   });
> ```

@samuelmale sorry I don't understand. Since the delete button calls this function [on click](https://github.com/openmrs/openmrs-esm-patient-management/blob/3d41c0bce421c938006cfdb5cc7447264beab172/packages/esm-patient-list-management-app/src/list-details-table/list-details-table.component.tsx#L214), doesn't this already happen?

denniskigen (approved):
This is impossible to test after we’ve removed the Patient list related frontend modules from the distro, but I’m confident these changes put us in a better position for when we restore that functionality in the future.

NethmiRodrigo:
Oh right, yeah I completely forgot that we removed that. Do you think it's better to just close this PR @denniskigen

denniskigen:
I think we can merge it in and sort things out when we restore the two modules in the future.

denniskigen:
Oh wait, both frontend modules still exist in the e2e test docker environment. Maybe we should explicitly omit them from inclusion in `spa-assemble-config.json`.

PR #1502, opened 2025-02-25
(feat) O3-4370: Preserve search context when discarding visit form in queues app

Muppasanipraneeth:
## Requirements

## Summary

Currently, when adding a patient to a queue from the service queues app, users are directed to a search pane in the workspace. After entering a search term, a list of matching patients is displayed. From this list, users can select a patient to add to the queue. For patients without an active visit, clicking the "Start Visit" button opens a form to start a new visit. However, if a user clicks the discard button on that form, the entire workspace closes. This forces users to restart the entire process of adding a patient to the queue.

This PR modifies the discard button behavior to keep the workspace open and return users to the search pane with their previous search results still visible. This allows users to easily select the same patient again or choose a different patient without repeating the queue addition process.

### Before

https://github.com/user-attachments/assets/9ae6bc1e-bf68-4866-938d-caf4f82f53e9

### After

https://github.com/user-attachments/assets/2b5b5ff6-d128-4946-be00-53c5faf6f18b

## Related Issue

https://openmrs.atlassian.net/browse/O3-4370


Muppasanipraneeth:
[patient chart pr](https://github.com/openmrs/openmrs-esm-patient-chart/pull/2265)

denniskigen (approved):
LGTM. Thanks, @Muppasanipraneeth. Any qualms, @chibongho, @donaldkibet?

denniskigen:
@Muppasanipraneeth @chibongho shouldn't we extend this change to the [ExistingVisitForm] component (https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/create-queue-entry/existing-visit-form/existing-visit-form.component.tsx#L19) too?

denniskigen:
So, I've just noticed that there's an issue with how this `handleBackToSearchList` function works. From it's 
[implementation](https://github.com/muppasanipraneeth/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx#L41-L42), it seems that when invoked, the handler is meant to:

1. `setIsPatientSearchOpen(true)` - this reopens the patient search interface by setting its visibility state to true.
2. `closeWorkspace(serviceQueuesPatientSearchWorkspace) - This closes the service queues workspace that was opened for adding a patient to the queue.

From my testing, it appears that the function only works the first time round, after which subsequent triggers don't seem to do anything.

chibongho:
Yeah... this is a tricky one. FWIW, I hit a similar issue with the ward app: https://openmrs.atlassian.net/browse/O3-4300

packages/esm-patient-search-app/src/patient-search-context.provider.ts
ibacher, line 1: @denniskigen Shouldn't we keep this as `patient-search-context` since it exports both the Provider _and_ the hook.

ibacher (approved):

denniskigen:
Thanks, @Muppasanipraneeth!

PR #1504, opened 2025-02-26
(fix) Standardize appointment component props and fix filter behaviour

denniskigen:
## Requirements

## Summary

Relates to https://github.com/openmrs/openmrs-esm-patient-management/pull/1449.

Makes the following refactors to the Appointments app:

- Rename `appointmentServiceType` to `appointmentServiceTypes` in components that use it. Prior to https://github.com/openmrs/openmrs-esm-patient-management/pull/1449, this prop was a string, but is now an array of strings. The refactor also updates the components to handle the new prop type.
- Rename the `filterCancelled` prop to `excludeCancelledAppointments` in components that use it. The latter is more descriptive of the purpose of the prop.
- Fix `hasActiveFilters` logic to only consider service type filters. There's a tangential issue where the Today's appointments component shows an empty filter state when it shouldn't. It should show an empty state instead if there are no scheduled appointments for today. Cleaning up this logic allows me to isolate that issue so it can be fixed in a separate PR.

Additionally, following this [discussion](https://github.com/openmrs/openmrs-esm-patient-management/pull/1504#discussion_r1971535142), this PR modifies the appointments service type filter. It removes the `initialSelectedItem` prop from the `MultiSelect` component, which previously auto-selected the first available service type from the backend. With this change, the filter starts in a blank state, showing all appointments by default. Users must now explicitly select which service types they want to filter by, making the filter behavior more intuitive and predictable.

### No pre-selected filters

CleanShot 2025-02-26 at 8  52 01@2x

## Related Issue
https://openmrs.atlassian.net/browse/O3-4335

### DIFF
--- a/packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
@@ -4,35 +4,39 @@ import { useAppointmentList } from '../../hooks/useAppointmentList';
 import AppointmentsTable from '../common-components/appointments-table.component';
 
 interface AppointmentsListProps {
-  appointmentServiceType?: string[];
+  appointmentServiceTypes?: string[];
+  date: string;
+  excludeCancelledAppointments?: boolean;
   status?: string;
   title: string;
-  date: string;
-  filterCancelled?: boolean;
 }
+
 const AppointmentsList: React.FC<AppointmentsListProps> = ({
-  appointmentServiceType,
+  appointmentServiceTypes,
+  date,
+  excludeCancelledAppointments = false,
   status,
   title,
-  date,
-  filterCancelled = false,
 }) => {
   const { appointmentList, isLoading } = useAppointmentList(status, date);
 
-  const appointments = filterByServiceType(appointmentList, appointmentServiceType).map((appointment) => ({
-    id: appointment.uuid,
-    ...appointment,
-  }));
+  const appointmentsFilteredByServiceType = filterByServiceType(appointmentList, appointmentServiceTypes).map(
+    (appointment) => ({
+      id: appointment.uuid,
+      ...appointment,
+    }),
+  );
 
-  const activeAppointments = filterCancelled
-    ? appointments.filter((appointment) => appointment.status !== 'Cancelled')
-    : appointments;
+  const activeAppointments = excludeCancelledAppointments
+    ? appointmentsFilteredByServiceType.filter((appointment) => appointment.status !== 'Cancelled')
+    : appointmentsFilteredByServiceType;
+
   return (
     <AppointmentsTable
       appointments={activeAppointments}
+      hasActiveFilters={appointmentServiceTypes?.length > 0}
       isLoading={isLoading}
       tableHeading={title}
-      hasActiveFilters={appointmentServiceType?.length > 0 || filterCancelled}
     />
   );
 };
--- a/packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import { filterByServiceType } from '../utils';
 import { useAppointmentList } from '../../hooks/useAppointmentList';
 import AppointmentsTable from '../common-components/appointments-table.component';
 
 interface AppointmentsListProps {
-  appointmentServiceTypes?: string[];
+  appointmentServiceTypes?: Array<string>;
   date: string;
   excludeCancelledAppointments?: boolean;
   status?: string;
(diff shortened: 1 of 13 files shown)


denniskigen:
FYI, @Samstar10.

packages/esm-appointments-app/src/hooks/useClinicalMetrics.ts
```diff
@@ -46,36 +45,41 @@ export function useAllAppointmentsByDate() {
...
     error,
+    isLoading,
     isValidating,
     mutate,
+    totalProviders: providersCount ? providersCount : 0,
   };
 }
 
-export const useScheduledAppointment = (serviceUuid: string[]) => {
+export const useScheduledAppointments = (appointmentServiceTypeUuids: string[]) => {
```
denniskigen, line 61: Renamed to plural for clarity.

packages/esm-appointments-app/src/hooks/useClinicalMetrics.ts
```diff
@@ -46,36 +45,41 @@ export function useAllAppointmentsByDate() {
...
 
-  const { data, error, isLoading } = useSWR<{
-    data: Array<any>;
-  }>(url, openmrsFetch);
+  const { data, error, isLoading } = useSWR<
+    {
+      data: Array<Appointment>;
+    },
+    Error
+  >(url, selectedDate ? openmrsFetch : null);
```
denniskigen, line 70: Makes it so that we make the request only if `selectedDate` exists.

packages/esm-appointments-app/src/hooks/useClinicalMetrics.ts
```diff
@@ -46,36 +45,41 @@ export function useAllAppointmentsByDate() {
...
 
-  const totalScheduledAppointments = !isEmpty(serviceUuid)
-    ? data?.data?.filter((appt) => serviceUuid.includes(appt?.service?.uuid))?.length ?? 0
-    : data?.data?.length ?? 0;
+  const appointments = data?.data ?? [];
+
+  const totalScheduledAppointments =
+    appointmentServiceTypeUuids.length > 0
```
denniskigen, line 75: We don't need `lodash.isEmpty` to do this check.

Samstar10:
> FYI, @Samstar10.

Thank you @denniskigen

packages/esm-appointments-app/src/hooks/useClinicalMetrics.ts
```diff
@@ -37,45 +36,59 @@ export const useClinicalMetrics = () => {
   };
 };
 
-export function useAllAppointmentsByDate() {
+export const useAppointmentsForDate = () => {
```
denniskigen, line 39: This shared hook abstracts away the API call to `/ws/rest/v1/appointment/all?forDate=${selectedDate}` and gets reused in `useAllAppointmentsByDate` and `useScheduledAppointments`.

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -55,11 +55,12 @@ const AppointmentsHeader: React.FC<AppointmentHeaderProps> = ({ title, appointme
         {typeof onChange === 'function' && (
           <MultiSelect
             id="serviceTypeMultiSelect"
-            label={t('filterAppointmentsByServiceType', 'Filter appointments by service type')}
+            // TODO: Figure out why we need to set the initial selected items
+            initialSelectedItems={serviceTypeOptions.length > 0 ? [serviceTypeOptions[0].id] : []}
```
denniskigen, line 59: @Samstar10 I don't understand why we need to initialize the filter to the first appointment service type fetched from the backend. Doesn't it make more sense to not have the filter set at all on initial launch - which to me means the same as "show all appointments"?

Samstar10: Yes Dennis it makes more sense to not have a filter set at all on initial launch. This was an oversight on my end.

denniskigen: Resolved in the latest commit. Thanks, @Samstar10!

denniskigen:
@NethmiRodrigo can we cut a patch and include this fix?

packages/esm-appointments-app/src/appointments/scheduled/appointments-list.component.tsx
```diff
@@ -1,38 +1,44 @@
...
-  const appointments = filterByServiceType(appointmentList, appointmentServiceType).map((appointment) => ({
-    id: appointment.uuid,
-    ...appointment,
-  }));
+  const appointmentsFilteredByServiceType = filterByServiceType(appointmentList, appointmentServiceTypes).map(
+    (appointment) => ({
+      id: appointment.uuid,
+      ...appointment,
+    }),
+  );
```
NethmiRodrigo, line 28: Question (purely for me to learn how to use `useMemo` better than just putting it everywhere xD): 
I'm wondering if wrapping `appointmentsFilteredByServiceType` in its own useMemo would help. My concern is that if `appointmentsFilteredByServiceType` recalculates on every render (creating a new array reference each time), it would trigger `activeAppointments` to recalculate unnecessarily.
But if the props `appointmentList` and `appointmentServiceTypes` change on every render anyway, is there any benefit to using `useMemo` for `appointmentsFilteredByServiceType`?

packages/esm-appointments-app/src/header/appointments-header.component.tsx
```diff
@@ -10,11 +10,11 @@ import styles from './appointments-header.scss';
 
 interface AppointmentHeaderProps {
   title: string;
-  appointmentServiceType?: string[];
+  appointmentServiceTypes?: string[];
```
NethmiRodrigo, line 13: nit:
```suggestion
  appointmentServiceTypes?: Array<string>;
```

NethmiRodrigo (approved):
Thanks @denniskigen!

NethmiRodrigo (approved):
LGTM!

PR #1497, opened 2025-02-20
(feat) O3-4400 - render queue entry status in patient banner

chibongho:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

As part of the work to [remove the custom visit header](https://openmrs.atlassian.net/browse/O3-4361) from the patient chart, we need to move the queue entry status indicator from the visit header to the patient banner.

See codependent PR in patient-chart: https://github.com/openmrs/openmrs-esm-patient-chart/pull/2254

Note that this PR does not confirm to [the mockup](https://app.zeplin.io/project/60d59321e8100b0324762e05/screen/679761bd13becda18acdf2cd). It's only contains the work to move the functionality into the patient banner.

## Screenshots

image
(The same stuff that's in the visit header will be removed in a separate PR)

PR #1505, opened 2025-02-27
Revert "(feat) O3-4400 - render queue entry status in patient banner"

NethmiRodrigo:
Reverts openmrs/openmrs-esm-patient-management#1497


NethmiRodrigo:
Looking at the visit header and the patient banner, right now the queue entry status would appear in both, so opting to revert this for now:
image
I'll make a revive PR for this again to get in after the release

denniskigen (approved):

PR #1507, opened 2025-02-27
(feat) O3-4483: Display patient phone number as column in patient list view

Muppasanipraneeth:
## Requirements

## Summary

- Add a column to the Patient Lists that displays the Patient’s phone number (if known).

## Screenshots

Screenshot 2025-02-27 at 11 34 50 AM


denniskigen (approved):
LGTM. Thanks, @Muppasanipraneeth!

mseaton:
Just happened to notice this commit @denniskigen .  "Telephone Number" is not a standard OpenMRS attribute that will be present (or named this way) in every implementation.  At minimum I would think this needs to be configurable?

denniskigen:
Good shout, @mseaton. That assumption is [being made](https://github.com/search?q=org:openmrs+%3D%3D%3D+%22Telephone+Number%22+language:TSX&type=code&l=TSX) in more than one place. What's a more robust check?

denniskigen:
How can I query for `emr. telephoneAttributeType` from the client? It's not exposed via systemsetting AFAICT, @mseaton.

mseaton:
@denniskigen - still, it's a good point.  Although I see methods in emrapi to support this global property, this property is not explicitly defined in config.xml (which is why it doesn't show up when querying the endpoint, unless on a system that has explicitly created it).  And I don't actually see anywhere in our PIH configurations where we set this global property or use it.  Still...this does exist and we could start to use it, though we could also look at alternatives.  I do think this kind of thing is best defined on the backend, ideally.

mogoodrich:
> @denniskigen - still, it's a good point. Although I see methods in emrapi to support this global property, this property is not explicitly defined in config.xml (which is why it doesn't show up when querying the endpoint, unless on a system that has explicitly created it). And I don't actually see anywhere in our PIH configurations where we set this global property or use it. Still...this does exist and we could start to use it, though we could also look at alternatives. I do think this kind of thing is best defined on the backend, ideally.

Thanks @mseaton ... I guess I should have checked this comment before doing all that research myself as well to comment on the other PR... :)

PR #1518, opened 2025-03-10
(refactor) Standardize appointments form workspace and naming

denniskigen:
## Requirements

## Summary

This change refactors the Appointments app as follows:

- Renames the `AppointmentsForm` component file name to `appointments-form.workspace.tsx` to align with our naming conventions.
- Moves workspace registrations in the `routes.json` file from using the `extensions` key to using the `workspace` key as is standard practice for workspace registrations.
- Removing the `create-appointment` workspace in favor of the similar `create-appointments-form-workspace`. Both seemed to be doing the same thing. Removing the duplication adds clarity.
- Renaming the `PatientCancelAppointmentModal` component to `CancelAppointmentModal` to align with our naming conventions.

These changes standardize the naming conventions for the Appointments app and make the codebase more consistent.

##### Create

CleanShot 2025-03-10 at 5  30 41@2x

##### Edit

CleanShot 2025-03-10 at 5  30 21@2x

### Likewise for translations

CleanShot 2025-03-10 at 5  29 19@2x

CleanShot 2025-03-10 at 5  29 05@2x


packages/esm-appointments-app/src/routes.json
```diff
@@ -148,8 +107,22 @@
...
+      "name": "cancel-appointment-modal",
+      "component": "cancelAppointmentModal"
+    }
+  ],
+  "workspaces": [
+    {
+      "name": "create-appointments-form-workspace",
+      "component": "appointmentsFormWorkspace",
+      "title": "createNewAppointment",
+      "type": "form"
```
denniskigen, line 119: @vasharma05 @ibacher what should this type prop map to in this case where we're using the same form in two different modes - `create` and `edit` mode.

ibacher: Seems like it should just be a single workspace and we pass some kind of prop(s) to tell it whether we're editing it or not? I don't think we want to use type to distinguish that.

denniskigen: Makes sense. Any qualms with that, @chibongho?

chibongho: That sounds fine. We do something similar with the Visit form in patient chart.

chibongho (approved):

PR #1519, opened 2025-03-10
(fix)O3-3580: Input validation errors in the registration page form stop the form from submitting any data

Pratibha0280:
## Requirements

- [x]  If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.

## Summary
Ensure proper validation handling by blocking the submission process at the backend.

## Screenshots

After Fix: 
[Screencast from 2025-03-10 20-33-17.webm](https://github.com/user-attachments/assets/6e139a53-f243-4ffc-aeaf-17abfc954212)

## Related Issue


packages/esm-patient-registration-app/src/patient-registration/patient-registration.component.tsx
```diff
@@ -70,8 +70,12 @@ export const PatientRegistration: React.FC<PatientRegistrationProps> = ({ savePa
   const onFormSubmit = async (values: FormValues, helpers: FormikHelpers<FormValues>) => {
     const abortController = new AbortController();
     helpers.setSubmitting(true);
-
     const updatedFormValues = { ...values, identifiers: filterOutUndefinedPatientIdentifiers(values.identifiers) };
+    const errors = await helpers.validateForm();
+    if (Object.keys(errors).length > 0) {
```
yoursanonymous, line 75: The errors variable could be received as null as well. The good practice would be to add null check before doing any computation on the variable just like how we are doing computation on the variable to find length.
The better condition should have been: if(errors!=null && Object.keys(errors).length>0)

Pratibha0280: Thankyou for the review @yoursanonymous 
Formik's validateForm has been used here, which returns an empty object {} in case of no errors instead of null.

denniskigen:
Thanks for taking a stab at this, @Pratibha0280. Are you able to replicate the reported issue on dev3? I note that the original repro video uses a different instance than the ones we maintain.

denniskigen:
As mentioned in the issue, we fixed this in https://github.com/openmrs/openmrs-esm-patient-management/pull/1553 a while ago. What'd be worth figuring out is whether the ICAP Ethiopia team have updated their implementation to include the fix. Closing this as stale for now. Feel free to reopen if the issue is still reproducible after incorporating the fix.

PR #1553, opened 2025-04-15
(feat) Validate identifier regex patterns in the registration form

denniskigen:
## Requirements

## Summary

The [IdentifierInput component](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx) in the Registration form should validate identifier values against their defined regex patterns as the user types.

The component should use the [format and formatDescription](https://rest.openmrs.org/#list-patientidentifiertype-resource) properties (when available) to perform real-time validation and provide immediate feedback to users. Currently, when a user enters an invalid identifier value, the UI only shows validation errors after form submission, displaying the backend error message.

This change will improve the user experience by providing immediate visual feedback and clear error messages about format requirements after typing into the identifier field.

### Before

CleanShot 2025-04-15 at 10  03 38@2x

### After

CleanShot 2025-04-15 at 2  30 44@2x

## Related Issue
https://openmrs.atlassian.net/browse/O3-4644
https://openmrs.atlassian.net/browse/O3-3580

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx
+++ b/packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx
@@ -57,9 +57,8 @@ const IdentifierInput: React.FC<IdentifierInputProps> = ({ patientIdentifier, fi
       if (regex.test(value)) {
         return;
       }
-      return t('invalidIdentifierFormat', '{{description}}', {
-        description: identifierType.formatDescription || `Expected format: ${identifierType.format}`,
-      });
+
+      return identifierType.formatDescription ?? `Expected format: ${identifierType.format}`;
     } catch (e) {
       console.error('Invalid regex pattern:', identifierType.format);
       return;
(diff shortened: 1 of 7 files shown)


packages/esm-patient-registration-app/src/patient-registration/input/custom-input/identifier/identifier-input.component.tsx
```diff
@@ -43,6 +43,29 @@ const IdentifierInput: React.FC<IdentifierInputProps> = ({ patientIdentifier, fi
...
+    if (!identifierType?.format) {
+      return;
+    }
+
+    try {
+      const regex = new RegExp(identifierType.format);
+      if (regex.test(value)) {
+        return;
+      }
+      return t('invalidIdentifierFormat', '{{description}}', {
```
ibacher, line 60: The only thing I question about this is this part of the PR, specifically using the translation function here. I'm struggling to see who can leverage this. On the one hand, no one can do anything useful with translating this in Transifex since it's just string interpolation. Presumably, an implementation could override the translation, but even then, they'd want something that varies depending on the identifier type.

I think for this one, we just return the `identifierType.formatDescription` and defer translating that to the backend. We could then probably provide a useful frontend translation for the fallback text, namely, `Expected format: {{identifierTypeFormat}}`.

denniskigen:
Updated:

CleanShot 2025-04-17 at 2  07 36@2x

ibacher (approved):

PR #1540, opened 2025-04-01
(fix) O3-4516: Use narrow width variant for appointments and ward workspaces

EDSONZ-WASSWA:
## Requirements
I have made sure that the “Create new appointment“ workspace, the “Add patient to queue“ workspace and the “Admission requests“ workspaces now use the “normal“ workspace size.
These widgets now match the correct size as the ticket describes
screenshots
before:
Appointment-before
Admission-request-before
Add patient-before
After:
Appointment-after
Admission-request-after
Add-patient-after
Ticket-Link:
[[ticket](https://openmrs.atlassian.net/browse/O3-4516l)](https://openmrs.atlassian.net/browse/O3-4516l)
@denniskigen  review this please

PR #1586, opened 2025-05-05
(fix) O3-4516 Ward App - Make admission request workspace wider

chibongho:
## Requirements

## Summary

A [previous commit](https://github.com/openmrs/openmrs-esm-patient-management/pull/1540) changed the patient-search workspace's and the admission request workspace's widths to be standard (narrow). However. per the [mockup](https://app.zeplin.io/project/65f454477b16814668c38d02/screen/65fda236f0ad0df6e5d786ec), the admission request workspace should be wider than the standard workspace width in order to display the buttons in the pending admission patient cards in one row.

 In the mockup, the workspace is 420px wide. We don’t actually have that exact size, but it’s pretty close to the “wider” workspace size.

## Screenshots

Before:
image

After:
image

## Related Issue

https://issues.openmrs.org/browse/O3-4516

### DIFF
--- a/packages/esm-ward-app/src/routes.json
+++ b/packages/esm-ward-app/src/routes.json
@@ -92,7 +92,7 @@
       "component": "admissionRequestWorkspace",
       "title": "admissionRequests",
       "type": "pending-admission-requests",
-      "width": "narrow"
+      "width": "wider"
     },{
       "name": "create-admission-encounter-workspace",
       "component": "createAdmissionEncounterWorkspace",


mseaton (approved):

NethmiRodrigo (approved):

NethmiRodrigo:
@denniskigen if there would be a patch of patient management that would go in, this seems like a good fix to be included.

ibacher:
I'm going to merge this in for now, but I think we should regard this as a temporary visual fix rather than a long-term one. Long-term, the "narrow" workspace is supposed to be the same size as the designs, so I'm worried that either that definition needs to be widened a bit or we need to slightly rethink the design of the request card. It may be that we need to, e.g., break up the text into multiple lines so it can really work as-designed.

FWIW, the "wider" workspace is solely intended to handle something like the internal navigation in the form engine and I don't love having it used in other places, as it means we lose about 100px of usable space, which can get a little compressed on some of the screen sizes we aim to support.

chibongho:
Maybe we can do something like moving the cancel button into the top right corner of the card, and turning it into a trash can icon.

PR #1592, opened 2025-05-06
(feat) Use dayjs to format time in ward patient cards

denniskigen:
## Requirements

## Summary

Replaces the `age` function with `dayjs` (using the `relativeTime` plugin) to format "time on ward" and "time since admission". The previous `age` function did not handle timezones correctly, causing the displayed time to be offset by the user's timezone. Using `dayjs.fromNow()` ensures accurate, user-localized relative times in the UI.

Users will now see correct durations regardless of their timezone.

### Before
Screenshot 2025-05-07 at 07 41 42

### After
Screenshot 2025-05-07 at 07 41 16

## Other

Slack discussion https://openmrs.slack.com/archives/C06P2GVUGEP/p1743497996163409.


packages/esm-ward-app/translations/en.json
```diff
@@ -20,6 +20,8 @@
   "clinicalNotes": "Clinical notes",
   "countItems_one": "{{count}} {{item}}",
   "countItems_other": "{{count}} {{item}}",
+  "days_one": "{{count}} days",
```
denniskigen, line 23: I can't work out why the translation strings use the wrong plural suffixes

packages/esm-ward-app/src/ward-workspace/admit-patient-form-workspace/admit-patient-form.test.tsx
```diff
@@ -115,7 +114,7 @@ describe('Testing AdmitPatientForm', () => {
           results: [
             {
               bedId: 1,
-              bedNumber: 1,
+              bedNumber: '1',
```
denniskigen, line 117: I'm surprised that this is typed as a string. Is that intentional?

packages/esm-ward-app/src/ward-workspace/admit-patient-form-workspace/admit-patient-form.test.tsx
```diff
@@ -195,7 +194,7 @@ describe('Testing AdmitPatientForm', () => {
     expect(mockedAssignPatientToBed).toHaveBeenCalledWith(3, mockPatientAlice.uuid, 'encounter-uuid');
     expect(mockedShowSnackbar).toHaveBeenCalledWith({
       kind: 'success',
-      subtitle: '{{patientName}} has been successfully admitted and assigned to bed bed3',
+      subtitle: 'Alice Johnson has been successfully admitted and assigned to bed bed3',
```
denniskigen, line 197: This is owing to the interpolation syntax changes in the React i18next mock that align it much closer to the actual implementation.

__mocks__/react-i18next.js
```diff
@@ -29,25 +29,25 @@ const renderNodes = (reactNodes) => {
...
+  return translatedString;
 };
-useMock.i18n = {};
+
+useMock.i18n = { language: 'en_US' };
 
 module.exports = {
   // this mock makes sure any components using the translate HoC receive the t function as a prop
-  Trans: ({ children }) => renderNodes(children),
+  Trans: ({ children }) => (Array.isArray(children) ? renderNodes(children) : renderNodes([children])),
```
denniskigen, line 48: I've just carried over these changes from a similarly named manual mock in Patient Chart to get the interpolation behaviour in the test environment closer in line with the actual implementation.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.component.tsx
```diff
@@ -11,12 +14,38 @@ const WardPatientTimeOnWard: React.FC<WardPatientTimeOnWardProps> = ({
...
+  const days = Math.floor(dur.asDays());
+  const hours = dur.hours();
+  const minutes = dur.minutes();
+  const seconds = dur.seconds();
+
+  const parts = [];
+  // TODO: Fix translations to use the correct pluralization.
+  // Keys and strings extracted from translations below don't use the correct plural suffixes.
+  if (days > 0) {
+    parts.push(t('days', '{{count}} days', { count: days }));
```
NethmiRodrigo, line 32: Looking at their [tutorial](https://youtu.be/SA_9i4TtxLQ?si=44WWfsUnsgn2tVF7&t=485), I think we're not supposed to have a default string.
```suggestion
    parts.push(t('days', { count: days }));
```
I tested locally and it seems to work (I did change the translation strings in `en.json` manually to have `"days_one": "{{count}} day"` ):
Screenshot 2025-05-07 at 1 01 43 PM

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.component.tsx
```diff
@@ -11,12 +14,38 @@ const WardPatientTimeOnWard: React.FC<WardPatientTimeOnWardProps> = ({
...
+
+  if (!encounterAssigningToCurrentInpatientLocation?.encounterDatetime) {
+    return null;
+  }
+
+  const dur = dayjs.duration(dayjs().diff(encounterAssigningToCurrentInpatientLocation.encounterDatetime));
+  const days = Math.floor(dur.asDays());
+  const hours = dur.hours();
+  const minutes = dur.minutes();
+  const seconds = dur.seconds();
```
NethmiRodrigo, line 26: @denniskigen a question since I tried to get it working on the `vitals overdue` tag component as well. The `relativeTime` plugin's `fromNow` [function](https://day.js.org/docs/en/display/from-now) automatically gives a string that says `x hours ago` or `x minutes ago` right? They say its localized, but I couldn't figure out how we can add translations for it.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.component.tsx
denniskigen: I haven't yet determined why we need to manually change the default value in `en.json` to get pluralization working correctly. As a last resort, we could add a comment block with hardcoded key and default value pairings that have the correct suffixes, but I'd prefer to avoid that approach. I was hoping someone might have encountered this before and has a straightforward solution.

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-time-on-ward.component.tsx
denniskigen: To get translations to work with `fromNow`, we'd need to import the target [locales](https://day.js.org/docs/en/durations/locale#docsNav) into the component. Something like:

```tsx
// existing bits omitted for brevity
import relativeTime from 'dayjs/plugin/relativeTime';
import updateLocale from 'dayjs/plugin/updateLocale';
import { getLocale } from '@openmrs/esm-framework';

dayjs.extend(updateLocale);

const WardPatientTimeSinceAdmission: React.FC<WardPatientTimeSinceAdmissionProps> = ({
  firstAdmissionOrTransferEncounter,
}) => {
  // Get the current locale from the framework
  const locale = getLocale();
  // Update dayjs locale
  dayjs.locale(locale);

 // ... existing code
}


Then, when you change the locale to `fr`, you'd get this:

<img width="298" alt="Screenshot 2025-05-07 at 12 07 23" src="https://github.com/user-attachments/assets/f59fef7d-6edd-4c39-9826-25f250cb9449" />

Admittedly, this isn't great for a couple of reasons:

 1. The translation strings are controlled by dayjs, which means we'd have to override them using a comment block if we want more granular control through Transifex.
 2. Manually importing the target locales is unwieldy and antithetical to how we manage i18n everywhere else.

NethmiRodrigo (approved):

ibacher (approved):

PR #1628, opened 2025-05-20
(fix) O3-4662: Labs banner should only include active orders

mseaton:
See:  https://issues.openmrs.org/browse/O3-4662


chibongho (approved):
LGTM

denniskigen (approved):
Thanks!

PR #1634, opened 2025-05-21
(feat) O3-4734 Make Ward / Queues / Appointments apps available as extensions

chibongho:
## Requirements

## Summary

Currently in the refapp, the Ward / Queues / Appointments apps are not standalone apps, Rather, they are mounted inside the home app. For distributions (like PIH’s) that want to have these apps mounted standalone, we need to make those apps, at the top level, be extensions.

Annoyingly, while each of those apps already has a `root` component as an extension, we cannot use it as it includes its own `<BrowserRouter>`. This PR exposes the (almost) top-level components that those `BrowserRouters` route to.

### DIFF
--- a/packages/esm-service-queues-app/src/views/queue-table-by-status-view.component.tsx
+++ b/packages/esm-service-queues-app/src/views/queue-table-by-status-view.component.tsx
@@ -1,19 +1,16 @@
 import React from 'react';
-import { useParams } from 'react-router-dom';
 import { useQueue } from '../hooks/useQueue';
 import QueueTablesForAllStatuses from './queue-tables-for-all-statuses.component';
 
 interface QueueTableByStatusViewProps {
-  /**
-   * If provided, this will be used to fetch the queue instead of the one in the URL.
-   * This is useful for when this component is used as an extension.
-   */
-  customQueueUuid?: string;
+  queueUuid: string;
 }
 
-const QueueTableByStatusView: React.FC<QueueTableByStatusViewProps> = ({ customQueueUuid }) => {
-  const { queueUuid } = useParams();
-  const { queue, isLoading: isLoadingQueue, error } = useQueue(customQueueUuid ?? queueUuid);
+/**
+ * This component renders the several tables, one for each status, for a given queue.
+ */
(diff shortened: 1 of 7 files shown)


packages/esm-service-queues-app/src/views/queue-table-by-status-view.component.tsx
```diff
@@ -3,9 +3,17 @@ import { useParams } from 'react-router-dom';
...
+   * If provided, this will be used to fetch the queue instead of the one in the URL.
+   * This is useful for when this component is used as an extension.
+   */
+  customQueueUuid?: string;
+}
+
+const QueueTableByStatusView: React.FC<QueueTableByStatusViewProps> = ({ customQueueUuid }) => {
   const { queueUuid } = useParams();
-  const { queue, isLoading: isLoadingQueue, error } = useQueue(queueUuid);
+  const { queue, isLoading: isLoadingQueue, error } = useQueue(customQueueUuid ?? queueUuid);
```
mseaton, line 16: This seems odd.  Might need to discuss this...

chibongho: `useParams()` is a feature of `react-router-dom`. It gets a param in the URL, like `/service-queues/:queueUuid`. For this to work, this component needs to be rendered within the same React DOM tree as where the `<BrowserRouter>` is defined. However, this is not the case when the component is used as an extension.

One less awkward way to implement this would be to create a thin component that reads the `useParams()` value and pass it to this component, and have this component always use the passed in `queueUuid`.

mseaton: Yes, that sounds better.  As implemented here, you could go to a url that contains a queue uuid, and it wouldn't render the queue with that uuid.  I'd think we want to aggressively avoid that.  I'd prefer if we always just pass in the queue uuid as a prop as you say.  Then, depending on the rendering context, we can either set the value of that prop automatically from useParams, or set it manually as needed.

mseaton (approved):
Nice!  Thanks @chibongho

PR #1646, opened 2025-05-30
(fix) Change null arg to timestamp in date constructor

princeavo:
This commit updates the `null` argument passed to the `Date` constructor
 to the `0` timestamp value as there is no version of the constructor
 accepting `null` values.


jwnasambu:
@princeavo Could you please attach the relevant ticket link to this PR? Having the ticket linked helps with traceability and makes it easier for others to follow the context of the change. Thanks

denniskigen (commented):
Thanks for working on this, @princeavo. Unfortunately, it looks like this utility file isn't used anywhere in the app. My recommendation is that we remove the file altogether.

denniskigen:
Utility file deleted [here](https://github.com/openmrs/openmrs-esm-patient-management/commit/c4257db32eea712fb6be7622686eb2a9925f32e8).

PR #1649, opened 2025-06-02
O3 4757 E2E Add automated testing for canceling an admission request

EDSONZ-WASSWA:
For the test, requires being in the inpatient location and have  one patient in the Admission request list.

test-screenshort
[test-vid.webm](https://github.com/user-attachments/assets/e571bae7-cbbc-4143-ab44-c4454f63740d)


jwnasambu:
@EDSONZ-WASSWA, as @gitcliff pointed out above, the failing job is caused by an issue with the deletePatient function. Although it is imported in the e2e/core/test.ts file, it's not recognized as a valid function. The error logs show:
TypeError: (0, _commands.deletePatient) is not a function,
which suggests that the deletePatient function is either not exported correctly or is improperly imported.

How to fix the issue:
Verify deletePatient export
Check the commands module to ensure that the deletePatient function is correctly exported. It might be missing or not exported as expected.

// Ensure this exists in commands.ts or commands.js
export const deletePatient = async (api, uuid) => {
  // Implementation for deleting a patient
};
Correct the Import in test.ts
Ensure the import matches the export type from the commands file.

If deletePatient is a named export:

import { deletePatient } from '../commands';
If it's a default export:

import deletePatient from '../commands';
Make sure the export and import styles are consistent to resolve the error.

yarn.lock
```diff
@@ -3388,14 +3388,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@playwright/test@npm:^1.50.1":
-  version: 1.50.1
-  resolution: "@playwright/test@npm:1.50.1"
+"@playwright/test@npm:^1.52.0":
```
denniskigen, line 3391: Bumping the `@playwright/test` package requires a [commensurate bump to the Dockerfile](https://github.com/openmrs/openmrs-esm-patient-management#updating-playwright). Those two versions have to match.

PR #1546, opened 2025-04-03
(BREAKING) O3-4469: Move queue priority configuration in service queues to top level config

jwnasambu:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary
I refactored the code to move the priorityConfigs configuration outside of the queueTables.columnDefinitions section and defining it at the top level. These changes help simplify the column definition logic and make the priority color configuration more accessible and easier to manage.

## Screenshots

https://github.com/user-attachments/assets/dcbb47a4-a80a-43cc-aa53-00730c53dc51

## Related Issue
https://openmrs.atlassian.net/browse/O3-4469

PR #1657, opened 2025-06-10
(BREAKING) O3-4790: Standardize config schemas and improve type safety

harshthakkr:
## Requirements

## Summary
This PR standardizes the config schema files in the `esm-patient-management` repository by:

- Adding proper types and default values
- Defining appropriate validators where needed
- Updating the corresponding TypeScript interfaces to reflect these changes

## Related Issue
[O3-4790](https://openmrs.atlassian.net/browse/O3-4790)

[O3-4790]: https://openmrs.atlassian.net/browse/O3-4790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


harshthakkr:
I don't know why the tests fail, I haven't made any changes to those test files.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -320,11 +331,11 @@ export const esmPatientRegistrationSchema = {
...
-        useAddressHierarchyLabel: {
-          _type: Type.Object,
-          _description: 'Whether to use custom labels for address hierarchy',
-          _default: {},
-        },
+        // useAddressHierarchyLabel: {
+        //   _type: Type.Object,
+        //   _description: 'Whether to use custom labels for address hierarchy',
+        //   _default: {},
+        // },
```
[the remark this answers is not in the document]
harshthakkr: Pardon me for not clarifying earlier. I don’t have enough information about the object’s structure, such as its properties. However, based on the description, it seems to be a boolean value rather than an object.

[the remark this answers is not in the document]
denniskigen: Looks like it's [not being used anywhere](https://github.com/search?q=repo:openmrs/openmrs-esm-patient-management%20useAddressHierarchyLabel&type=code). So perhaps we should just delete it.

packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -22,7 +22,6 @@ export const configSchema = {
         'Disable the default "keyup search" for instant patient search as typing concludes on tablet devices',
     },
     searchFilterFields: {
-      _type: Type.Object,
```
denniskigen, line 25: Isn't the `type` still valid here?

harshthakkr:
The problem is even on my updated main branch the tests fail.
Screenshot 2025-07-01 at 7 03 32 PM

denniskigen:
You should be seeing the same test failures that occur in the CI locally https://github.com/openmrs/openmrs-esm-patient-management/actions/runs/16000693880/job/45134610562?pr=1657.  Is your local branch up to date with the remote?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -177,12 +182,12 @@ export const configSchema = {
   },
   defaultFacilityUrl: {
     _type: Type.String,
-    _default: '',
+    _default: null,
```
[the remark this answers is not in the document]
denniskigen: So, a few things here:

  1. It looks like this property isn't used anywhere in the code, so we can safely remove it from the config schema.

  2. You're right, when using the `Type.String` validator, `null` isn't a valid default. That's because when the validation logic in [module-config.ts](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-config/src/module-config/module-config.ts#L722) runs, the `isString` validator will error out because this check:

      ```ts
      export const isString = validator((val) => typeof val === 'string', 'must be a string');
      ```
      will return `false`.

  3. In this case, I think there are the following possibilities:

     i. Default to using an empty string:

      ```ts
      defaultFacilityUrl: {
          _type: Type.String,
          _default: '',
          _description: 'Custom URL to load default facility if it is not in the session',
        }
      ```

     ii. Make the property optional by removing the type constraint:

      ```ts
      defaultFacilityUrl: {
        _default: null,
        _description: 'Custom URL to load default facility if it is not in the session',
      }
      ```

     iii. Use undefined instead of null:

      ```ts
      defaultFacilityUrl: {

        _type: Type.String,
        _default: undefined,
        _description: 'Custom URL to load default facility if it is not in the session',
        }
      ```

  4. Based on this property's description, "Custom URL to load default facility if it is not in the session", this property should be optional since it's only used when there's no facility in the session. To that end, I'm in favour of removing the type constraint because:
      - The property is clearly optional - it's only used as a fallback when the facility is not in the session object.
      - Setting `_default: null` without `_type: Type.String` allows the property to be either a string URL or `null`.
      - This matches the semantic meaning: "no default facility URL configured" versus "here is a default facility URL"
      - `null` more clearly represents "no default URL configured" whereas an empty string could be ambiguous.

  5. Maybe there should be an additional type that allows for optional string values. Looking at the existing [Type enum](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-config/src/types.ts#L1), there isn't a built-in type for "string or null/undefined". Absent that, maybe the pattern should be to omit the `_type` for optional properties. IMO, the type system should be primarily used for enforcing required, well-defined value types.

packages/esm-active-visits-app/src/config-schema.ts
```diff
@@ -18,31 +19,64 @@ export interface IdentifiersDefinition {
...
+  };
+}
+
 export const configSchema = {
   activeVisits: {
     identifiers: {
       _type: Type.Array,
       _description: 'Customizable list of identifiers to display on active visits table',
       _elements: {
+        _type: Type.Object,
```
brandones, line 36: No, this is not what `Type.Object` is for. `Type.Object` is for unstructured objects. It works like `Type.Array`, but for objects. It should be avoided when possible, in favor of `Type.Array`.

I almost wonder if we should deprecate it, since one can always define an array config that achieves the same thing:

```
_type: Type.Array,
_elements: {
  key: { _type: Type.String },
  value: { _type: Type.String }
}
```

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -125,6 +125,7 @@ export const esmPatientRegistrationSchema = {
   sectionDefinitions: {
     _type: Type.Array,
     _elements: {
+      _type: Type.Object,
```
brandones, line 128:
```suggestion
```

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -320,11 +335,6 @@ export const esmPatientRegistrationSchema = {
             "Whether to fill the addresses by levels, i.e. County => subCounty, the current field is dependent on it's previous field.",
           _default: false,
         },
-        useAddressHierarchyLabel: {
```
brandones, line 323: Is this unused?

packages/esm-patient-search-app/src/config-schema.ts
brandones: Wow everything about this section of the config schema is totally nuts. I think I need to rewrite the documentation. Something is really not making sense to people. I can't believe this is passing the config system's validation. Maybe I should rewrite that too.

brandones: @harshthakkr could you attempt to fix this section, now that (if you read my other comments) you understand `Type.Object` and the difference between elements and trees? You are correct to get rid of all these `Type.Object`s; none of them are correct. Make sure `_default` is defined at the right level etc.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -195,6 +200,7 @@ export const configSchema = {
       _description:
         "Custom columns for queue tables can be defined here. These columns will be referenced by their `id` in the `tableDefinitions` columns configuration. If the provided `id` matches a built-in column, the custom configuration will override the built-in column's configuration.",
       _elements: {
+        _type: Type.Object,
```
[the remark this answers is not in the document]
brandones: and all the others too

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -306,15 +314,15 @@ export const configSchema = {
...
+                _default: '',
               },
             },
-            visitQueueNumberAttributeUuid: {
-              _type: Type.String,
-              _description:
-                'The UUID of the visit attribute that contains the visit queue number. This must be set to use the queue-number column if the top-level `visitQueueNumberAttributeUuid` config element is not set.',
-              _default: null,
-            },
+          },
```
brandones, line 320: I don't think this is right.

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -144,18 +151,17 @@ export const configSchema: ConfigSchema = {
         },
       ],
       _elements: {
+        _type: Type.Object,
```
brandones, line 154: None of this is right

packages/esm-ward-app/src/ward-patient-card/card-rows/pending-items-row.component.tsx
```diff
@@ -12,7 +12,7 @@ export interface PendingItemsRowProps {
 }
 
 const PendingItemsRow: React.FC<PendingItemsRowProps> = ({ id, wardPatient }) => {
-  const { orders, showPendingItems } = useElementConfig('pendingItems', id);
+  const { orders, showPendingItems } = useElementConfig('pendingItems', id) || {};
```
brandones, line 15: The fallback is probably only necessary because you broke the config schema

packages/esm-ward-app/src/ward-patient-card/row-elements/ward-patient-header-address.component.tsx
```diff
@@ -13,7 +13,7 @@ const WardPatientAddress: React.FC<WardPatientAddressProps> = ({ patient, id })
 
   return (
     <>
-      {config.fields?.map((field, i) =>
+      {config?.fields?.map((field, i) =>
```
brandones, line 16: again

packages/esm-patient-registration-app/src/config-schema.ts
harshthakkr: https://github.com/openmrs/openmrs-esm-patient-management/pull/1657#discussion_r2156770457

packages/esm-service-queues-app/src/config-schema.ts
harshthakkr: According to the type of `config` this is correct: 
```js
export type ColumnConfig = ActionsColumnConfig &
  PatientIdentifierColumnConfig &
  PriorityColumnConfig &
  StatusColumnConfig &
  VisitAttributeQueueNumberColumnConfig;
```

packages/esm-active-visits-app/src/config-schema.ts
```diff
@@ -18,49 +19,89 @@ export interface IdentifiersDefinition {
...
+  };
+}
+
 export const configSchema = {
   activeVisits: {
     identifiers: {
       _type: Type.Array,
-      _description: 'Customizable list of identifiers to display on active visits table',
       _elements: {
+        id: {
```
harshthakkr, line 35: In the respective ts interface, `id` is defined that's why I have added it here.

packages/esm-patient-search-app/src/config-schema.ts
```diff
@@ -4,58 +4,32 @@ export const configSchema = {
...
-        },
-        personAttributes: [
-          {
-            attributeTypeUuid: '14d4f066-15f5-102d-96e4-000c29c2a5d7',
-          },
-        ],
-      },
       gender: {
-        _type: Type.Object,
-        _description: 'Configuration for gender field',
```
harshthakkr, line 49: I have removed `_description` from the config tree elements because it causes the UI to break down, by removing it, it works fine. [Another example](https://github.com/openmrs/openmrs-esm-form-builder/pull/552#discussion_r2208216852)

brandones: Filed https://openmrs.atlassian.net/browse/O3-4898

packages/esm-active-visits-app/src/config-schema.ts
```diff
@@ -18,31 +19,64 @@ export interface IdentifiersDefinition {
...
           _type: Type.String,
-          _default: null,
+          _default: '',
           _description: 'Name of the desired identifier to filter data returned from the visit resource.',
         },
       },
-      _default: null,
+      _default: [],
+    },
+    attributes: {
```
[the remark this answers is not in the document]
brandones: Point me to where he introduced it. Link the blame.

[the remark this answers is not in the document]
harshthakkr: [Blame](https://github.com/harshthakkr/openmrs-esm-patient-management/blame/3242ab8140f39411c8e4755bdc0e59eefa381e78/packages/esm-active-visits-app/src/config-schema.ts#L55)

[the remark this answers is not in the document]
brandones: Oh I see, I understand your confusion! Looks like Dennis identified a config usage that wasn't actually in the schema [here in active-visits.resource.ts](https://github.com/harshthakkr/openmrs-esm-patient-management/blob/3242ab8140f39411c8e4755bdc0e59eefa381e78/packages/esm-active-visits-app/src/active-visits-widget/active-visits.resource.ts#L106), which is crazy and bad. So that is why we are adding this to the schema.

In the future, you can investigate this kind of thing yourself by using the search function in VSCode or GitHub. Here, all I did was find all files using `useConfig` in `esm-active-visits-app` (there were only two), and then search in those files for the text `attributes`.

brandones (approved):
Looks good! And thank you for following my guidelines about how to split your commits; it really saved a ton of work on my end.

packages/esm-patient-registration-app/src/config-schema.ts
```diff
@@ -300,54 +297,65 @@ export const esmPatientRegistrationSchema = {
...
-          _description: 'Whether to use a fixed day and month for estimated date of birth',
           _default: false,
+          _description: 'Whether to use a fixed day and month for estimated date of birth',
         },
         dayOfMonth: {
           _type: Type.Number,
-          _description: 'The custom day of the month use on the estimated date of birth',
           _default: 0,
+          _description: 'The custom day of the month use on the estimated date of birth',
+          _validators: [validators.inRange(0, 30)],
```
denniskigen, line 352:
```suggestion
          _validators: [validators.inRange(0, 31)],
```

The range (0, 31) matches JavaScript's Date constructor behavior used in `calcBirthdate()`:

  - 0 is valid: new Date(year, month, 0) means "last day of previous month" - allows setting estimated births to end of previous month
  - 1-31 covers all valid days: JavaScript handles month overflow gracefully (Feb 31st → March 2nd)
  - Original (0, 30) was too restrictive: Would incorrectly reject valid day 31

This ensures the config accepts all meaningful date values while preventing invalid inputs that could break date calculations.

denniskigen, line 352:
```suggestion
          _description: 'The custom day of the month use on the estimated date of birth  (0 = last day of previous month, 1-31 = specific day)',
          _validators: [validators.inRange(0, 31)],
```

The range (0, 31) matches JavaScript's Date constructor behavior used in `calcBirthdate()`:

  - 0 is valid: new Date(year, month, 0) means "last day of previous month" - allows setting estimated births to end of previous month
  - 1-31 covers all valid days: JavaScript handles month overflow gracefully (Feb 31st → March 2nd)
  - Original (0, 30) was too restrictive: Would incorrectly reject valid day 31

This ensures the config accepts all meaningful date values while preventing invalid inputs that could break date calculations.

denniskigen: Additionally, the test in `patient-registration.test.tsx` is using the wrong `Date` method to compute [dayOfMonth](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-registration-app/src/patient-registration/patient-registration.test.tsx#L136). Since dayOfMonth represents "day of the month" in the birth date calculation, the test should use `getDate()` to get a realistic day-of-month value (1-31) rather than a day-of-week value (0-6).

Without this fix, the test would set dayOfMonth to 0-6 instead of 1-31, which works but doesn't reflect real-world usage where users would configure actual calendar days.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -154,20 +160,24 @@ export const configSchema = {
   },
   contactAttributeType: {
```
denniskigen, line 161: Tangential to your changes, but it looks like the `contactAttributeType` property in the `ConfigObject` interface we export is wrongly typed as `Array<string>` instead of `string`.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -177,23 +187,14 @@ export const configSchema = {
...
-  defaultFacilityUrl: {
-    _type: Type.String,
-    _default: '',
-    _description: 'Custom URL to load default facility if it is not in the session',
   },
   defaultIdentifierTypes: {
     _type: Type.Array,
-    _element: {
+    _default: ['05ee9cf4-7242-4a17-b4d4-00f707265c8a', 'f85081e2-b4be-4e48-b3a4-7994b69bb101'],
+    _description: 'The identifier types to be display on all patient search result page',
```
denniskigen, line 194:
```suggestion
    _description: 'The identifier types to be displayed on all patient search result page',
```

packages/esm-active-visits-app/src/active-visits-widget/active-visits.test.tsx
```diff
@@ -25,8 +25,9 @@ const mockObsConcepts: Array<OpenmrsResource> = [
 
 const mockConfig: ActiveVisitsConfigSchema = {
   activeVisits: {
-    ...getDefaultsFromConfigSchema(configSchema).activeVisits,
+    ...getDefaultsFromConfigSchema<ActiveVisitsConfigSchema>(configSchema).activeVisits,
     obs: ['160225AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA', '5484AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA'],
+    identifiers: undefined,
```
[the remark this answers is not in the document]
denniskigen: We should probably remove `/ID Number/` from `standardColumnHeaders`.

denniskigen:
Mind fixing up the conflicts from #1546?

denniskigen (approved):
Thanks, @harshthakkr!

harshthakkr:
Thanks @denniskigen!

packages/esm-ward-app/src/config-schema.ts
```diff
@@ -27,14 +27,14 @@ export const addressFields = [
 
 type AddressField = keyof typeof addressFields;
 
-export const configSchema: ConfigSchema = {
+export const configSchema = {
```
chibongho, line 30: Why is the `ConfigSchema` removed? Doesn't it help us type the object better?

chibongho (commented):
I have one small comment. Other than that, I didn't a quick glance and it looks fine.

packages/esm-ward-app/src/config-schema.ts
denniskigen: Can you effect that suggestion, @harshthakkr?

chibongho: Yeah, if nothing else, at least putting the type there is self-documenting and makes it easier to people to know how the object should look.

That said, I think we have a way around not having negated types, by doing something like this [here](https://github.com/openmrs/openmrs-esm-core/blob/b98269009f46d7c0f60eaa0118be9c5b3a6cd4b0/packages/framework/esm-config/src/types.ts#L17-L22):

```
export interface ConfigSchema = BaseConfigSchema | RecursiveConfigSchema;

interface BaseConfigSchema {
  _type: Type;
  _validators?: Array<Validator>;
  _elements?: ConfigSchema;
}

interface RecursiveConfigSchema {
  [key: string]: ConfigSchema | ConfigValue;
  _type: never;
  // [key: `_${string}`]: never; // more aggressively prevent keys with '_' prefix
}
```

harshthakkr: It throws an error for config tree elements, should I write `_type: undefined as never` instead of nothing?

chibongho: Maybe for now, let's just add the `ConfigSchema` typing wherever you can and add a TODO comment where you can't. We can make the typings better later.

harshthakkr:
Can we handle type related things in a separate PR? I think it will need a lot of work. @denniskigen @chibongho

PR #1670, opened 2025-06-17
(feat) O3-4821: Should be able to configure which metrics cards appear on which pages

brandones:
## Requirements

## Summary

The metrics cards strike me as an archetypical example of a good use of the extension system. The new [designs](https://docs.google.com/presentation/d/10RY_Nw2dmiyEdJUkmNw3FuX45VvjcRehBb1VyIA3LwA/edit?slide=id.g33784ab62f4_1_440#slide=id.g33784ab62f4_1_440) call for information about appointments to be shown on the service-queues page. This should be accomplished by making an appointments card from the appointments ESM attach to the service queues metrics cards slot.

Reflecting on the designs more I realized that they are not going to be functional for more users. The designs call for five cards, tightly compressed together in a row, on a very wide screen. So the designs need to be reworked, at which point we can fine-tune what happens when there are more than three cards in the slot.

## Screenshots

https://github.com/user-attachments/assets/e2758ed4-afbb-4be3-a140-632b8814987c

## Related Issue
https://openmrs.atlassian.net/browse/O3-4821

## Other

The tests did not seem to add value, and certainly not more value than their maintenance burden, so I deleted them.


brandones:
This is blocking quite a bit of additional work. Would love a review sometime soon @ibacher @denniskigen @chibongho

ibacher:
Apologies for the lateness to reviewing this. Yeah, metric cards are a perfect use-case for the extension system. @brandones I think some of the other stuff we've merged in around this introduced some conflicts. Could you resolve those?

brandones:
Got it. Thanks @ibacher !

PR #1716, opened 2025-06-28
(fix) Make the visitQueueNumberAttributeUuid error crash the app

brandones:
## Requirements

## Summary
Currently, if the queue is configured to show queue number but the attribute UUID isn't set, it shows an unhelpful error message, then a wrong error message, then crashes with another unuseful error. Instead, it should just crash immediately with a helpful error message. Since this will only come up _during the process of configuration_, it should never appear in a production setting.

## Related Issue
Undoing https://openmrs.atlassian.net/browse/O3-4183

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2945,9 +2945,9 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@openmrs/esm-api@npm:6.3.1-pre.3140":
-  version: 6.3.1-pre.3140
-  resolution: "@openmrs/esm-api@npm:6.3.1-pre.3140"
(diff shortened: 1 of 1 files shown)


brandones:
Wait, it's even worse than I thought. The error is being displayed even when queue-number _isn't_ configured as part of the table. So I need to fix this because right now it just crashes the app all the time.

brandones:
This is the E2E failure:

image

Looks unrelated to this PR.

brandones:
E2E problem resolved itself!

denniskigen (approved):

denniskigen:
Thanks, @brandones!

PR #1780, opened 2025-07-17
(fix) O3-4891: limit appointment note to 255 characters

mseaton:
## Related Issue
https://issues.openmrs.org/browse/O3-4891


chibongho:
Should we try to remove the 255 char limit on the backend instead?

denniskigen (approved):
LGTM. Thanks, @mseaton!

PR #1810, opened 2025-07-25
(fix): Fetch all queue locations using fhirFetchAll

NethmiRodrigo:
## Requirements

## Summary

This PR updates the `useQueueLocations` hook to use `useFhirFetchAll` to fetch all queue locations. Previously this only returned 10. This also adds a loading indicator to render until the queue locations are loaded.

## Screenshots

Loading indicator:
Screenshot 2025-07-25 at 3 44 16 PM

Queue locations is now no longer limited to only 10 locations:
https://github.com/user-attachments/assets/1d656685-7802-4854-bd3b-8900c9eaafff

## Other

Related slack thread - https://openmrs.slack.com/archives/C039BLMGMMX/p1752498396351849

### DIFF
--- a/packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts
+++ b/packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import { fhirBaseUrl, getLocale, openmrsFetch } from '@openmrs/esm-framework';
+import { fhirBaseUrl, getLocale, openmrsFetch, useFhirFetchAll } from '@openmrs/esm-framework';
 import useSWRImmutable from 'swr/immutable';
 
 interface FHIRResponse {
@@ -11,14 +11,12 @@ interface FHIRResponse {
 
 export function useQueueLocations() {
   const apiUrl = `${fhirBaseUrl}/Location?_summary=data&_tag=queue location`;
-  const { data, error, isLoading } = useSWRImmutable<{ data: FHIRResponse }>(apiUrl, openmrsFetch);
+  const { data, error, isLoading } = useFhirFetchAll<fhir.Location>(apiUrl);
 
   const queueLocations = useMemo(
-    () =>
-      data?.data?.entry
-        ?.map((response) => response.resource)
-        .sort((a, b) => a.name.localeCompare(b.name, getLocale())) ?? [],
-    [data?.data?.entry],
+    () => data?.map((response) => response).sort((a, b) => a.name.localeCompare(b.name, getLocale())) ?? [],
+    [data],
   );
+
   return { queueLocations, isLoading, error };
 }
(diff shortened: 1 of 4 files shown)


packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts
```diff
@@ -11,14 +11,12 @@ interface FHIRResponse {
 
 export function useQueueLocations() {
   const apiUrl = `${fhirBaseUrl}/Location?_summary=data&_tag=queue location`;
-  const { data, error, isLoading } = useSWRImmutable<{ data: FHIRResponse }>(apiUrl, openmrsFetch);
+  const { data, error, isLoading } = useFhirFetchAll<fhir.Location>(apiUrl);
```
chibongho, line 14: You should be able to pass in an additional param `{immutable: true}` to make it immutable.

chibongho: `useSWR()` optimistically returns any cache data while doing a fetch in the background. It also does that every so often to revalidate the data to make sure it is not stale. How often it is can be configurable. If we don't anticipate the data to change over the course of the web page, we can use `useSWRImmutable()` instead to disable the revalidation. 

See: https://swr.vercel.app/docs/revalidation

packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts
```diff
@@ -1,5 +1,5 @@
 import { useMemo } from 'react';
-import { fhirBaseUrl, getLocale, openmrsFetch } from '@openmrs/esm-framework';
+import { fhirBaseUrl, getLocale, openmrsFetch, useFhirFetchAll } from '@openmrs/esm-framework';
 import useSWRImmutable from 'swr/immutable';
 
 interface FHIRResponse {
```
ibacher, line 5: Do we need this type any more?

packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts
ibacher: I worry that using immutable is not long-term sustainable across application. Unless there's truly a huge amount of overhead to a request, it doesn't make a huge difference performance-wise and, in this case, it's pretty easy to imagine a case where marking this as immutable will do the wrong thing (since it's designed to be a long-running SPA and some implementations will filter which locations are visible based on the currently logged-in user, meaning if we mark this as immutable, some people will end up seeing the wrong list of locations).

packages/esm-service-queues-app/src/patient-queue-header/patient-queue-header.scss
```diff
@@ -11,6 +11,10 @@
   padding-right: layout.$spacing-03;
 }
 
+.dropdownSkeletonContainer {
+  width: 300px;
```
denniskigen, line 15:
```suggestion
  width: 18.75rem;
```

nit

denniskigen (approved):
LGTM. Thanks, @NethmiRodrigo. Could we figure out where else we could extend this change?

PR #1983, opened 2025-09-18
(feat) O3-4974: Show filter appointments by provider dropdown if provider column is enabled

Debbielou:
## Requirements

## Summary

Currently, the appointments view lacks a  filter for providers. While users can filter by service type, they should also have an option to filter by provider if desired.

Added a new feature for provider filter dropdown in the appointments view. This filter should be dynamically displayed or hidden based on whether the provider column is enabled. Users will only see the option to filter by provider when the provider column is active.

[Screencast from 2026-02-11 08-35-49.webm](https://github.com/user-attachments/assets/7b759721-6d2f-41e2-8601-a59531d13c34)
## Related Issue
https://openmrs.atlassian.net/browse/O3-4974


brandones (changes requested):
See comment

brandones (changes requested):
It looks like you are showing the full set of providers in the provider drop down. Don't do that; get the list of providers for the filter from the actual data. This way we do not include providers for whom there are no rows.

brandones (changes requested):
The list of providers in the filter dropdown should not change when a provider is selected

PR #2003, opened 2025-10-02
(chore) Fix type errors and refactor dashboard links

denniskigen:
## Requirements

## Summary

This commit updates the OpenMRS dependencies and refactors the dashboard links to use the `DashboardExtension` component from the styleguide. This resolves an issue where the auto-update workflow [was failing](https://github.com/openmrs/openmrs-esm-patient-management/actions/workflows/update-openmrs-deps.yml) due to type errors related to upstream changes in how the framework handles icon properties. Changes made here include:

- Update OpenMRS dependencies to the latest `next` tagged versions
- Fix TypeScript error from framework upgrade (8.0.1-pre.3318 → 3349) where the `icon` property became more strictly typed
- Refactor appointments, service queues, ward, and patient list apps to use the `DashboardExtension` from `@openmrs/esm-styleguide` instead of local implementations
- Rename `createDashboardLink.component.tsx` to `createDashboardLink.tsx` (proper naming for factory functions)
- Remove duplicate local interfaces in favor of `Omit` utility with framework types
- Update dashboard metadata to include `basePath` and use framework types

## Screenshots

CleanShot 2025-10-02 at 12 00 32@2x

## Related Issue
https://github.com/openmrs/openmrs-esm-core/pull/1454
https://github.com/openmrs/openmrs-esm-patient-chart/pull/2756


packages/esm-appointments-app/src/dashboard.meta.ts
```diff
@@ -1,22 +1,21 @@
-import { Calendar } from '@carbon/react/icons';
-
 export const dashboardMeta = {
-  name: 'appointments',
+  path: 'appointments',
   slot: 'clinical-appointments-dashboard-slot',
   title: 'Appointments',
-};
+  basePath: `${window.spaBase}/home`,
+} as const;
```
denniskigen, line 6: Using `as const` here instead of explicit typing because no single framework type
has both `basePath` (required by `DashboardExtension`) and `slot` (needed for
extension registration). The const assertion also preserves literal types which
is critical for the strict icon type checking introduced in framework v8.0.1-pre.3349.

packages/esm-appointments-app/src/createDashboardLink.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { BrowserRouter } from 'react-router-dom';
+import { DashboardExtension, type DashboardExtensionProps, type IconId } from '@openmrs/esm-framework';
+
+export const createDashboardLink = (config: Omit<DashboardExtensionProps, 'icon'> & { icon?: IconId }) => () => (
```
denniskigen, line 5: We're keeping this local implementation instead of using the framework's `createDashboardLink` from `@openmrs/esm-patient-common-lib` because they have different signatures. The framework version expects `basePath` as a component prop `({ basePath }) => ...`, but our home dashboard extensions don't receive it that way.
Our version takes `basePath` in the config itself, which matches how these extensions are currently registered in `routes.json`.

It'd be great if we could have a single implementation in the framework that handles all the possible scenarios flexibly.

packages/esm-patient-list-management-app/src/createDashboardLink.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+import { BrowserRouter } from 'react-router-dom';
+import { DashboardExtension, type DashboardExtensionProps, type IconId } from '@openmrs/esm-framework';
+
+export const createDashboardLink = (config: Omit<DashboardExtensionProps, 'icon'> & { icon?: IconId }) => () => (
```
denniskigen, line 5: This pattern makes the `icon` property optional because the home dashboard configs don't have icons. The type `Omit<DashboardExtensionProps, 'icon'> & { icon?: IconId }` allows configs without icons while maintaining strict typing for those that include them.

chibongho: We should probably just have `icon` be optional in `DashboardExtensionProps` when we merge these implementations of `createDashboardLink()`

ibacher: Yeah, the icon property should definitely be optional...

PR #2011, opened 2025-10-06
(fix) O3-5074: Prevent home page crashing when default dashboard is not found

donaldkibet:
## Requirements

## Summary
Add null check before accessing `activeDashboard` to prevent application crash when the configured or fallback dashboard doesn't exist in available dashboards.

The issue occurs in several scenarios:
1. When a user's role is mapped to a non-existent dashboard
2. When service-queues package is not installed (happens for kenya-emr)
3. When users have unconfigured roles (falls back to hardcoded 'service-queues')

The code has a hard coded fallback to 'service-queues' which creates a hidden dependency - deployments without service-queues will crash on the home page.

## Related Issue

https://thepalladiumgroup.atlassian.net/browse/KHP3-8575

### DIFF
--- a/packages/esm-home-app/src/default-dashboard-redirect.component.tsx
+++ b/packages/esm-home-app/src/default-dashboard-redirect.component.tsx
@@ -16,5 +16,9 @@ export function DefaultDashboardRedirect() {
   const dashboards = ungroupedDashboards as Array<DashboardConfig>;
   const activeDashboard = dashboards.find((dashboard) => dashboard.name === defaultDashboard);
 
+  if (!activeDashboard && dashboards.length > 0) {
+    return <Navigate to={`/home/${dashboards[0].name}`} />;
+  }
+
   return <Navigate to={`/home/${activeDashboard.name}`} />;
 }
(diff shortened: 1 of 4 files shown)


packages/esm-home-app/src/default-dashboard-redirect.component.tsx
```diff
@@ -16,5 +16,9 @@ export function DefaultDashboardRedirect() {
   const dashboards = ungroupedDashboards as Array<DashboardConfig>;
   const activeDashboard = dashboards.find((dashboard) => dashboard.name === defaultDashboard);
 
+  if (!activeDashboard && dashboards.length > 0) {
+    return <Navigate to={`/home/${dashboards[0].name}`} />;
+  }
+
   return <Navigate to={`/home/${activeDashboard.name}`} />;
```
ibacher, line 23: So what happens if there's no active dashboard and no dashboards available? Won't that still leave us in the same place?

ibacher (commented):
Thanks @donaldkibet! Just one edge case to cover and I think this is probably fine.

PR #2285, opened 2026-02-27
(fix) Prevent errors when queue entries have no priority configured

donaldkibet:
## Requirements

## Summary
This PR prevents undefined error when the queue priority hasn't been set. This happens in our distro(KeHMIS) since priority is evaluated using TEWS calculation from triage.

### DIFF
--- a/packages/esm-service-queues-app/src/modals/move-queue-entry.modal.tsx
+++ b/packages/esm-service-queues-app/src/modals/move-queue-entry.modal.tsx
@@ -120,7 +120,7 @@ const MoveQueueEntryModal: React.FC<MoveQueueEntryModalProps> = ({ queueEntry, c
         disableSubmit: (queueEntry, formState) =>
           formState.selectedQueue == queueEntry.queue.uuid &&
           formState.selectedStatus == queueEntry.status.uuid &&
-          formState.selectedPriority == queueEntry.priority.uuid,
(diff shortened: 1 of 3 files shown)


packages/esm-service-queues-app/src/modals/move-queue-entry.modal.tsx
```diff
@@ -120,7 +120,7 @@ const MoveQueueEntryModal: React.FC<MoveQueueEntryModalProps> = ({ queueEntry, c
         disableSubmit: (queueEntry, formState) =>
           formState.selectedQueue == queueEntry.queue.uuid &&
           formState.selectedStatus == queueEntry.status.uuid &&
-          formState.selectedPriority == queueEntry.priority.uuid,
+          formState.selectedPriority == queueEntry?.priority?.uuid,
```
ibacher: I don't love this is a general implementation. Yes, this fixes the issue for KeHMIS, but not for anyone else. I can't merge this in if users can just create entries without setting priorities, which is what this blocks.

ibacher (changes requested):
@donaldkibet Thanks, but I don't think this is mergeable in this state. We still need to try to prevent users from creating queue entries without priorities in the RefApp and not just KeHMIS. If the allowed laxness can be hidden behind configuration, I suppose this is acceptable, but as-is, this will break any other implementation.

PR #2298, opened 2026-03-05
(fix) Make My Lists tab show lists created by the logged-in user

icrc-jofrancisco:
## Requirements

## Summary
The Patient Lists page currently contains four tabs:

- Starred Lists – lists starred by the current user
- System Lists – lists with cohort type "System List"
- My Lists – lists with cohort type "My List"
- All Lists – all available lists

This pull request changes the behavior of the "My Lists" tab.

Instead of displaying patient lists with the cohort type "My List", the tab will now display lists created by the currently logged-in user, regardless of the cohort type.

## Other
The current behavior is confusing for many users because the "My Lists" tab does not actually represent lists created by them. Instead, it filters by a cohort type that happens to be named "My List", which can easily lead to misunderstandings.

This mismatch between the tab name and the underlying logic frequently causes users to:

- Assume they are seeing lists they personally created
- Make mistakes when managing or locating their own lists

By changing the tab to show lists created by the logged-in user, the behavior becomes more intuitive and aligns with the expectation implied by the tab name.

Resulting Behavior

- Starred Lists: unchanged
- System Lists: unchanged
- My Lists: now shows lists created by the logged-in user
- All Lists: unchanged

This change improves usability and reduces user confusion when navigating and managing patient lists.

Thanks,


packages/esm-patient-list-management-app/src/api/hooks.ts
```diff
@@ -77,13 +76,19 @@ export function useAllPatientLists({ isStarred, type }: PatientListFilter) {
     type: cohort.cohortType?.display,
     size: cohort.size,
     location: cohort.location,
+    creatorUuid: cohort?.creator?.uuid,
```
jayasanka-sack:
```suggestion
    creatorUuid: cohort.creator?.uuid,
```

packages/esm-patient-list-management-app/src/api/hooks.ts
```diff
@@ -25,16 +25,15 @@ interface PatientListResponse {
...
+  const custom =
+    'custom:(uuid,name,description,display,size,attributes,cohortType,location:(uuid,display),creator:(uuid,display))';
   const query: Array<[string, string]> = [
     ['v', custom],
     ['totalCount', 'true'],
   ];
   const config = useConfig<PatientListManagementConfig>();
 
-  if (type === PatientListType.USER) {
-    query.push(['cohortType', config.myListCohortTypeUUID]);
```
jayasanka-sack: It would be better to remove this config as well, since it could otherwise introduce ambiguity for anyone going through the configuration.

packages/esm-patient-list-management-app/src/api/types.ts
```diff
@@ -69,6 +69,10 @@ export interface OpenmrsCohort {
   type?: string;
   size: number;
   cohortType?: CohortType;
+  creator?: {
+    uuid: string;
+    display: string;
+  };
```
jayasanka-sack: if we request for a creator in custom representation, the creator will be `{
    uuid: string;
    display: string;
  }` or `null`. Therefore, we need to add null type as well.

jayasanka-sack (commented):
Thanks for the PR @icrc-jofrancisco! I added some comments. :)

packages/esm-patient-list-management-app/src/api/hooks.ts
icrc-jofrancisco: Done. Thanks!

packages/esm-patient-list-management-app/src/api/hooks.ts
icrc-jofrancisco: I agree. I kept it for compatibility reasons, but I do agree that it doesn't make sense to have a tab called “My lists” that is just a filter for a cohort type. Removed from the settings to avoid confusion. Thanks

packages/esm-patient-list-management-app/src/api/types.ts
icrc-jofrancisco: Done. Thanks!

PR #2315, opened 2026-03-12
(feat) Enhance patient list location handling and validation

icrc-jofrancisco:
## Requirements

## Summary
Improves the usability and data integrity of the Location property in Cohorts / Patient Lists. The update surfaces the list location more clearly, allows setting/editing it directly in the UI, and introduces an optional configuration to enforce location consistency when adding patients.

### _Display Cohort Location in Header_

If a patient list has a location defined, it is now displayed in the cohort header for better visibility and context.
image

### Set / Edit Location on Patient Lists

Users can now define or modify the Location when creating or editing a patient list.

Key details:
- A searchable dropdown lists all system locations.
- While locations are loading, an inline field spinner is shown. 
- The field is pre-populated with the current user's location.
- The location remains optional, maintaining previous behavior.

image

Once selected, the location updates immediately and the header reflects the change in real time.

### `enforcePatientListLocationMatch` Configuration

A new configuration flag has been introduced:

`enforcePatientListLocationMatch = false`
By default, enforcement is **disabled**.

In environments where users operate across multiple locations, patients are sometimes added to lists belonging to the wrong location. Previous approaches (other modules) used warning pop-ups that still allowed the action, which led to inconsistent data.

New Behavior (when enabled)

When `enforcePatientListLocationMatch = true`:

Patients can only be added to lists whose location matches at least one of the patient's identifier locations.
If no matching location is found, the patient cannot be added and an error message is shown.
Patient location does not

### Patient File Integration

The same validation rules apply when a patient is added to a list directly from the patient file.

Add patient to list

When adding a patient to multiple lists simultaneously, results are handled per list.
If adding to one list fails but succeeds for another, both messages will be displayed:
- One error message for the failed list
- One success message for the successful list

Patient location does not

### Impact

- Improves visibility of list location
- Reduces data inconsistencies across locations
- Adds optional strict validation for organizations that require location integrity
- Keeps backward compatibility by leaving enforcement disabled by default

## Related Issue
N/A
## Other

Thanks,


jayasanka-sack:
Thanks for the PR @icrc-jofrancisco!

Since we decided to go with the data-filteration module to handle cross-location issues, I’m not entirely sure this is still needed. What do you think?

Also, would be great to hear how you’re planning to proceed with it on your end.


@ibacher, would like to hear your take on this.

icrc-jofrancisco:
> Thanks for the PR @icrc-jofrancisco!
> 
> Since we decided to go with the data-filteration module to handle cross-location issues, I’m not entirely sure this is still needed. What do you think?
> 
> Also, would be great to hear how you’re planning to proceed with it on your end.
> 
> @ibacher, would like to hear your take on this.

Thanks for your feedback @jayasanka-sack !

That’s a fair point. Even with the data-filtration module handling the cross-location logic, we believe this addition is still valuable for improving the user experience. By handling this on the client side, we can avoid unnecessary backend calls and provide more immediate, context-specific error messages for our scenario.

To keep it non-intrusive, we’ve implemented it behind a configuration, similar to our approach in other modules. It will be disabled by default, so it won't affect users who don't explicitly need it.

In addition, we have multiple environments with different configurations. This will make it much easier for us to set things up and will perfectly meet our needs.

Thanks,

PR #2320, opened 2026-03-13
fix: average wait time card never fetches data

shubhangiisinghh:
PR Description:

The average wait time card was always showing '--' and never loading actual data.

Turns out useAverageWaitTime was being called with an empty string as the statusUuid. The hook skips the API call if either serviceUuid or statusUuid is falsy — so the request was never made.

The fix is just passing defaultStatusConceptUuid from config, which is what waiting-patients.extension.tsx already does right next door.


jayasanka-sack (approved):
Thanks for fixing this, @shubhangiisinghh!

shubhangiisinghh:
Much grateful to have contributed @jayasanka-sack

PR #2322, opened 2026-03-14
(fix) Remove unnecessary type assertion in mapVisitQueueEntryProperties

shubhangiisinghh:
queueEntry.queue is already typed as Queue, which has a location: Location field defined in the type. The as any cast to access location.uuid was unnecessary and bypassed TypeScript's type checking.

Removed the cast so the type system can do its job properly.


jayasanka-sack (approved):
Thanks @shubhangiisinghh! 🙌

For future PRs, please follow the PR title guidelines here: [https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#pull-requests](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#pull-requests).

PR #2323, opened 2026-03-14
(fix) O3-5488: Unnecessary event listener re-registration in combo-input component

Pratikdeb07:
## Requirements

---

## Summary

The `useEffect` hook in `combo-input.component.tsx` did not include a dependency array.  
Because of this, the effect executed on every render and repeatedly added and removed a `click` event listener on `window`.

Although the cleanup function removed the listener each time, the listener was still unnecessarily re-registered on every render cycle.

This PR fixes the issue by adding an empty dependency array (`[]`) so that the event listener is registered only once when the component mounts and removed when the component unmounts.

This change aligns the component with React best practices and avoids unnecessary operations during re-renders.

---

## Screenshots

No visual UI changes. This update only improves the lifecycle behavior of the event listener.

---

## Related Issue

https://openmrs.atlassian.net/browse/O3-5488

---

## Files Changed

packages/esm-patient-registration-app/src/patient-registration/input/combo-input/combo-input.component.tsx

---

## Other

This change improves performance and follows React best practices by preventing unnecessary effect executions on every render.


divv919 (approved):
Looks good, i think this can be merged? @jayasanka-sack

jayasanka-sack (approved):
Good catch, @Pratikdeb07!

jayasanka-sack:
By the way, your PR description says _'validated by existing tests'_ and there are no tests for this component. Don't tick that box if it's not true.

PR #2326, opened 2026-03-15
(fix): O3-5490 Prevent AbortController recreation on every render in appointment hooks

Pratikdeb07:
## Requirements

## Summary

AbortController was being instantiated inside several React hook bodies. Since React hooks execute on every render, this resulted in a new AbortController being created each time the component rendered.

As a result, requests initiated with previous controllers could not be properly aborted when the component unmounted.

This PR fixes the issue by storing the AbortController instance in a `useRef` so it persists across renders, and aborting the request during component cleanup using `useEffect`.

## Changes Made

- Store `AbortController` in `useRef` instead of creating it on every render.
- Abort the request in `useEffect` cleanup.
- Updated the following files:

- `packages/esm-appointments-app/src/hooks/useAppointmentList.ts`
- `packages/esm-appointments-app/src/hooks/usePatientAppointmentHistory.ts`
- `packages/esm-service-queues-app/src/queue-patient-linelists/queue-linelist.resource.ts`

## Screenshots

Not applicable. This change only affects request lifecycle management and does not introduce UI changes.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5490

## Other

Existing tests pass successfully after the change.

### DIFF
--- a/packages/esm-appointments-app/src/hooks/useAppointmentList.ts
+++ b/packages/esm-appointments-app/src/hooks/useAppointmentList.ts
@@ -3,18 +3,24 @@ import useSWR from 'swr';
 import { openmrsFetch, restBaseUrl } from '@openmrs/esm-framework';
 import { type AppointmentsFetchResponse } from '../types';
 import { useAppointmentsStore } from '../store';
+import { useRef, useEffect } from 'react';
 
 export const useAppointmentList = (appointmentStatus: string, date?: string) => {
   const { selectedDate } = useAppointmentsStore();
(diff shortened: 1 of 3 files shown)


sohamdhande (changes requested):
The AbortController useRef approach is on the right track, but there are a few issues to address before this is ready to merge:

The useEffect cleanup has an empty deps array in all three files — this doesn't match the actual hook dependencies
startDate parameter in usePatientAppointments is untyped
No tests covering the new abort/cleanup behaviour
React imports should be ordered first across all changed files

PR #2327, opened 2026-03-15
fix: Update type casting for patient name in appointment display (O3-5491)

divv919:
## Requirements

## Summary
This PR removes unsafe any casts from the translated appointment display in [batch-change-appointment-statuses.modal.tsx:147]  The issue was limited to the text rendered through the appointmentDisplay translation, not to the bulk status update logic itself.

What was fixed

The modal was passing patientName, serviceName, and currentStatus into the Trans component using any casts to satisfy TypeScript. That bypassed type checking for values that are already well-defined on the Appointment model. This change removes those broad casts, keeps serviceName and currentStatus as properly typed interpolated values, and uses a narrower React.ReactNode cast only for patientName where it is rendered inside the strong element.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5491


jayasanka-sack:
Thanks for the PR @divv919! the correct fix would be to use values prop and this has been already addressed in https://github.com/openmrs/openmrs-esm-patient-management/pull/2518/changes
So, closing the PR! Thanks again! 💛

PR #2331, opened 2026-03-16
fix (O3-5493) : fix: remove invalid CSS backgroundColor value in queue linelist toolbar

Pratikdeb07:
## Requirements

## Summary
This PR fixes an invalid CSS value used in the queue linelist toolbar.

The `TableToolbar` component was using `backgroundColor: 'color'`, which is not a valid CSS value. Browsers ignore this value, causing the toolbar background styling to be inconsistent.

The invalid value has been removed so the component can inherit the correct background styling from the Carbon theme used in the OpenMRS O3 frontend.

## Screenshots
Before:
Toolbar background style was ignored because `'color'` is not a valid CSS value.

After:
Toolbar now correctly inherits the background styling.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5493

## Other
This is a small UI fix that improves consistency and prevents invalid CSS from being applied.


jayasanka-sack:
This component got removed recently. Therefore, closing the PR. Thanks @Pratikdeb07!

PR #2332, opened 2026-03-16
(fix) O3-5495: Improve gender handling and translation in unscheduled appointments table

Pratikdeb07:
## Requirements

## Summary
This PR improves the handling of patient gender in the Unscheduled Appointments table.

Previously, gender was derived using a hard-coded ternary comparison:

`visit.gender === 'F' ? 'Female' : 'Male'`

This implementation had two issues:
- It used hard-coded English strings, which bypasses the translation system.
- It assumed only two gender values (`F` and `M`), causing any other values to incorrectly display as `Male`.

This PR replaces the fragile comparison with a safer mapping-based approach and uses translation keys to support proper internationalization. Missing translation keys (`female`, `male`, `other`) were also added to the module's `en.json` file.

## Screenshots
UI remains visually the same for `F` and `M` values (Female/Male).  
For any other gender values, the table will now correctly display **Other** instead of incorrectly showing **Male**.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5495

## Other
This change improves internationalization support and ensures gender values are handled safely without assuming only two possible values.

### DIFF
--- a/packages/esm-appointments-app/src/appointments/unscheduled/unscheduled-appointments.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/unscheduled/unscheduled-appointments.component.tsx
@@ -20,6 +20,7 @@ import { Download } from '@carbon/react/icons';
 import { ConfigurableLink, EmptyCard, useConfig, usePagination } from '@openmrs/esm-framework';
 import { useUnscheduledAppointments } from '../../hooks/useUnscheduledAppointments';
 import { exportUnscheduledAppointmentsToSpreadsheet } from '../../helpers/excel';
+import { getGender } from '../../helpers';
 import { getPageSizes, useSearchResults } from '../utils';
 import { type ConfigObject } from '../../config-schema';
 
@@ -62,7 +63,7 @@ const UnscheduledAppointments: React.FC = () => {
         {visit.name}
       </ConfigurableLink>
     ),
-    gender: visit.gender === 'F' ? 'Female' : 'Male',
+    gender: getGender(visit.gender),
     phoneNumber: visit.phoneNumber === '' ? '--' : visit.phoneNumber,
     identifier: visit?.identifier,
   }));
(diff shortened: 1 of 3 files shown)


Pratikdeb07:
@ibacher  @denniskigen please review this pr.

packages/esm-appointments-app/src/appointments/unscheduled/unscheduled-appointments.component.tsx
```diff
@@ -62,7 +62,8 @@ const UnscheduledAppointments: React.FC = () => {
         {visit.name}
       </ConfigurableLink>
     ),
-    gender: visit.gender === 'F' ? 'Female' : 'Male',
+    gender:
+      visit.gender === 'F' ? t('female', 'Female') : visit.gender === 'M' ? t('male', 'Male') : t('other', 'Other'),
```
jayasanka-sack: Nice catch! 

By the way, a getGender helper already exists here: https://github.com/openmrs/openmrs-esm-patient-management/blob/51ff93900c1a31baf35a18976ebc98dcaa185665/packages/esm-appointments-app/src/helpers/functions.ts#L44

Let's use that.

jayasanka-sack (changes requested):

packages/esm-appointments-app/src/appointments/unscheduled/unscheduled-appointments.component.tsx
jayasanka-sack: Although we need to add the missing translation keys.

jayasanka-sack:
@Pratikdeb07 haven’t heard from you in a while on this PR. Could you take a look at the review comments and push an update when you get a chance?

Pratikdeb07:
Done! Thanks for your valueable review @jayasanka-sack .

jayasanka-sack:
Excellent! The function uses the translation key 'unknown', but that one seems missing in the translations file for some reason. Could you please fix it?

Pratikdeb07:
Thanks for pointing out that @jayasanka-sack ! It is fixed now.

packages/esm-appointments-app/translations/en.json
```diff
@@ -87,6 +87,7 @@
   "failedToEndActiveVisit": "Failed to end active visit for {{patient}}",
   "failedToLoadMetrics": "Failed to load",
   "failedToUpdateVisit": "Failed to update visit",
+  "female": "Female",
```
ibacher: The translations for these should be fetched from the core translations. We want to move away from needing to provide these in all applications.

Pratikdeb07:
Thanks for your review @ibacher ! I have used chore translations now.

PR #2333, opened 2026-03-16
feat (O3-5496): avoid mounting hidden extension panels in scheduled appointments

Pratikdeb07:
## Requirements

## Summary
Currently, extension panels in the Scheduled Appointments view are always mounted in the DOM and hidden using `display: none`.

This means that all extensions mount and run their hooks even when they are not visible. As a result, hidden panels may still trigger unnecessary network requests and consume additional memory.

This PR replaces the `display: none` approach with conditional rendering so that only the active tab panel is mounted. Hidden panels are no longer rendered, preventing unnecessary hook execution and improving performance.

## Screenshots
No visual UI changes are expected.  
The active tab continues to render normally, while inactive panels are not mounted instead of being hidden.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5496

## Other
This change follows recommended React practices by avoiding mounting components that are not visible, improving efficiency and reducing unnecessary work during rendering.

### DIFF
--- a/packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
+++ b/packages/esm-appointments-app/src/appointments/scheduled/scheduled-appointments.component.tsx
@@ -175,11 +175,11 @@ function ExtensionWrapper({
     return null;
   }
 
+  if (currentTab !== extension.name) {
+    return null;
+  }
   return (
-    <div
-      key={extension.name}
-      className={styles.container}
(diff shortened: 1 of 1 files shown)


Josephkagimu1 (commented):
1. No tests — The PR claims "validated by existing tests" but there are no tests verifying tab-switching behavior ,

2. Title label is not a feat. No new feature is added; it's an optimization of existing behavior.
3. Mind about conflict resolution 

Thanks for working on this @Pratikdeb07

PR #2336, opened 2026-03-16
fix: type unscheduled appointments in excel export helpers

shubhangiisinghh:
Description:

exportUnscheduledAppointmentsToSpreadsheet in excel.ts was accepting Array<any> even though the shape of each appointment object is well-defined — it comes directly from useUnscheduledAppointments which already maps it to a consistent structure.

Extracted that mapped shape as MappedUnscheduledAppointment and exported it from useUnscheduledAppointments.ts so it can be reused. Also typed the internal createWorksheet helper properly instead of any[].

No behaviour changes.

Related Issue
Will link once Jira access is granted.


Pratikdeb07 (commented):
MappedUnscheduledAppointment looks good overall, but consider using Date | string or ensuring proper conversion.
Also, in createWorksheet as string cast seems like a workaround — if data is correctly typed upstream, this cast shouldn’t be necessary.

dilankavishka:
@jayasanka-sack We can close this PR because this issue was fixed in here : https://github.com/openmrs/openmrs-esm-patient-management/pull/2518

jayasanka-sack:
Thanks for the PR @shubhangiisinghh! Closing as this was already addressed in #2517!

shubhangiisinghh:
Hey @jayasanka-sack , thanks for the context.                                                                                                                         
Just wanted to flag this for future reference since I had opened #2336 about a month earlier covering the same typing issue in the excel helpers. I understand #2518 had a broader scope and proper Jira linkage which makes sense as a better fit. Would be helpful to know the best way to keep a PR visible while waiting for review so this kind of overlap can be avoided. Happy to contribute more, just want to make sure efforts don't get duplicated. Thanks again.

jayasanka-sack:
Hey @shubhangiisinghh, yeah fair point. it does suck when work overlaps.

Just sharing some context on why this happened:

- The PR wasn’t linked to the Jira ticket (and ticket wasn’t linked back on the PR description either), so it wasn’t very visible
- The PR template bits (ticket link, title format, etc.) were removed, which made it harder to track
- It didn’t fully cover the scope of the ticket (that one was for the whole appointments app)
- There were also some review comments that didn’t get responses for a while

Also during GSoC we get a pretty high volume of PRs, so if something isn’t linked/active it’s easy to miss.

#2518 was properly linked + tracked through the ticket, so it ended up being picked up there.

For future PRs, if you keep the links + template + replies in place, it’s much easier for us to spot ongoing work and avoid this kind of clash.

Appreciate the effort here though... feel free to pick up another one! ❤️

PR #2344, opened 2026-03-17
(fix) O3-5505: Ensure isSubmitting resets when checkAppointmentConflict throws

Pratikdeb07:
## Requirements

---

## Summary

This PR fixes an issue where the `isSubmitting` state in the appointments form could remain `true` if an error occurs while checking appointment conflicts.

In `handleSaveAppointment`, `setIsSubmitting(true)` is called before the async request `checkAppointmentConflict()`. If this request throws an error (e.g., network failure or API error), execution exits the function before `setIsSubmitting(false)` is called. This causes the submit button to remain disabled and the form becomes stuck in a submitting state.

To resolve this:

- The logic inside `handleSaveAppointment` is wrapped in a `try/catch/finally` block.
- The `.then()` promise chain was converted to `async/await` for cleaner error handling.
- The `finally` block ensures `setIsSubmitting(false)` is always executed, regardless of success, failure, or early return.

This guarantees that the submit button is re-enabled and the form does not remain stuck after an error.

---

## Screenshots

Not applicable.  
This change only affects error handling logic and does not modify the UI.

---

## Related Issue

https://openmrs.atlassian.net/browse/O3-5505

---

## Other

This change improves reliability of the appointments form by ensuring the submitting state is always reset, preventing a poor user experience where the form becomes permanently disabled after an error.

### DIFF
--- a/packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
+++ b/packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
@@ -333,78 +333,81 @@ const AppointmentsForm: React.FC<Workspace2DefinitionProps<AppointmentsFormProps
   // Same for creating and editing
   const handleSaveAppointment = async (data: AppointmentFormData) => {
     setIsSubmitting(true);
-    // Construct appointment payload
-    const appointmentPayload = constructAppointmentPayload(data);
+    try {
+      // Construct appointment payload
+      const appointmentPayload = constructAppointmentPayload(data);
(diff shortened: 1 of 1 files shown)


Ranjesh2002 (changes requested):

jayasanka-sack (changes requested):
Hi! There are some merge conflicts that need to be resolved. Could you please fix them and let me know once it’s ready for review?

PR #2374, opened 2026-03-21
(fix) O3-5527: Avoid duplicate DOM ids in Autosuggest component

Pratikdeb07:
## Requirements

## Summary
This PR fixes an issue in the Autosuggest component where a hardcoded `id="autosuggest"` was used for the `<Search>` element.

When multiple Autosuggest components are rendered on the same page (e.g., in patient registration or relationships section), this results in duplicate DOM ids, which violates HTML standards.

This can lead to:
- Accessibility issues (incorrect label associations and screen reader behavior)
- Incorrect DOM lookups (e.g., `document.getElementById` returns only the first match)
- Potential UI inconsistencies

## Screenshots
N/A (No visible UI changes)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5527

## Other
N/A


Brijesh-0106 (commented):
name is just an alias for searchProps.id, so if searchProps.id isn't provided, the fallback becomes autosuggest-undefined which isn't great. Maybe enforce id as a required prop, or at least add a safer fallback?

PR #2375, opened 2026-03-21
(fix) O3-5528: Use SPA navigation instead of history.back in cancelRegistration

Pratikdeb07:
## Requirements

## Summary
This PR replaces the use of `window.history.back()` in `cancelRegistration` with the OpenMRS SPA navigation utility.
Using `window.history.back()` bypasses the SPA routing system and depends on the browser history stack, which can lead to unpredictable navigation and may take users outside the application.
Replaced `window.history.back()` with `navigate()` from `@openmrs/esm-framework`
Ensures consistent and controlled navigation within the SPA

## Validation
Manually tested navigation from patient registration page
Verified that cancel action routes to the home page consistently
Confirmed no navigation outside the SPA

## Screenshots
N/A (No UI changes)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5528

## Other
This change aligns navigation behavior with OpenMRS frontend best practices and improves reliability of user flow.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
@@ -1,5 +1,5 @@
 import camelCase from 'lodash-es/camelCase';
-import { parseDate } from '@openmrs/esm-framework';
+import { parseDate , navigate } from '@openmrs/esm-framework';
 import {
   type Encounter,
   type FormValues,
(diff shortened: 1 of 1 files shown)


packages/esm-patient-registration-app/src/patient-registration/patient-registration-utils.ts
```diff
@@ -16,7 +16,7 @@ export function scrollIntoView(viewId: string) {
 }
 
 export function cancelRegistration() {
-  window.history.back();
```
sidd4win, line 19: Good fix! Replacing `window.history.back()` with OpenMRS `navigate()` is the right approach — ensures SPA routing stays consistent and avoids browser history issues in single-page navigation.

jayasanka-sack:
This page isn’t part of a strict flow. Since the registration button lives in the navigation bar, users can land there from anywhere.

So this is a wrong move. So the current approach is the natural UX. Thanks for the PR.

PR #2384, opened 2026-03-22
(fix) O3-5534: Replace index-based keys in address suggestions

Pratikdeb07:
## Requirements

## Summary
This PR fixes an issue in the address search autocomplete component where `index` was used as the React `key` for rendering suggestion items.

Using `index` as a key in a dynamically updated list (filtered on each keystroke) can cause React to incorrectly reuse DOM elements across renders. This may lead to:
- Incorrect suggestion being selected on click
- Wrong item highlighted
- UI inconsistencies during typing

### Changes made:
- Replaced `key={index}` with `key={address}` (stable string value)
- Ensured each suggestion item has a consistent and unique key

Since `addressOptions` is derived from a `Set<string>`, each value is already unique and suitable for use as a React key.

This fix ensures correct component reconciliation and stable UI behavior.

## Screenshots
N/A (No visual UI changes, but fixes incorrect behavior during interaction)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5534
## Other
This change follows React best practices by avoiding index-based keys in dynamic lists and improves reliability of the autocomplete component.


jayasanka-sack (approved):
Looks good, since it comes from a set, it guarantees to be unique. Thanks @Pratikdeb07!

PR #2394, opened 2026-03-23
(fix) O3-5543: Prevent historical time scheduling and rescheduling

danirudp:
## Requirements

## Summary
**Problem:** The Appointment Form lacked client-side temporal validation for the current day. This allowed users to successfully schedule new appointments, or reschedule existing ones, to times that had already passed today.

**Solution:**
* Implemented a "Smart Time Guard" inside the `appointmentsFormSchema` using Zod's `.superRefine` for cross-field validation.
* Utilized `dayjs` with `'minute'` granularity to accurately compare selected times against the current time, preventing edge-case bugs where users are blocked from booking in the current minute due to seconds misalignments.
* Added conditional logic (`isTimeChanged`) to safely bypass the time check if a user is simply editing metadata (like Notes) on a historical appointment without actually altering the time.
* Elevated the `isEditing` variable to the top of the `AppointmentsForm` scope to ensure the validation schema had proper visibility without throwing TypeScript initialization errors.
* **Architecture Preservation:** Safely merged this new feature with the existing validation guards, ensuring the previous duration limits (must be > 0 and <= 1440 minutes) remain 100% intact.

## Screenshots
**Before (Bug)**

https://github.com/user-attachments/assets/49dd3c39-d69e-424e-9531-744cc9784aeb

https://github.com/user-attachments/assets/275142a8-3cba-4c15-9726-498d056f252c

**After (Fix)**

https://github.com/user-attachments/assets/cb031c58-4e90-4070-9b23-375fe5a6f6b9

https://github.com/user-attachments/assets/4c02c663-040f-463c-87bd-dd8b57829629

## Related Issue
https://openmrs.atlassian.net/browse/O3-5543

## Other
- **Refactoring:** Extracted the time-to-datetime conversion logic into a shared `buildAppointmentDateTime` helper function to keep the codebase DRY and maintainable.
- **Dependency Cleanup:** Resolved duplicate import bindings and restored the original import order as per maintainer feedback.
- **Test Stability:** Implemented `jest.useFakeTimers()` and `jest.setSystemTime()` in the test suite to "freeze" the execution time to 2023. This prevents future build failures caused by the test data's hardcoded dates falling into the "past" relative to the real-world clock.
- **Workflow Support:** Ensured "All Day" appointments bypass temporal validation to support retrospective data entry in low-resource clinical settings.


denniskigen (changes requested):
Good direction — the backend (openmrs-module-appointments) doesn't enforce past-time validation at all, so this frontend check is the only real guard. Two blocking issues to fix first.

denniskigen:
One more thing — the existing tests in `appointments-form.test.tsx` use `2024-01-04` as the appointment date across ~8 test cases. With the new past-time validation, those will all fail because the form will reject the date before `saveAppointment` is ever called. You'll need to either freeze the system clock with `jest.setSystemTime()` or use a future date in those tests.

PR #2397, opened 2026-03-24
(fix) O3-5545: Fix calendar prev/next buttons not updating the displayed month

Shashank8521:
## Requirements

## Summary

The Prev and Next buttons were updating a local `calendarSelectedDate` state inside `MonthlyHeader`, while the calendar grid and data fetching read `selectedDate` from `useSelectedDate()`. The two sources fell out of sync, so clicking Prev/Next changed the label but not the view.

**Fix:** Lift the month state up to `AppointmentsCalendarView` and pass the navigation handlers down to `MonthlyHeader`. The calendar now has a single source of truth for the displayed month, and the route-driven `selectedDate` continues to control the appointments table's date filter without unintended coupling between the two.

## Screenshots
Before

https://github.com/user-attachments/assets/2655e978-5956-4833-82e6-3dda0e53c85c

After

https://github.com/user-attachments/assets/75c1312c-1451-4276-97e3-7e04c526be10

## Related Issue
https://openmrs.atlassian.net/browse/O3-5545

### DIFF
--- a/packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
+++ b/packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
@@ -1,6 +1,5 @@
-import React from 'react';
-import dayjs from 'dayjs';
-import { useParams } from 'react-router-dom';
+import React, { useEffect, useState } from 'react';
+import dayjs, { type Dayjs } from 'dayjs';
 import { useTranslation } from 'react-i18next';
 import { useAppointmentsCalendar } from '../hooks/useAppointmentsCalendar';
 import AppointmentsHeader from '../header/appointments-header.component';
@@ -11,13 +10,23 @@ import { useSelectedDate } from '../hooks/useSelectedDate';
 const AppointmentsCalendarView: React.FC = () => {
   const { t } = useTranslation();
   const selectedDate = useSelectedDate();
-  const { calendarEvents } = useAppointmentsCalendar(dayjs(selectedDate).toISOString(), 'monthly');
+  const [calendarSelectedDate, setCalendarSelectedDate] = useState<Dayjs>(dayjs(selectedDate));
+
+  useEffect(() => {
+    setCalendarSelectedDate(dayjs(selectedDate));
+  }, [selectedDate]);
+
+  const { calendarEvents } = useAppointmentsCalendar(calendarSelectedDate.toISOString(), 'monthly');
 
   return (
     <div data-testid="appointments-calendar">
       <AppointmentsHeader title={t('calendar', 'Calendar')} />
       <CalendarHeader />
-      <MonthlyCalendarView events={calendarEvents} />
+      <MonthlyCalendarView
+        events={calendarEvents}
+        calendarSelectedDate={calendarSelectedDate}
+        setCalendarSelectedDate={setCalendarSelectedDate}
+      />
     </div>
   );
 };
--- a/packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
+++ b/packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import dayjs, { type Dayjs } from 'dayjs';
 import { useTranslation } from 'react-i18next';
 import { useAppointmentsCalendar } from '../hooks/useAppointmentsCalendar';
@@ -11,10 +11,6 @@ const AppointmentsCalendarView: React.FC = () => {
   const { t } = useTranslation();
   const selectedDate = useSelectedDate();
   const [calendarSelectedDate, setCalendarSelectedDate] = useState<Dayjs>(dayjs(selectedDate));
-
-  useEffect(() => {
-    setCalendarSelectedDate(dayjs(selectedDate));
-  }, [selectedDate]);
 
   const { calendarEvents } = useAppointmentsCalendar(calendarSelectedDate.toISOString(), 'monthly');
 
--- a/packages/esm-appointments-app/src/calendar/monthly/monthly-calendar-view.component.tsx
+++ b/packages/esm-appointments-app/src/calendar/monthly/monthly-calendar-view.component.tsx
@@ -1,29 +1,72 @@
-import React from 'react';
-import dayjs from 'dayjs';
+import React, { useCallback } from 'react';
 import isBetween from 'dayjs/plugin/isBetween';
+import dayjs, { type Dayjs } from 'dayjs';
+import { Button } from '@carbon/react';
+import { formatDate } from '@openmrs/esm-framework';
+import { useTranslation } from 'react-i18next';
 import { type DailyAppointmentsCountByService } from '../../types';
 import { monthDays } from '../../helpers';
+import DaysOfWeekCard from './days-of-week.component';
 import MonthlyViewWorkload from './monthly-workload-view.component';
-import MonthlyHeader from './monthly-header.component';
-import { useSelectedDate } from '../../hooks/useSelectedDate';
 import styles from '../appointments-calendar-view-view.scss';
+import headerStyles from './monthly-header.scss';
 
 dayjs.extend(isBetween);
 
+const DAYS_IN_WEEK = ['SUN', 'MON', 'TUE', 'WED', 'THUR', 'FRI', 'SAT'];
+
 interface MonthlyCalendarViewProps {
   events: Array<DailyAppointmentsCountByService>;
+  calendarSelectedDate: Dayjs;
+  setCalendarSelectedDate: React.Dispatch<React.SetStateAction<Dayjs>>;
 }
 
-const MonthlyCalendarView: React.FC<MonthlyCalendarViewProps> = ({ events }) => {
-  const selectedDate = useSelectedDate();
+const MonthlyCalendarView: React.FC<MonthlyCalendarViewProps> = ({
+  events,
+  calendarSelectedDate,
+  setCalendarSelectedDate,
+}) => {
+  const { t } = useTranslation();
+
+  const handleSelectPrevMonth = useCallback(() => {
+    setCalendarSelectedDate((prev) => prev.subtract(1, 'month'));
+  }, [setCalendarSelectedDate]);
+
+  const handleSelectNextMonth = useCallback(() => {
+    setCalendarSelectedDate((prev) => prev.add(1, 'month'));
+  }, [setCalendarSelectedDate]);
 
   return (
     <div className={styles.calendarViewContainer}>
-      <MonthlyHeader />
+      <>
+        <div className={headerStyles.container}>
+          <Button
+            aria-label={t('previousMonth', 'Previous month')}
+            kind="tertiary"
+            onClick={handleSelectPrevMonth}
+            size="sm">
+            {t('prev', 'Prev')}
+          </Button>
+          <span>{formatDate(calendarSelectedDate.toDate(), { day: false, time: false, noToday: true })}</span>
+          <Button aria-label={t('nextMonth', 'Next month')} kind="tertiary" onClick={handleSelectNextMonth} size="sm">
+            {t('next', 'Next')}
+          </Button>
+        </div>
+        <div className={headerStyles.workLoadCard}>
+          {DAYS_IN_WEEK.map((day) => (
+            <DaysOfWeekCard key={day} dayOfWeek={day} />
+          ))}
+        </div>
+      </>
       <div className={styles.wrapper}>
         <div className={styles.monthlyCalendar}>
-          {monthDays(dayjs(selectedDate)).map((dateTime, i) => (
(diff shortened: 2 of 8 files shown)


packages/esm-appointments-app/src/calendar/monthly/monthly-header.component.tsx
```diff
@@ -1,27 +1,31 @@
...
-  const [calendarSelectedDate, setCalendarSelectedDate] = useState(dayjs(selectedDate));
+  const calendarSelectedDate = dayjs(selectedDate);
 
   const handleSelectPrevMonth = useCallback(() => {
-    setCalendarSelectedDate(calendarSelectedDate.subtract(1, 'month'));
-  }, [calendarSelectedDate, setCalendarSelectedDate]);
+    navigate({
+      to: `${spaHomePage}/appointments/calendar/${calendarSelectedDate.subtract(1, 'month').format('YYYY-MM-DD')}`,
+    });
+  }, [calendarSelectedDate]);
```
vivektarun: Minor: setCalendarSelectedDate has been removed from the dependency array, which is correct since the state is gone. But the useCallback here adds no real benefit now since calendarSelectedDate is derived on every render anyway. This could just be a regular function.

vivektarun (dismissed):
Try to figure out these.

chibongho (approved):
Thanks!

denniskigen (commented):
Hey @Shashank8521 — thanks for digging into this! You correctly identified the root cause: the old `MonthlyHeader` had local state that diverged from the route-driven `selectedDate`. 

However, the fix here re-introduces the same pattern — `calendarSelectedDate` state in `AppointmentsCalendarView` that can drift from the route. The simpler fix is to just use route navigation in `MonthlyHeader` (which you already did!) and keep `MonthlyCalendarView` using it. No local state needed, no prop drilling, no duplicated header markup.

See the inline comments for specifics.

packages/esm-appointments-app/src/calendar/monthly/monthly-calendar-view.component.tsx
```diff
@@ -1,29 +1,72 @@
...
       <div className={styles.wrapper}>
         <div className={styles.monthlyCalendar}>
-          {monthDays(dayjs(selectedDate)).map((dateTime, i) => (
-            <MonthlyViewWorkload key={i} dateTime={dateTime} events={events} />
+          {monthDays(calendarSelectedDate).map((dateTime, i) => (
+            <MonthlyViewWorkload
+              key={i}
+              dateTime={dateTime}
+              events={events}
+              calendarSelectedDate={calendarSelectedDate}
```
denniskigen: This inlines a copy of `MonthlyHeader` and drives it from local state, but the route-based fix you made to `MonthlyHeader` already solves the problem. Keep using `MonthlyHeader` and let `useSelectedDate()` drive the calendar:

```suggestion
import React from 'react';
import dayjs from 'dayjs';
import isBetween from 'dayjs/plugin/isBetween';
import { type DailyAppointmentsCountByService } from '../../types';
import { monthDays } from '../../helpers';
import MonthlyViewWorkload from './monthly-workload-view.component';
import MonthlyHeader from './monthly-header.component';
import { useSelectedDate } from '../../hooks/useSelectedDate';
import styles from '../appointments-calendar-view-view.scss';

dayjs.extend(isBetween);

interface MonthlyCalendarViewProps {
  events: Array<DailyAppointmentsCountByService>;
}

const MonthlyCalendarView: React.FC<MonthlyCalendarViewProps> = ({ events }) => {
  const selectedDate = useSelectedDate();

  return (
    <div className={styles.calendarViewContainer}>
      <MonthlyHeader />
      <div className={styles.wrapper}>
        <div className={styles.monthlyCalendar}>
          {monthDays(dayjs(selectedDate)).map((dateTime, i) => (
            <MonthlyViewWorkload key={i} dateTime={dateTime} events={events} />
          ))}
        </div>
      </div>
    </div>
  );
};

export default MonthlyCalendarView;
```

This keeps the component identical to the original except that `MonthlyHeader` now does route navigation internally.

packages/esm-appointments-app/src/calendar/monthly/monthly-workload-view.component.tsx
```diff
@@ -7,18 +7,22 @@ import { spaHomePage } from '../../constants';
 import { isSameMonth } from '../../helpers';
 import { type DailyAppointmentsCountByService } from '../../types';
 import MonthlyWorkloadViewExpanded from './monthly-workload-view-expanded.component';
-import { useSelectedDate } from '../../hooks/useSelectedDate';
 import styles from './monthly-view-workload.scss';
 
 export interface MonthlyWorkloadViewProps {
   events: Array<DailyAppointmentsCountByService>;
   dateTime: Dayjs;
```
denniskigen: The `calendarSelectedDate` prop threading through here and into `MonthlyWorkloadViewExpanded` becomes unnecessary if we keep using `useSelectedDate()`. This file can stay unchanged from main.

omsrivastava512:
hey @Shashank8521, any updates on this?

packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
```diff
@@ -11,13 +10,23 @@ import { useSelectedDate } from '../hooks/useSelectedDate';
 const AppointmentsCalendarView: React.FC = () => {
   const { t } = useTranslation();
   const selectedDate = useSelectedDate();
-  const { calendarEvents } = useAppointmentsCalendar(dayjs(selectedDate).toISOString(), 'monthly');
+  const [calendarSelectedDate, setCalendarSelectedDate] = useState<Dayjs>(dayjs(selectedDate));
+
+  useEffect(() => {
```
omsrivastava512: `calendarSelectedDate` should be held as local state initialized from `selectedDate`. Syncing it with the URL creates the exact bug, re-couples the two things you were trying to decouple - table view and calendar view.

packages/esm-appointments-app/src/calendar/appointments-calendar-view.component.tsx
```diff
@@ -11,13 +10,23 @@ import { useSelectedDate } from '../hooks/useSelectedDate';
...
 
   return (
     <div data-testid="appointments-calendar">
       <AppointmentsHeader title={t('calendar', 'Calendar')} />
       <CalendarHeader />
-      <MonthlyCalendarView events={calendarEvents} />
+      <MonthlyCalendarView
+        events={calendarEvents}
+        calendarSelectedDate={calendarSelectedDate}
+        setCalendarSelectedDate={setCalendarSelectedDate}
```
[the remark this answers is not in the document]
omsrivastava512: The approach @denniskigen suggested works if the calendar is a fully isolated view, but since `AppointmentsHeader` consumes `useSelectedDate()`, then navigating months via the route would silently change the table's date filter, which is the UX concern @chibongho raised. Could you confirm what sits with you?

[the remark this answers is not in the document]
omsrivastava512: Let me explain the issue better using this scenario:

Imagine a receptionist is looking at today's appointments (April 14). They want to check if May has any bookings.

If the receptionist clicks "Next Month," the URL changes to `/calendar/2026-06-14`. The calendar shows May, but today's appointment list vanishes because the table is now trying to show appointments for May 14th.

https://github.com/user-attachments/assets/a6f80ee5-bf8d-4a0f-b361-7fbe676970f5

In this video, the user had no intention to change any dates but they were implicitly changed only as a side effect of exploring the calendar. This can be very confusing for the user.

If we keep state local, on clicking "Next Month," only the local state changes. The calendar shows May, but the URL stays on April 14. Today's appointment list stays on the screen.

omsrivastava512 (changes requested):
> So the `:date` param in the route, be it `/:date` or `/calendar/:date`, sets the date filter of the main appointments table.

@chibongho from what I understand so far, the shared `useSelectedDate()` inside the `<CalendarHeader/>` is the cause behind this. The hook just grabs the `date` from whatever the current URL is. It doesn't care whether you're on /calendar/:date or /:date.  Could we add an exception clause, something like, 

```
  const isCalendarRoute = location.pathname.includes('/calendar');
  ...
  if (isCalendarRoute) return  dayjs().startOf('day').format('YYYY-MM-DD');
```
  
  A bit hacky (and I personally don't like this approach)
  
The cleanest approach IMO is keeping the `date` as a local state without touching the URL. Calendar View exists only to lookup for appointments in a compact way. It has no utility being coupled with navigation.

Shashank8521:
Thanks @omsrivastava512 for the review.

You're right — the `useEffect` was re-coupling the local calendar month to the route after initialization.
I originally kept it because the page was remounting anyway, so the issue didn’t as seen as such.

I’ve removed it so `calendarSelectedDate` is now only initialized from `selectedDate` on mount. This keeps the calendar month independent after opening and allows it to maintain its own state, as suggested by @chibongho, while still allowing remount-based reinitialization when needed. can you review?

I’m also happy to add tests for this, as discussed with @denniskigen, if this approach looks good — or make any further changes if needed.

Sorry for the late reply @chibongho — I was waiting for @denniskigen’s review.

packages/esm-appointments-app/src/calendar/monthly/monthly-calendar-view.component.tsx
omsrivastava512: hey @Shashank8521, I think Dennis's comment here still remains valid. Instead of using inline JSX, which makes it redundant with the `MonthlyView` please make the changes within `MonthlyView` component keeping this modular.

packages/esm-appointments-app/src/calendar/monthly/monthly-header.component.tsx
```diff
@@ -1,27 +1,31 @@
...
-import { formatDate } from '@openmrs/esm-framework';
+import { formatDate, navigate } from '@openmrs/esm-framework';
 import DaysOfWeekCard from './days-of-week.component';
 import styles from './monthly-header.scss';
+import { spaHomePage } from '../../constants';
 import { useSelectedDate } from '../../hooks/useSelectedDate';
 
 const DAYS_IN_WEEK = ['SUN', 'MON', 'TUE', 'WED', 'THUR', 'FRI', 'SAT'];
 
 const MonthlyHeader: React.FC = () => {
```
omsrivastava512: This component remains unutilised as we aren't using it anywhere else. Also, the navigation-based implementation is redundant to the state-based one we just did. I suggest passing the `setCalendarSelectedDate` here as prop and using this component in `mcvc.tsx`, or better pass the handler functions to keep things simple, both by keeping the handling responsibility to the parent and keeping the prop-types simple.

omsrivastava512 (changes requested):

denniskigen:
Thanks for working on this, @Shashank8521! The root cause identification is correct — the prev/next buttons were updating local state that nothing read.

After discussing with @chibongho, we've agreed that the `:date` route param should only control the appointments table date filter, not the calendar. This aligns with the concern @omsrivastava512 raised — navigating months in the calendar shouldn't change the URL or silently shift the table's date. The backend's `/appointment/appointmentSummary` endpoint is agnostic about where the dates come from.

That means the fix should use **local state** in `AppointmentsCalendarView`, which is the direction this PR takes. To address the remaining feedback from @omsrivastava512 and @vivektarun:

1. **Drop the route navigation changes in `MonthlyHeader`** — Since we're going with local state, the `navigate()` calls are unnecessary and would reintroduce the coupling we're trying to avoid.

2. **Keep `MonthlyHeader` as a component instead of inlining its markup** — As @omsrivastava512 noted, the PR duplicates the header JSX into `MonthlyCalendarView`. Instead, pass `calendarSelectedDate` and the navigation handlers down to `MonthlyHeader` as props.

3. **Remove `useSelectedDate()` from calendar components** — Components like `MonthlyWorkloadView` should use the `calendarSelectedDate` prop instead of `useSelectedDate()`. This is already correct in the PR.

4. **Use functional updater form for state** — The `setCalendarSelectedDate((prev) => prev.add(1, 'month'))` pattern is correct and avoids stale closure issues.

Shashank8521:
Thanks @denniskigen  for clarifying the intended architecture! I’ll update this.

Shashank8521:
I've addressed the previously mentioned points:
 
1. Removed route-based navigation from MonthlyHeader and kept the calendar fully driven by local state

2. Kept MonthlyHeader as a separate component and passed calendarSelectedDate along with navigation handlers via props

And, I noticed the calendar UI/UX improvements discussed in [O3-5617.](https://openmrs.atlassian.net/browse/O3-5617) and [O3-5616 ](https://openmrs.atlassian.net/browse/O3-5616). I’ve started working on those styling and layout refinements. Would you prefer that I include those changes in this PR, or open a separate PR for them? I have a draft ready and can open a separate PR if preferred.

Could you please review the changes? Please let me know if anything still needs adjustment or a test is required.@denniskigen @chibongho @jayasanka-sack

Thanks!

omsrivastava512 (approved):
LGTM.

denniskigen (approved):
Thanks, @Shashank8521!

PR #2399, opened 2026-03-24
(fix) O3-5548: Replace deprecated substr() with substring and handle edge cases

Pratikdeb07:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. The title follows the conventional commit format.

## Summary
This PR replaces the deprecated `String.prototype.substr()` with `String.prototype.substring()` in `visit.resource.ts`.

The `substr()` method is deprecated in modern JavaScript and may be removed in future environments. Additionally, it uses a different parameter signature (`start, length`) compared to `substring()` (`start, end`), which can lead to confusion and potential bugs.

The implementation updates the logic to use `substring()` with correctly calculated start and end indices. It also ensures safe handling of edge cases where no non-digit character is found when parsing the concentration value.

Additionally, the implementation safely handles the case where no non-digit character is found while parsing the concentration value. Since search() returns -1 in such scenarios, the logic ensures the substring extends to the end of the string, preventing incorrect index calculations or runtime issues.
## Screenshots
N/A (No UI changes)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5548

## Other
- Improves compatibility with modern JavaScript standards
- Ensures future-proof implementation
- Adds safer index handling for edge cases
- No functional changes introduced

### DIFF
--- a/packages/esm-active-visits-app/src/visits-summary/visit.resource.ts
+++ b/packages/esm-active-visits-app/src/visits-summary/visit.resource.ts
@@ -45,8 +45,10 @@ export function getDosage(strength: string, doseNumber: number) {
   if (concentrationStartIndex >= 0) {
     strengthUnits = strength.substring(i, concentrationStartIndex);
     const j = strength.substring(concentrationStartIndex + 1).search(/\D/);
-    const concentrationQuantity = parseInt(strength.substr(concentrationStartIndex + 1, j));
-    const concentrationUnits = strength.substring(concentrationStartIndex + 1 + j);
+    const start = concentrationStartIndex + 1;
+    const end = j === -1 ? strength.length : start + j;
+    const concentrationQuantity = parseInt(strength.substring(start, end));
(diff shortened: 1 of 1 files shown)


IvanescuChristian (commented):
The conversion looks good. The mapping is done properly and all and all it s a genuine improvement.
The substr(start,-1) would return an empty string whilst the proposed code captures all the remaning digits to the end. Great find!

vivektarun (changes requested):
Good fix for the deprecated `substr()` and the edge case handling with `j === -1` is a 
Nice improvement.

jayasanka-sack (changes requested):
Thanks for the PR, @Pratikdeb07!

A test like, `expect(getDosage('500mg/5', 1)).toBe(...)` would make this PR self-justifying.
Without it the "fixes edge cases" claim in the title isn't anchored to anything the CI can verify.

PR #31, opened 2021-08-04
MF-658: set default DOB to blank

walteronoh:
### Description
Ticket [MF-658](https://issues.openmrs.org/browse/MF-658)
Sets the default dob to blank.

PR #2402, opened 2026-03-25
(fix) O3-5551: Improve queue entry error handling

IvanescuChristian:
## Requirements

## Summary
When adding a patient to queue fails with a 400 response (e.g. "Queue entry cannot start before the visit start date"), the app crashes with React error #31 (Objects are not valid as a React child).

**Root cause:** At least 2 parents use QueueFields. One catches the re-thrown error (`add-patient-to-queue.component.tsx`), the other doesn't (`existing-visit-form.component.tsx`). The missing `.catch()` causes an unhandled rejection that crashes React.

**Fix (3 changes):**
    1. Replaced manual error extraction with existing `getErrorMessage()` and `isDuplicateQueueEntryError()` utils from `queue-entry-error.utils.ts` as suggested by @denniskigen
    2. Changed `throw error` to `throw new Error(getErrorMessage(error))` so the re-thrown error is a clean Error with a string message — the promise still rejects so parent forms stay open on failure
    3. Added `.catch()` in `existing-visit-form.component.tsx` since this parent was missing error handling — the error is already handled by QueueFields via snackbar

## Screenshots

CurrentServiceError
CorrectedServiceError
NewCorrectedOutput
ServiceQueueErrorProof

## Related Issue
https://openmrs.atlassian.net/browse/O3-5551


denniskigen (changes requested):
Thanks for looking into this — the error handling here does need improvement. A couple of things to address:

denniskigen (changes requested):
Thanks for looking into this — removing the `throw error` at the end of the catch block is the right fix for the React crash. But the `await error?.responseBody` change is based on a wrong assumption.

ranmori (commented):
great stuff , just wanna  check; since throw new Error(getErrorMessage(error)) re-throws a plain Error rather than the original OpenmrsFetchError, any parent that was checking error?.responseBody downstream would lose that context. Is there a case where a parent form needs access to the full error shape after the re-throw, or are all consumers just checking for the string message?

PR #2408, opened 2026-03-25
(fix) O3-5525: Fix URL decoding in queue services table

Nab104:
## Requirements

## Summary

This PR fixes an issue in the queue services table where URL decoding was handled incorrectly.

Previously, the implementation used `window.location.pathname.replace('%20', ' ')`, which only replaces the first occurrence of `%20`. This caused incorrect parsing of multi-word service names (e.g., `Primary%20Health%20Care`), affecting variables like `service` and `locationUuid`.

## Changes

- Replaced manual string replacement with `decodeURIComponent(window.location.pathname)`
- Updated `service` and `locationUuid` to use `const`
- Ensured correct parsing of multi-word service names from the pathname

## Screenshots

N/A (No UI changes)

## Related Issue

O3-5525: https://openmrs.atlassian.net/browse/O3-5525

## Other

This change ensures proper decoding of all URL-encoded characters and improves reliability when parsing pathname segments in the queue services table.


denniskigen (dismissed):
Good fix — `replace('%20', ' ')` only handles the first occurrence, so multi-word service names would break path parsing. `decodeURIComponent` is the right approach. LGTM.

packages/esm-service-queues-app/src/queue-patient-linelists/queue-services-table.component.tsx
```diff
@@ -6,9 +6,11 @@ import QueuePatientBaseTable from './queue-linelist-base-table.component';
 const ServicesTable: React.FC = () => {
   const { t } = useTranslation();
 
-  const currentPathName: string = window.location.pathname.replace('%20', ' ');
-  let service = currentPathName.split('/')[6];
-  let locationUuid = currentPathName.split('/')[8];
+  const currentPathName: string = decodeURIComponent(window.location.pathname);
```
denniskigen: One thing to watch out for: decoding the full pathname before splitting means any encoded slash in a service name (e.g. `ANC%2FFollow-up`) becomes a real `/` and shifts the segment indices. Safer to split first, then decode the individual segments:

```suggestion
  const pathSegments = window.location.pathname.split('/');
  const service = decodeURIComponent(pathSegments[6]);
  const locationUuid = pathSegments[8];
```

This is unlikely to hit in practice (service names rarely contain slashes), but it's the more correct approach.

packages/esm-service-queues-app/src/queue-patient-linelists/queue-services-table.component.tsx
```diff
@@ -6,9 +6,10 @@ import QueuePatientBaseTable from './queue-linelist-base-table.component';
 const ServicesTable: React.FC = () => {
   const { t } = useTranslation();
 
-  const currentPathName: string = window.location.pathname.replace('%20', ' ');
-  let service = currentPathName.split('/')[6];
-  let locationUuid = currentPathName.split('/')[8];
+  const pathSegments = window.location.pathname.split('/');
+  const service = decodeURIComponent(pathSegments[6]);
```
ashuverma25, line 10: Ensuring correct URL decoding in the service tables is a crucial fix for localized environments where special characters are common. The implementation correctly handles these cases using decodeURIComponent and keeps the UI data-consistent.

ashuverma25 (commented):
I've reviewed the URL decoding logic for the queue services table. The changes look solid and correctly address data consistency issues for O3.

jayasanka-sack (commented):
This fixes the issue, but can't we just use `useParams()`? this hardcoded indices seems super fragile.

jayasanka-sack:
Alright, a few things: the component was written pretty poorly to begin with. For example, using hardcoded indices to access specific parts of the URL instead of `useParams` is a bad idea, and it’s also unnecessary to pass both the service name and the service UUID.

I went through the code and figured out that the component is orphaned. Nothing links to it anymore, and the link was removed in 56580824. Even before it was removed, the link was already wrapped in /* TODO: Uncomment this when functionality of the patient list works */, so it’s been dead for a while.

Closing this PR! Thanks for the contribution, will open a followup PR to get the component removed.

PR #2428, opened 2026-03-30
(fix) O3-5559: Truncate previous queue entry time for actual validation

IvanescuChristian:
## Requirements

## Summary

When editing a queue entry's start time, the time picker only accepts 
hours and minutes (HH:MM) but the if statement used to compare against a timestamp 
with seconds precision. This created impossible validation windows where 
no valid time could be entered (e.g. previous entry at 5:31:59 PM — 
user enters 5:31 → rejected as "before previous entry", enters 5:32 → 
rejected as "in the future").

**Fix:** Truncate the previous queue entry's start time downwards and end up comparing HH:MM with HH:MM, and use < instead of <= (as equals is just a now check which is pointless). The error message now also displays the truncated 
time ( the actual checked time ) .

## Screenshots
1time_before_error
2time_after_error
3time_in_future_is_in_past
4_time_before_error2
time_after_success

## Related Issue
https://openmrs.atlassian.net/browse/O3-5559


ranmori (commented):
Brilliant!!! The updated error message now showing the actual checked time is a nice UX improvement too. One thought: is there a test case covering the boundary scenario (e.g. previous entry at HH:MM:59) to ensure the floor truncation is consistent across time zones or locales where time formatting might differ?

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -122,7 +122,7 @@ const QueueFields = React.memo(({ setOnSubmit, defaultInitialServiceQueue }: Que
           mutateQueueEntries();
         })
         .catch((error) => {
-          const errorMessage = error?.responseBody?.error?.message || error?.message || '';
+          const errorMessage = error?.responseBody?.error?.message || error?.message || ''; // this section of code is undefined
```
jayasanka-sack: Let's get this comment removed. It's not related.

jayasanka-sack (changes requested):
Good one! Thanks @IvanescuChristian!

It seems there are no proper unit test coverage for this one. 

Can you add some tests, especially:

- Previous entry at HH:MM:SS with seconds, user enters HH:MM -> should be allowed. 
- User enters a time strictly before the truncated minute-> should show the error.
- previousQueueEntryStartTime is null/undefined -> validation skips, no error. 

Let me know once you update the PR with tests, we can get this merged.

IvanescuChristian:
> Good one! Thanks @IvanescuChristian!
> 
> It seems there are no proper unit test coverage for this one.
> 
> Can you add some tests, especially:
> 
>     * Previous entry at HH:MM:SS with seconds, user enters HH:MM -> should be allowed.
> 
>     * User enters a time strictly before the truncated minute-> should show the error.
> 
>     * previousQueueEntryStartTime is null/undefined -> validation skips, no error.
> 
> 
> Let me know once you update the PR with tests, we can get this merged.

@jayasanka-sack 
I have updated the PR with the required tests which pass. Awaiting confirmation and merging.

(Edit) If you find the time I would highly appreciate if you would take the time to investigate another PR of mine : https://github.com/openmrs/openmrs-esm-patient-management/pull/2402
         Thank you for your time , until next , VeryFriendlySolver

jayasanka-sack (approved):
Thanks for fixing this, @IvanescuChristian!

PR #2446, opened 2026-03-31
(fix) O3-5553: Type-safe bed management app

LowSugarCoke:
## Requirements

## Summary
Replaces unsafe `any` types in the bed management app with explicit TypeScript types.
  - `useBedType` in `bed-form.resource.ts`: typed the `useSWR` generic as `{ data: { results: Array<BedType> } }` and explicitly typed the `uniqueBedTypes` accumulator as `Array<BedType>`
  - `useBedsGroupedByLocation` in `summary.resource.ts`: typed the `result` state as `Array<Array<Bed & { location: Location }>>` and imported the `Location` type

## Screenshots
N/A

## Related Issue
https://openmrs.atlassian.net/browse/O3-5531
https://openmrs.atlassian.net/browse/O3-5553

## Other
This PR is related to the following PR:
https://github.com/openmrs/openmrs-esm-patient-management/pull/2442
That PR addressed unsafe `any` types in the test code, while this PR focuses on fixing implicit `any` types in the implementation.


omsrivastava512 (changes requested):
Good PR overall. Two things before this merges: extract that nested intersection into a named type alias; and answer whether this is all to the O3-5553 and there are no more implicit or explicit `any` typed variables (I can find plenty in the same file itself).

- Try running `grep` to find any explicit `any` declarations. 
- Use this to rule out the implicit ones:

  ```shell
  tsc --noImplicitAny --noEmit
  ```

- or try having the `noImplicitAny` flag enabled in the tsconfig (remember not to include it in your commit).

Requesting changes.

janithRS (commented):
Thanks @LowSugarCoke for the PR! Shall we also override the eslint for this module with no-explicit-any to error so that we won't face this issue again.

You can see the same change has been done to a diff module in here https://github.com/openmrs/openmrs-esm-patient-management/pull/2379 :)

packages/esm-bed-management-app/src/bed-administration/form/bed-form.resource.ts
omsrivastava512: This might give an `unknown` type error. Try a conditional statement here. 
Something like:
```
throw new Error(`...${error instanceof Error ? error.message : String(error)}`);
```

`String()` safely coerces anything to a readable string.

packages/esm-bed-management-app/src/summary/summary.resource.ts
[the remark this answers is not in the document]
omsrivastava512: This should work.
```
const updatedWards = (await Promise.all(promises)).filter(
          (ward): ward is BedWithLocation[] => ward !== null
        ); 
if (isSubscribed) {
          setResult(updatedWards);
        }
```

omsrivastava512: `type` is a string, add a fallback blank string to avoid type error coming from optional chaining here.

packages/esm-bed-management-app/src/left-panel-link.component.tsx
```diff
@@ -15,7 +15,7 @@ function LinkExtension({ config }: { config: LinkConfig }) {
 
   let urlSegment = useMemo(() => decodeURIComponent(last(location.pathname.split('/'))), [location.pathname]);
```
omsrivastava512: `last()` returns `string | undefined`, but `urlSegment` is used as `string`. Perhaps use null coalescing.

omsrivastava512 (changes requested):
I've pointed some other TS issues that were missed earlier. Please look into them. Nothing else I could spot in the touched files. Thank you.

packages/esm-bed-management-app/src/summary/summary.resource.ts
LowSugarCoke: Done

PR #2447, opened 2026-03-31
fix: add null guard for startedAt in QueueDuration component

ranmori:
## Requirements

## Summary

While performing a technical audit of the esm-service-queues-app for GSoC 2026, I identified a potential UI crash/invalid rendering in the QueueDuration component.
Currently, if a QueueEntry is initialized or fetched without a valid startedAt timestamp, the dayjs().diff() calculation can return NaN or invalid strings. This PR adds a defensive guard clause to the DurationString sub-component to return a placeholder (--) when the timestamp is missing, ensuring UI stability.

## Screenshots
This is a logic-level fix to prevent invalid text (NaN) from appearing in the Duration column of the Service Queue table.

## Related Issue
Jira access currently pending approval (Requested March 30, 2026). This fix addresses a stability edge-case found during GSoC proposal research for the 'Service Queues Improvements' project

## Other
This fix was tested locally against the dev3 environment to ensure the component renders the placeholder correctly when startedAt is null."

### DIFF
--- a/packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
+++ b/packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
@@ -13,6 +13,12 @@ type QueueEntryResponse = FetchResponse<{
   }>;
   totalCount: number;
 }>;
+// // TODO (GSoC 2026 - Performance): This repString causes severe over-fetching.
+// For each patient, it fetches full concept objects for status/priority (including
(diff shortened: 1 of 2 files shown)


jayasanka-sack:
Thanks for the PR! I'm closing this because `QueueEntry.startedAt` is non-optional, so the edge case this addresses isn't actually possible. Good instinct on defensive coding tho. feel free to look for other areas where it might apply. 💛

PR #2450, opened 2026-03-31
(fix) O3-5579: Isolate appointment cache per patient to prevent cross-patient data leakage

omeriinnocent:
## Requirements

## Summary
This pull request improves how the `usePatientAppointments` and `usePatientAppointmentHistory` hooks manage data fetching by updating their usage of SWR cache keys to depend on both patient and date parameters. This ensures that appointment data is properly refetched when either the patient or date changes. Additionally, a new test has been added to verify this behavior.

**SWR cache key improvements:**
* Updated the SWR cache key in `usePatientAppointments` (in both `queue-linelist.resource.ts` and `patient-appointments.resource.ts`) to include `patientUuid` and `startDate`, ensuring data is fetched separately for each patient and date combination. [[1]](diffhunk://#diff-e6f4d88746bb55e26e809414ed7d5385d3eb903ebc104710493728b064981e7aL67-R67) [[2]](diffhunk://#diff-d9c88febf7b56b3c3dda9f9fa573553c0da73372ffd6586526a792dde6facaa1L29-R29)
* Updated the SWR cache key in `usePatientAppointmentHistory` to include `patientUuid` and `selectedDate`, improving data separation and refetching logic.

**Testing enhancements:**
* Added a new test (`queue-linelist.resource.test.tsx`) to verify that `usePatientAppointments` fetches data separately when patient or date values change.
Added a regression test to verify that usePatientAppointments (and related resources) trigger distinct SWR fetches when input parameters like patientUuid and startDate change. This ensures data is not incorrectly reused across different queries. All related tests pass successfully.

## Screenshots
image
## Related Issue
https://openmrs.atlassian.net/browse/O3-5579


revert_commit.md
```diff
@@ -0,0 +1,5 @@
+This reverts commit b7cf0e1712de4d4f383695393294013a951840a3.
```
ELVIS-KATO: I see changes in this file, this should likely be removed before merging unless it's an  intentional documentation.

omsrivastava512 (approved):
Well, since `appointmentsSearchUrl` is a static string without params and patient context only lives in the fetcher body, SWR would have no way to differentiate between patients at the cache level - so this fix does make sense.

About the HIPAA/GDPR framing in the ticket though - slow down there. Easy. This is a stale cache flash during patient switching issue at worst, not persistent cross-patient data leakage. The fix is worth doing either way, but worth keeping the language accurate.

omsrivastava512:
hey @denniskigen I think we can get this merged. Also, @bhavya-jpg, is this what you were talking about?

omsrivastava512:
Hey @omeriinnocent, could you attach a demo video of an example of this stale state in action? That would help. Thank you.

bhavya-jpg:
@omsrivastava512 Yes, this PR fixes the exact issue I was pointing out regarding the SWR cache keys! However, I noticed that this PR only partially fixes the bug. The exact same missing dependency array issue still exists in two other hooks: usePatientAppointmentHistory.ts (in esm-appointments-app) and queue-linelist.resource.ts (in esm-service-queues-app). Would you like me to open a follow-up PR to fix the remaining files so we can fully resolve this data crossover issue repo-wide?"

omeriinnocent:
Thanks @omsrivastava512 for the keen eye.

omeriinnocent:
Thanks @bhavya-jpg , good catch surely, yes we can solve that together as we wait for the final feedback from @denniskigen .

omsrivastava512:
> Would you like me to open a follow-up PR to fix the remaining files so we can fully resolve this data crossover issue repo-wide?"

Sure, can we create a dependent PR stacked over this one?

bhavya-jpg:
@omsrivastava512 yeah we can work over a stacked pr over this one ,let wait for what dennis has to say ,I'll also try to manually see the cache keys using the SWR devtools that dennis mentioned in the meet

omsrivastava512:
Well, @bhavya-jpg or @omeriinnocent, either of you could just push the remaining changes and they'd be reviewed once for all instead of relying on a go ahead from Dennis.

omsrivastava512:
...and the SWR dev tools findings too.

bhavya-jpg:
@omsrivastava512 i am trying to work on this ,but i am facing some issue in the code ,idky there is some issue coming up in my local server

omeriinnocent:
@bhavya-jpg , @omsrivastava512 hope this solves the issues,could there be any left? happy to get your feedback.

jayasanka-sack (changes requested):
This is a good one! thanks @omeriinnocent. 

Please add a test confirming that different values produce separate SWR fetches, and we can get this merged.

omeriinnocent:
thanks @jayasanka-sack for the review,  i added a regression test confirming that different inputs produce distinct SWR keys and trigger separate fetches (verified via multiple openmrsFetch calls with different request bodies).

jayasanka-sack:
@omeriinnocent that’s great! Please ensure that the tests cover all the places where you’ve made changes. Adding test coverage only in one hook isn’t enough. You should cover all three hooks.

omeriinnocent:
yes @jayasanka-sack ,added a regression test to verify that usePatientAppointments (and related resources) trigger distinct SWR fetches when input parameters like patientUuid and startDate change. This ensures data is not incorrectly reused across different queries. All related tests pass successfully.

jayasanka-sack:
Hey @omeriinnocent I'm on mobile right now, and I can't check the diff properly. I fetched a review from claude. Can you verify following comments:


<hr/>

The fix is correct and the approach is right. Three call sites were keying SWR’s cache on a URL string that didn’t include patientUuid or startDate, so when those changed the cache returned stale data from the previous patient/date — exactly the cross-patient leakage in O3-5579. Switching to a tuple key [url, patientUuid, startDate] is the standard SWR pattern and resolves it cleanly. Tests are present for all three hooks and assert the right thing (separate fetches when inputs change).
Approving in spirit, with a few things worth tightening before merge.
Issues
1. Inconsistent null-guard between hooks. usePatientAppointmentHistory and patient-appointments.resource.ts both gate on patientUuid ? [...] : null, but the diff for queue-linelist.resource.ts shows only the import-side test file plus a truncated context line — I can’t see whether the same patientUuid ? guard was applied there. If it wasn’t, that hook will still fire a request with patientUuid: undefined in the body. Worth confirming the actual useSWR call in that file got the same treatment, not just the key shape.
2. The tests don’t actually prove the bug is fixed. Each test rerenders with both a new patientUuid and a new startDate, then asserts two fetches happened. That passes today, but it would also pass under the old buggy code if the URL happened to differ — and it doesn’t isolate which key component is doing the work. A stronger test: rerender with the same URL but a different patientUuid only, and assert (a) two fetches occurred and (b) the second result is not the first patient’s cached data. That’s the actual regression you’re guarding against. Same for startDate.
3. Missing the negative case. No test asserts that when patientUuid and startDate are unchanged across rerenders, only one fetch happens (i.e., caching still works). Without it, someone could “fix” a future flake by adding Date.now() to the key and the suite would stay green.
4. usePatientAppointmentHistory test depends on a mocked useSelectedDate. That’s fine, but the hook itself reads selectedDate from that module and includes it in the key — the test should also verify the date-only-change case (same patient, different selected date → new fetch). Right now both inputs change together.

omeriinnocent:
Hi @jayasanka-sack ,thanks for the review, have added queue-linelist.resource.ts now also uses the same guard: patientUuid ? [appointmentsSearchUrl, patientUuid, startDate] : null (so no request runs with patientUuid undefined).
Tests are now strengthened to cover patientUuid-only change, date-only change (incl. selectedDate for usePatientAppointmentHistory), and the cache-hit case (same inputs → no extra fetch).

bhavya-jpg (commented):
@omeriinnocent have you manually checked the SWR cache keys using the  SWR devtools after fixing things ,I'll recommend you doing that ,I was able to reproduce this issue on web and see the swr cache keys actually so I'll recommend that you manually see to it that cross  patient data lekage isnt happening

omeriinnocent:
thanks @bhavya-jpg ,this works now.

jayasanka-sack (approved):
Thanks for getting this done @omeriinnocent! Appreciate your effort and time!

omeriinnocent:
wooh, thanks so much @jayasanka-sack for your support.

PR #2470, opened 2026-04-06
(BREAKING) O3-5577: Refactor Patient Search App removing patient search context

EDSONZ-WASSWA:
## Requirements

## Summary

The patient search app was juggling three host environments (the desktop top-nav search, the tablet search overlay, and embeddable workspaces launched from Ward, Service Queues, and Appointments) by passing callbacks through two parallel React contexts: \`PatientSearchContext\` (legacy) and \`PatientSearchContext2\` (Workspace v2). Components like \`PatientBanner\` had to introspect which context was alive and branch accordingly, and the \`/search\` page was wrapping the tree in an empty \`PatientSearchContextProvider\` just to keep \`useContext()\` from throwing.

This PR deletes both contexts and passes the same callbacks (\`onPatientSelected\`, \`patientClickSideEffect\`, plus the workspace handles where applicable) as explicit props. Components no longer need to ask "which context am I in?", the call site decides.

### What changed

- Removed \`patient-search-context.ts\` along with both providers and \`usePatientSearchContext\` / \`usePatientSearchContext2\`.
- Introduced a single shared \`PatientSearchCallbackProps\` interface in \`types/index.ts\` that all banner / search / workspace components extend.
- \`PatientBanner\`, \`CompactPatientBanner\`, \`PatientSearch\`, \`RecentlySearchedPatients\`, \`AdvancedPatientSearchComponent\`, and \`PatientSearchLg\` now receive callbacks as props rather than reading them from context.
- \`PatientSearchOverlay\` now contains the search bar + advanced search directly instead of mounting the v1 \`PatientSearchWorkspace\`.
- Removed the empty \`<PatientSearchContextProvider value={{}}>\` wrapper from \`PatientSearchPageComponent\`.
- Tests were updated to drop the context wrappers; missing \`voided: false\` was added to fixtures; new unit tests were added for \`compact-patient-search.extension\`, \`patient-search-overlay\`, and \`patient-search-lg\`.

### What's removed (extensions / workspaces)

- \`patientSearchButton\` extension and the \`patient-search-button-slot\` it populated.
- \`patientSearchWorkspace\` (v1) workspace and its \`patient-search-workspace\` registration.
- Translation key \`searchPatientButton\`.
- The v1 fallback that rendered \`start-visit-button-slot\` from \`PatientBanner\`. Only \`start-visit-button-slot2\` is rendered now.

### Breaking changes

- Consumers of the v1 workspace must migrate to \`patient-search-workspace2\`, launched via \`launchWorkspace('patient-search-workspace2', { ... })\` or \`launchChildWorkspace(...)\`.
- Apps that previously included the \`patientSearchButton\` extension (or rendered \`patient-search-button-slot\`) should launch the workspace directly from their own button.
- \`startVisitPatientSearchActionButton\` in \`openmrs-esm-patient-chart\` (registered to \`start-visit-button-slot\` at [esm-patient-chart-app/src/routes.json](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/routes.json)) will silently stop rendering after this lands, since nothing renders that slot anymore. A follow-up PR in \`openmrs-esm-patient-chart\` should either move that extension to \`start-visit-button-slot2\` or drop it.

## Related Issue

[O3-5577](https://openmrs.atlassian.net/browse/O3-5577)

## Other

cc @chibongho @denniskigen @ibacher

[O3-5577]: https://openmrs.atlassian.net/browse/O3-5577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ELVIS-KATO (dismissed):

packages/esm-patient-search-app/src/patient-search-page/patient-banner/banner/patient-banner.component.tsx
```diff
@@ -130,58 +141,36 @@ const PatientBanner: React.FC<PatientBannerProps> = ({
...
-        onClick={handleClick}>
-        {children}
-      </button>
-    );
   } else {
     return (
       <ConfigurableLink
         className={styles.patientBanner}
-        onBeforeNavigate={handleBeforeNavigate}
+        onBeforeNavigate={() => patientClickSideEffect?.(patient.id, patient)}
```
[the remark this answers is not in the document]
chibongho: @ELVIS-KATO  `onPatientSelected` is not used here?

chibongho (commented):
Thanks @EDSONZ-WASSWA. Looks better than before.

- As this is a critical feature, can you list out what you did to test your changes, especially to components these files? (If you think existing tests cover your changes, say so.)
  - compact-patient-search.extension.tsx
  - patient-search-overlay.component.tsx
  - patient-search-lg.component.tsx
- Not your fault, but the naming of some of these existing components are quite similar, which makes it hard for me to review. While you're at it, can you add JSDoc comments to the above files (and others) to just briefly document what the component does and where it's used?
- The PR title / description check Github Actions failed. Please fix those.
- There are a couple failing E2E tests. I don't think it's related to this change, but can you investigate to make sure?

chibongho:
> Thanks @EDSONZ-WASSWA. Looks better than before.
> 
>     * As this is a critical feature, can you list out what you did to test your changes, especially to components these files? (If you think existing tests cover your changes, say so.)
>       
>       * compact-patient-search.extension.tsx
>       * patient-search-overlay.component.tsx
>       * patient-search-lg.component.tsx
> 
>     * Not your fault, but the naming of some of these existing components are quite similar, which makes it hard for me to review. While you're at it, can you add JSDoc comments to the above files (and others) to just briefly document what the component does and where it's used?
> 
>     * The PR title / description check Github Actions failed. Please fix those.
> 
>     * There are a couple failing E2E tests. I don't think it's related to this change, but can you investigate to make sure?

@EDSONZ-WASSWA Can you make sure these are all done?

EDSONZ-WASSWA:
> > Thanks @EDSONZ-WASSWA. Looks better than before.
> > ```
> > * As this is a critical feature, can you list out what you did to test your changes, especially to components these files? (If you think existing tests cover your changes, say so.)
> >   
> >   * compact-patient-search.extension.tsx
> >   * patient-search-overlay.component.tsx
> >   * patient-search-lg.component.tsx
> > 
> > * Not your fault, but the naming of some of these existing components are quite similar, which makes it hard for me to review. While you're at it, can you add JSDoc comments to the above files (and others) to just briefly document what the component does and where it's used?
> > 
> > * The PR title / description check Github Actions failed. Please fix those.
> > 
> > * There are a couple failing E2E tests. I don't think it's related to this change, but can you investigate to make sure?
> > ```
> 
> @EDSONZ-WASSWA Can you make sure these are all done?

@chibongho for compact-patient-search.extension.tsx i switched to tablet viewport, searched for a patient, clicked a result and it went to the chart. Checked that arrow keys still worked for navigating results and that the search bar cleared after picking someone. The tests in patient-search.test.tsx it already cover the loading, empty and error states so those helped too.

Now for patient-search-overlay.component.tsx i tested it two ways. Went directly to /openmrs/spa/search?query= on a tablet and the overlay came up with the query already filled in, and the back arrow took me to the home page. Also clicked the search icon in the tablet top nav, did a search and selected a patient then navigated to the chart as expected. No unit tests exist for this one since it really just wraps PatientSearchWorkspace and passes props through, so manual testing was the main way to verify it.

Then for patient-search-lg.component.tsx i tested in both desktop and tablet contexts since it uses the inTabletOrOverlay prop to adjust its layout, ran searches in both, pagination worked, selecting a patient navigated correctly in both cases. advanced-patient-search.test.tsx covers the parent component behavior so that gave some extra coverage. The component mostly just passes props down so there was not much that could go wrong here specifically.
I have fixed  the PR title / description check Github Action
The e2e are not related to my changes

packages/esm-patient-search-app/src/patient-search-context.ts
```diff
@@ -1,43 +0,0 @@
-import { createContext, useContext } from 'react';
-import { type Workspace2DefinitionProps } from '@openmrs/esm-framework';
-
-/**
- * @deprecated This context should be removed once the workspace v2 migration is completed,
- * as the callbacks in this context are incompatible with the new API
- */
-export interface PatientSearchContextProps {
-  /**
```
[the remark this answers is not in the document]
chibongho: These comments are still missing from `ClickablePatientContainerProps`.

packages/esm-patient-search-app/src/compact-patient-search/patient-search.component.tsx
```diff
@@ -6,13 +6,33 @@ import { type PatientSearchResponse } from '../types';
 import CompactPatientBanner from './compact-patient-banner.component';
 import Loader from './loader.component';
 import styles from './patient-search.scss';
-
+/**
+ * PatientSearchComponent is the core logic and UI for the compact search bar.
+ *
+ * Used by the CompactPatientSearchExtension in the top nav, it handles input state,
```
chibongho:
```suggestion
 * Used by the CompactPatientSearchExtension in the top nav in desktop mode, it handles input state,
```

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
```diff
@@ -10,42 +10,63 @@ import {
...
 
 interface CompactPatientBannerProps {
   patients: Array<SearchedPatient>;
+  nonNavigationSelectPatientAction?: (patientUuid: string, patient: fhir.Patient) => void;
+  patientClickSideEffect?: (patientUuid: string, patient: fhir.Patient) => void;
 }
+/**
+ * CompactPatientBanner displays a minimized summary of the currently active patient.
+ *
+ * It is used within the top navigation bar once a patient context is established,
```
chibongho:
```suggestion
 * It is used within the top navigation bar in desktop mode to display search results,
```

packages/esm-patient-search-app/src/patient-search-button/patient-search-button.component.tsx
chibongho, line 1: We should delete this file (and corresponding test file, export in `index.ts` and extension entry in `routes.json`). This should no longer be used. Apps that want to include search functionality should make their own button to launch the patient search workspace.

packages/esm-patient-search-app/src/patient-search-overlay/patient-search-overlay.component.tsx
```diff
@@ -18,8 +22,8 @@ interface PatientSearchOverlayProps extends PatientSearchContextProps {
  * - in the top nav, when the user clicks on the magnifying glass icon
  *   (in desktop mode, the inline CompactPatientSearchComponent is used instead)
  *
- * Although similar looking, this overlay behaves somewhat differently from a regular
- * workspace, and has its own overlay logic.
+ * This overlay provides a container for the patient search functionality
```
chibongho, line 27: Can we keep the old comment as well?

chibongho (changes requested):
I also found a regression, when I go to the queue app and try to search for a patient to add to a queue, clicking on a patient in the search result causing the workspace to close. Can you fix it?

EDSONZ-WASSWA:
@chibongho  I have addressed the review, and  removed the hardcoded patientClickSideEffect={() => closeWorkspace()} that was overriding the consuming app's own callbacks

chibongho:
Thanks @EDSONZ-WASSWA, looks good. Marking this is a breaking change since we are remove the search button extension. 

@denniskigen @ibacher , can you review this as well as this is a critical feature?

Testing done by me:
- used the patient search in the queues app to create a new queue entry
- searched for a patient in the top nav in desktop mode
- searched for a patient with the magnifying glass icon in top nav in tablet mode.

chibongho (approved):

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
```diff
@@ -10,42 +10,73 @@ import {
...
+ * CompactPatientBanner displays a minimized summary of the currently active patient.
+ *
+ * It is used within the top navigation bar in desktop mode to display search results,
+ * allowing quick visibility of the patient's name and ID without taking up much space.
+ */
+const CompactPatientBanner = forwardRef<HTMLDivElement, CompactPatientBannerProps>(
+  ({ patients, nonNavigationSelectPatientAction, patientClickSideEffect }, ref) => {
+    const fhirMappedPatients: Array<fhir.Patient> = useMemo(() => {
+      return patients.map(mapToFhirPatient);
+    }, [patients]);
```
ibacher:
```suggestion
    const fhirMappedPatients: Array<fhir.Patient> = useMemo(() => patients.map(mapToFhirPatient), [patients]);
```

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
```diff
@@ -10,42 +10,73 @@ import {
...
+          </ClickablePatientContainer>
+        );
+      },
+      [nonNavigationSelectPatientAction, patientClickSideEffect],
     );
-  }, []);
 
-  return <div ref={ref}>{fhirMappedPatients.map(renderPatient)}</div>;
-});
+    return <div ref={ref}>{fhirMappedPatients.map(renderPatient)}</div>;
```
ibacher: Why don't we just do this in the `useMemo()` above?

packages/esm-patient-search-app/src/compact-patient-search/recently-searched-patients.component.tsx
```diff
@@ -6,9 +6,18 @@ import type { PatientSearchResponse } from '../types';
 import CompactPatientBanner from './compact-patient-banner.component';
 import Loader from './loader.component';
 import styles from './patient-search.scss';
+/**
```
ibacher, line 12:
```suggestion

/**
```

packages/esm-patient-search-app/src/compact-patient-search/recently-searched-patients.component.tsx
```diff
@@ -6,9 +6,18 @@ import type { PatientSearchResponse } from '../types';
...
+ * RecentlySearchedPatients displays a list of patients the user has interacted with recently.
+ *
+ * It is rendered below the search bar (before a query is typed) to provide quick
+ * access to recent contexts without needing to perform a full search.
+ */
```
ibacher, line 17: It's weird for this doc string to be attached to the `Props` interface instead of the component itself. Please move it.

packages/esm-patient-search-app/src/compact-patient-search/recently-searched-patients.component.tsx
```diff
@@ -6,9 +6,18 @@ import type { PatientSearchResponse } from '../types';
...
+interface RecentlySearchedPatientsProps extends PatientSearchResponse {
+  patientClickSideEffect?: (patientUuid: string, patient: fhir.Patient) => void;
+}
```
ibacher: Why doesn't this support the `nonNavigationSelectPatientAction` as well?

packages/esm-patient-search-app/src/patient-search-overlay/patient-search-overlay.component.tsx
```diff
@@ -1,25 +1,32 @@
...
 
-interface PatientSearchOverlayProps extends PatientSearchContextProps {
+export interface PatientSearchOverlayProps {
   onClose: () => void;
-  handleSearchTermUpdated?: (value: string) => void;
   query?: string;
   header?: string;
+  handleSearchTermUpdated?: (query: string) => void;
+  nonNavigationSelectPatientAction?: (patientUuid: string, patient: fhir.Patient) => void;
+  patientClickSideEffect?: (patientUuid: string, patient: fhir.Patient) => void;
```
ibacher: The fact that we keep redefining these properties maybe indicates that we should have a shared interface everything can extend from.

packages/esm-patient-search-app/src/compact-patient-search.extension.tsx
```diff
@@ -4,10 +4,15 @@ import { useTranslation } from 'react-i18next';
 import { useConfig, navigate, interpolateString } from '@openmrs/esm-framework';
 import { type PatientSearchConfig } from './config-schema';
 import { useInfinitePatientSearch } from './patient-search.resource';
-import { PatientSearchContextProvider } from './patient-search-context';
 import useArrowNavigation from './hooks/useArrowNavigation';
 import PatientSearch from './compact-patient-search/patient-search.component';
 import styles from './compact-patient-search.scss';
+/**
```
ibacher, line 18: This is a detached doc string that should be moved to the appropriate component.

packages/esm-patient-search-app/src/compact-patient-search.extension.tsx
ibacher: If we're already doing this as a breaking PR, it seems like this is a good time to standardize the naming conventions.

I would suggest that instead of `nonNavigationSelectPatientAction` we standardize on `onPatientSelected()`. Here especially it feels like we should get the property name to be `nonNavigationSelectPatientAction()` if we keep that naming convention.

packages/esm-patient-search-app/src/patient-search-page/patient-banner/banner/patient-banner.component.tsx
ibacher: This import is no longer used I think.

packages/esm-patient-search-app/src/compact-patient-search.extension.tsx
ibacher:
```suggestion
  selectPatientAction?: (patientUuid: string) => void;
```

packages/esm-patient-search-app/src/patient-search-overlay/patient-search-overlay.component.tsx
EDSONZ-WASSWA: Hello @ibacher  I have Created a shared PatientSearchCallbackProps interface in types/index.ts that all components now extend instead of redefining the same props individually
Then for the onPatientSelected signature in PatientSearchCallbackProps uses ...args: any[] to accommodate both the compact extension use case (2 arguments) and the workspace v2 use case (4 arguments) in a single shared type. Happy to discuss if there is a cleaner approach you would prefer. I did this as a walk around of some errors i faced.

packages/esm-patient-search-app/src/types/index.ts
```diff
@@ -162,3 +162,18 @@ export interface SearchFieldConfig {
...
   max?: number;
 }
+
+export interface PatientSearchCallbackProps {
+  /**
+   * A function to execute instead of navigating the user to the patient
+   * dashboard. If null/undefined, patient results will be links to the
+   * patient dashboard.
+   */
+  onPatientSelected?: (patientUuid: string, patient: fhir.Patient, ...args: any[]) => void;
```
denniskigen: The unified `...args: any[]` makes the callback accept anything in the trailing arguments, which is a step back from the previous `PatientSearchContext2Props` typing that precisely declared `launchChildWorkspace` and `closeWorkspace` (using `Workspace2DefinitionProps['launchChildWorkspace']`). With the rest-args form, TS won't catch a caller that forgets the workspace plumbing or passes them in the wrong order. It's also one of 11 `no-explicit-any` lint errors blocking CI on this branch.

Worth considering: either keep the precise 4-arg signature (`launchChildWorkspace?`, `closeWorkspace?` as optional), or split into two narrow types (`PatientSearchCallbackProps` for the simple case, `WorkspacePatientSearchCallbackProps extends` it for the v2 workspace path) and have the page-banner accept the wider one. Either keeps the type contract honest at the call sites in `patient-banner.component.tsx:148` and `compact-patient-banner.component.tsx`.

packages/esm-patient-search-app/src/compact-patient-search.extension.tsx
```diff
@@ -43,8 +47,8 @@ const CompactPatientSearchComponent: React.FC<CompactPatientSearchProps> = ({
   const handlePatientSelection = useCallback(
     (event, index: number) => {
       event.preventDefault();
-      if (selectPatientAction) {
-        selectPatientAction(patients[index].uuid);
+      if (onPatientSelected) {
+        onPatientSelected(patients[index].uuid, null);
```
denniskigen: Passing `null` as the patient argument here surprised me. Every other call site passes the actual patient object. This is the keyboard-arrow-navigation path, and `patients[index]` is in scope, so could we pass the mapped patient (or at least `patients[index]`) instead? Otherwise any `onPatientSelected` consumer that uses the second arg has to special-case the compact extension.

packages/esm-patient-search-app/src/patient-search-page/patient-banner/banner/patient-banner.component.tsx
```diff
@@ -130,58 +126,36 @@ const PatientBanner: React.FC<PatientBannerProps> = ({
...
-    nonNavigationSelectPatientAction(patientUuid, patient);
-    patientClickSideEffect?.(patientUuid, patient);
-  }, [nonNavigationSelectPatientAction, patientClickSideEffect, patientUuid, patient]);
-
-  const handleBeforeNavigate = useCallback(() => {
-    patientClickSideEffect?.(patientUuid, patient);
-  }, [patientClickSideEffect, patientUuid, patient]);
-
-  if (context2) {
+  if (onPatientSelected) {
```
denniskigen, line 140: This file took the most logic change (collapsing the v1/v2 context branches into a single explicit-props path), but I don't see a unit test for it in the diff. The compact banner has tests, but this one carries the more interesting branching (`onPatientSelected` button vs. `ConfigurableLink`, including the `patientClickSideEffect` ordering on each branch). Could a minimal test be added to cover those two paths? Happy to be wrong if it's covered transitively somewhere.

packages/esm-patient-search-app/src/routes.json
```diff
@@ -34,15 +28,5 @@
       "slot": "start-visit-button-slot2",
```
denniskigen, line 28: One thing worth calling out under the (BREAKING) umbrella so downstream distros aren't surprised: `start-visit-button-slot` (v1) is no longer rendered anywhere after this PR. `patient-banner.component.tsx` only renders `start-visit-button-slot2` now. That orphans `startVisitPatientSearchActionButton` in `openmrs-esm-patient-chart` (registered into `start-visit-button-slot` at [`esm-patient-chart-app/src/routes.json:124`](https://github.com/openmrs/openmrs-esm-patient-chart/blob/main/packages/esm-patient-chart-app/src/routes.json#L124)). Either a follow-up PR in patient-chart should re-register against slot2, or anyone with custom extensions in that slot needs a heads-up that they'll silently stop rendering. (I see this is now mentioned in the PR body, thanks.)

packages/esm-patient-search-app/src/compact-patient-search.extension.tsx
```diff
@@ -4,20 +4,24 @@ import { useTranslation } from 'react-i18next';
...
-  /** An action to take when the patient is selected, other than navigation. If not provided, navigation takes place. */
-  selectPatientAction?: (patientUuid: string) => undefined;
   buttonProps?: object;
 }
 
+/**
+ * CompactPatientSearchExtension renders the inline search bar used in the top navigation menu.
+ *
+ * It is primarily used in desktop views. In tablet/mobile views, the search bar is hidden
+ * behind a magnifying glass icon (PatientSearchButton) that opens the PatientSearchOverlay instead.
```
denniskigen: Tiny one: the doc comment was added in this PR, but `PatientSearchButton` is also deleted in this same PR. The actual tablet/mobile flow goes through `PatientSearchIcon` (in `patient-search-icon/patient-search-icon.component.tsx`). Could you swap the name so the doc is correct on day one?

denniskigen (commented):
Thanks for the cleanup @EDSONZ-WASSWA, the context-free version is much easier to follow. A few inline comments below, plus a couple of items I'm tracking separately on the runtime regression and the lint failures.

EDSONZ-WASSWA:
@denniskigen  sorry for the delay we are doing exams this way, I have worked on the issues you talked about.
 mainly,  The ...args: any[] was thrown out entirely and replaced with two explicitly named optional parameters. launchChildWorkspace is typed as a function that takes a workspace name and optional props and returns a Promise<void>. closeWorkspace is typed as a function that returns Promise<boolean >

PR #2475, opened 2026-04-07
(fix) Add null safety and fallback handling in useAppointmentsCalendar hook

Safa-khanum:
## Requirements

- [x] This PR has a title that briefly describes the work done

## Summary

This PR improves the stability of the appointments calendar hook by handling potential runtime errors caused by undefined or unexpected data.

Changes include:
- Added null safety for API response (`data?.data ?? []`) to prevent crashes during reduce operations
- Safely handled `appointmentCountMap` using a fallback to avoid errors when undefined
- Added optional chaining and default values for `allAppointmentsCount`
- Added a default fallback return in `evaluateAppointmentCalendarDates` to prevent undefined return values

These changes ensure the hook behaves safely even when API data is missing or invalid.

## Screenshots

Not applicable (no UI changes)

## Related Issue

N/A

## Other

This is a small stability improvement focused on preventing edge-case crashes in the appointments calendar logic.

### DIFF
--- a/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
+++ b/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
@@ -27,7 +27,7 @@ export const useAppointmentsCalendar = (forDate: string, period: string) => {
   );
   // Transform API response into daily appointment counts grouped by service
 
-  const results: Array<DailyAppointmentsCountByService> = (data?.data ?? []).reduce((acc, service) => {
+   const results = (data?.data ?? []).reduce<DailyAppointmentsCountByService[]>((acc, service) => {
     const serviceName = service.appointmentService.name;
     const serviceUuid = service.appointmentService.uuid;
 
--- a/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
+++ b/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
@@ -26,9 +26,8 @@ export const useAppointmentsCalendar = (forDate: string, period: string) => {
     { errorRetryCount: 2 },
   );
   // Transform API response into daily appointment counts grouped by service
-  const safeData = data?.data ?? [];
 
-  const results: Array<DailyAppointmentsCountByService> = safeData.reduce((acc, service) => {
+  const results: Array<DailyAppointmentsCountByService> = (data?.data ?? []).reduce((acc, service) => {
     const serviceName = service.appointmentService.name;
     const serviceUuid = service.appointmentService.uuid;
 
--- a/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
+++ b/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
@@ -61,6 +61,8 @@ export const useAppointmentsCalendar = (forDate: string, period: string) => {
 
 function evaluateAppointmentCalendarDates(forDate: string, period: string) {
   if (period === 'daily') {
+   
+    console.warn(`Unhandled period value: ${period}, defaulting to daily`);
     return {
       startDate: dayjs(forDate).startOf('day').format(omrsDateFormat),
       endDate: dayjs(forDate).endOf('day').format(omrsDateFormat),
--- a/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
+++ b/packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
@@ -81,4 +81,10 @@ function evaluateAppointmentCalendarDates(forDate: string, period: string) {
       endDate: dayjs(forDate).endOf('month').format(omrsDateFormat),
     };
   }
+
+  // Default fallback (THIS is the new part)
+  return {
+    startDate: dayjs(forDate).startOf('day').format(omrsDateFormat),
+    endDate: dayjs(forDate).endOf('day').format(omrsDateFormat),
+  };
 }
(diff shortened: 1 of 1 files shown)


packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
```diff
@@ -66,4 +81,10 @@ function evaluateAppointmentCalendarDates(forDate: string, period: string) {
       endDate: dayjs(forDate).endOf('month').format(omrsDateFormat),
     };
   }
+
+  // Default fallback (THIS is the new part)
+  return {
+    startDate: dayjs(forDate).startOf('day').format(omrsDateFormat),
+    endDate: dayjs(forDate).endOf('day').format(omrsDateFormat),
+  };
```
omsrivastava512: The fallback behaviour itself is questionable because the function previously returned `undefined` for unhandled period strings, which would surface a bug visibly. 

This change makes it silently return a single-day range for **any unknown period string**, masking bad input rather than surfacing it.

 Not rejecting it outright, but a `console.warn` for the unhandled period would at least preserve debuggability without sacrificing the fallback.

omsrivastava512 (changes requested):
Good direction on the null safety fixes, the core changes here are right. Left a few inline comments on some specific points worth looking into before this merges. Nothing major, just some things that need a second pass.

omsrivastava512 (changes requested):

ELVIS-KATO:
Hello @Safa-khanum , could you please create a ticket on Jira and link the PR, this is better for follow up and one to know what the PR is actually all about. otherwise thanks for the work done.

Safa-khanum:
> Hello @Safa-khanum , could you please create a ticket on Jira and link the PR, this is better for follow up and one to know what the PR is actually all about. otherwise thanks for the work done.

Got it, I’ll create a Jira ticket and link the PR for proper follow-up. Thanks for pointing that out!

packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
```diff
@@ -26,27 +26,43 @@ export const useAppointmentsCalendar = (forDate: string, period: string) => {
...
   );
   // Transform API response into daily appointment counts grouped by service
-  const results: Array<DailyAppointmentsCountByService> = data?.data.reduce((acc, service) => {
+
+   const results = (data?.data ?? []).reduce<DailyAppointmentsCountByService[]>((acc, service) => {
     const serviceName = service.appointmentService.name;
     const serviceUuid = service.appointmentService.uuid;
-    Object.entries(service.appointmentCountMap).forEach(([key, value]) => {
+
+    Object.entries(service.appointmentCountMap ?? {}).forEach(([key, value]) => {
```
jayasanka-sack, line 34: From the backend, appointmentCountMap is always initialized and never returned as null. ([here](https://github.com/Bahmni/openmrs-module-appointments/blob/a150aa6dea858d1d8c8cfe7890a0387718f8cc7c/omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java#L120)) So we can safely rely on it being present and remove the fallback here. 

Silencing it in this case might hide unexpected issues if the response shape ever changes, instead of surfacing them early.

Also when opening PRs like this, it’s worth double-checking the backend data structures first. Otherwise these “maybe null” guards can creep in and add unnecessary noise to the code.

packages/esm-appointments-app/src/hooks/useAppointmentsCalendar.ts
```diff
@@ -26,27 +26,43 @@ export const useAppointmentsCalendar = (forDate: string, period: string) => {
...
       const existingEntry = acc.find((entry) => entry.appointmentDate === key);
+
       if (existingEntry) {
-        existingEntry.services.push({ serviceName, serviceUuid, count: value.allAppointmentsCount });
+        existingEntry.services.push({
+          serviceName,
+          serviceUuid,
+          count: value?.allAppointmentsCount ?? 0,
```
jayasanka-sack, line 41: This basically translates to:
"maybe value is null, maybe allAppointmentsCount is missing… not sure… let’s default to 0"

Check this:
https://github.com/Bahmni/openmrs-module-appointments/blob/a150aa6dea858d1d8c8cfe7890a0387718f8cc7c/omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java#L125

`value` is always a `DailyAppointmentServiceSummary` instance created in the same flow, and `allAppointmentsCount` is derived from `appointments.size()`. Since appointments is always initialized, this will consistently be a number (including 0), not null or undefined.

jayasanka-sack:
Thanks for the PR @Safa-khanum!

Looking through the changes, it seems most of the null/undefined guards were added without fully checking how the backend structures the response.

I’m onboard with data?.data ?? [] since that aligns with SWR’s loading state. The other fallbacks aren’t necessary and end up adding noise, and may even hide unexpected issues.

Given that, I’m closing this PR for now. Feel free to reopen after aligning with the backend data structures.
Thanks again! ❤️

Safa-khanum:
> Thanks for the PR @Safa-khanum!
> 
> Looking through the changes, it seems most of the null/undefined guards were added without fully checking how the backend structures the response.
> 
> I’m onboard with data?.data ?? [] since that aligns with SWR’s loading state. The other fallbacks aren’t necessary and end up adding noise, and may even hide unexpected issues.
> 
> Given that, I’m closing this PR for now. Feel free to reopen after aligning with the backend data structures. Thanks again! ❤️

Thanks for the detailed feedback!
I understand the concern regarding the null/undefined guards and how they may not fully align with the backend response structure. I’ll take a step back and review the API responses more thoroughly to ensure the frontend logic reflects the actual data shape.

I agree with your point about relying on data?.data in line with SWR’s loading state and avoiding unnecessary fallbacks that could introduce noise or mask issues.
I’ll rework the changes accordingly and make sure everything is aligned with the backend before reopening the PR.

Thanks again for the guidance! 🙏

PR #2491, opened 2026-04-13
(fix) O3-3734: Hide non-selectable adjacent month dates in DOB date picker

bhavya-jpg:
## Requirements

## Summary
This PR fixes a confusing UX issue in the patient registration Date of Birth date picker where dates outside the selected month appeared as greyed-out but unselectable elements.

Root cause:
The OpenmrsDatePicker migration from Flatpickr to react-aria-components introduced a behavior where "padding" dates (outside the current month) do not have click handlers. These dates were still rendered visually, creating a misleading impression that they were selectable.

Fix:
- Hides adjacent month padding dates using a CSS override targeting the `[data-outside-month]` attribute
- Ensures all visible dates are valid and selectable within the current month
- Removes misleading visual cues and improves clarity of interaction

## Screenshots
Screenshot 2026-04-13 at 10 38 15 PM
Screenshot 2026-04-13 at 10 38 26 PM

## Related Issue
https://openmrs.atlassian.net/browse/O3-3734

## Other
- Targets `[data-outside-month]`, a semantic attribute from react-aria-components, making it robust against class changes  
- CSS is scoped to `patient-registration.scss` to prevent side effects  
- No JavaScript changes required (pure CSS fix)


bhavya-jpg:
https://github.com/user-attachments/assets/3d3c9dab-a1fc-429f-af3d-cad7a70e2830

bhavya-jpg:
hi @VeronicaMuthee ,can I please get a review on this

VeronicaMuthee:
@bhavya-jpg, I don't know if hiding non-selectable adjacent month in the date-picker is the most durable fix but i would like others to weigh in ~ @ibacher @denniskigen @NethmiRodrigo

EDSONZ-WASSWA:
@VeronicaMuthee  how about we use Wrap React-Aria DatePicker with custom logic to filter out adjacent days

ibacher:
@bhavya-jpg Thanks for the contribution! I'm closing this because this approach does not seem responsive to what the ticket is suggesting or what we should do. The ticket asks that the dates not be greyed out, I think with the idea that these are valid dates to select. Instead, the implementation here hides those dates, which is not the behaviour we want. The dates should remain greyed out (consistent with Carbon's designs), but should be selectable (which is actually discussed on the ticket). Please feel free to submit a new PR responsive to this.

PR #2503, opened 2026-04-21
(feat) O3-5620: Show inpatient note edit history

chibongho:
## Requirements

## Summary

This PR adds a new feature to view the inpatient notes edit history.

## Screenshots

[Screencast from 2026-04-21 10-46-45.webm](https://github.com/user-attachments/assets/4c6dbecd-85a0-4512-87d5-93e722f34004)

## Related Issue

https://issues.openmrs.org/browse/O3-5620

### DIFF
--- a/e2e/specs/ward-patient-notes.spec.ts
+++ b/e2e/specs/ward-patient-notes.spec.ts
@@ -131,7 +131,7 @@ test('Add a patient note to an inpatient admission', async ({ page, api }) => {
 
   await test.step('Then I should see the modal showing the edit history with 2 versions', async () => {
     await expect(page.getByText('Note edit history')).toBeVisible();
-    // await expect(page.getByText('Sample patient note - edited')).toBeVisible();
+    await expect(page.getByRole('dialog').getByText('Sample patient note - edited')).toBeVisible();
     await expect(page.getByText('Sample patient note', { exact: true })).toBeVisible();
   });
 });
--- a/e2e/specs/ward-patient-notes.spec.ts
+++ b/e2e/specs/ward-patient-notes.spec.ts
@@ -54,11 +54,11 @@ test('Add a patient note to an inpatient admission', async ({ page, api }) => {
   });
 
   await test.step('And I select the bed for admission', async () => {
-    await page.getByText(`${bed.bedNumber} · Empty`).click();
+    await wardPage.selectBedForAdmission(bed.bedNumber);
   });
 
   await test.step('And I confirm admission by clicking "Admit"', async () => {
-    await page.getByRole('button', { name: 'Admit', exact: true }).click();
+    await wardPage.confirmAdmission();
   });
 
   await test.step('Then I should see a success message confirming the admission success', async () => {
@@ -91,6 +91,48 @@ test('Add a patient note to an inpatient admission', async ({ page, api }) => {
 
   await test.step('Then I should see a success message confirming the note was saved', async () => {
     await expect(page.getByText(/patient note saved/i)).toBeVisible();
+  });
+
+  await test.step('And when I click on the patient card to open the patient workspace', async () => {
+    await wardPage.clickPatientCard(fullName);
+  });
+
+  await test.step('Then I click the "Patient note" button in the siderail to open the patient note workspace again', async () => {
+    await wardPage.clickPatientNotesButton();
+  });
+
+  await test.step('Then I should see the note there', async () => {
+    await expect(page.getByText('Sample patient note')).toBeVisible();
+  });
+
+  await test.step('And when I click on the note overflow menu and select "Edit"', async () => {
+    await wardPage.selectEditNoteOption();
+  });
+
+  await test.step('Then I edit the note to "Sample patient note - edited"', async () => {
+    await wardPage.fillEditNote('Sample patient note - edited');
+  });
+
+  await test.step('And I save the edited note', async () => {
+    await wardPage.clickSaveEditButton();
+  });
+
+  await test.step('Then I should see a success message confirming the note was saved', async () => {
+    await expect(page.getByText(/patient note saved/i)).toBeVisible();
+  });
+
+  await test.step('Then I should see the note with a "Last edited by" line', async () => {
+    await expect(page.getByText(/last edited by/i)).toBeVisible();
+  });
+
+  await test.step('And when I click on the note overflow menu and select "View edit history"', async () => {
+    await wardPage.selectViewEditHistoryOption();
+  });
+
+  await test.step('Then I should see the modal showing the edit history with 2 versions', async () => {
+    await expect(page.getByText('Note edit history')).toBeVisible();
+    // await expect(page.getByText('Sample patient note - edited')).toBeVisible();
+    await expect(page.getByText('Sample patient note', { exact: true })).toBeVisible();
   });
 });
 
--- a/packages/esm-ward-app/src/routes.json
+++ b/packages/esm-ward-app/src/routes.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
-    "webservices.rest": ">=2.2.0",
+    "webservices.rest": ">3.3.0 <4.0.0",
     "emrapi": ">=2.0.0 <4.0.0"
   },
   "optionalBackendDependencies": {
@@ -70,6 +70,10 @@
     {
       "name": "delete-note-modal",
       "component": "deleteNoteModal"
+    },
+    {
+      "name": "note-history-modal",
+      "component": "noteHistoryModal"
     }
   ],
   "workspaces2": [
--- a/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.scss
+++ b/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.scss
@@ -1,5 +1,3 @@
-@use '@carbon/layout';
-
 .versionTile {
   background-color: #fff;
 }
--- a/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.scss
+++ b/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.scss
@@ -1,0 +1,5 @@
+@use '@carbon/layout';
+
+.versionTile {
+  background-color: #fff;
+}
--- a/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.tsx
@@ -1,0 +1,46 @@
+import React from 'react';
+import dayjs from 'dayjs';
+import { useTranslation } from 'react-i18next';
+import { ModalHeader, ModalBody } from '@carbon/react';
+import { type PatientNote } from '../types';
+import styles from './styles.scss';
+import modalStyles from './note-history.modal.scss';
+
+interface NoteHistoryModalProps {
+  close: () => void;
+  note: PatientNote;
+}
+
+const NoteHistoryModal: React.FC<NoteHistoryModalProps> = ({ close, note }) => {
+  const { t } = useTranslation();
+
+  const formatDate = (datetime: string) => dayjs(datetime).format('D MMM YYYY, HH:mm');
+
+  const versions = [
+    { note: note.encounterNote, recordedAt: note.lastEditedAt, recordedBy: note.lastEditedBy },
+    ...note.editHistory.map((v) => ({ note: v.note, recordedAt: v.recordedAt, recordedBy: v.recordedBy })),
+  ];
+
+  return (
+    <>
+      <ModalHeader closeModal={close}>{t('noteEditHistory', 'Note edit history')}</ModalHeader>
+      <ModalBody>
+        <div className={styles.notesContainer}>
+          {versions.map((v, i) => (
+            <div key={i} className={`${styles.noteTile} ${modalStyles.versionTile}`}>
+              <div className={styles.noteBody}>{v.note}</div>
+              <div className={styles.noteProviderName}>
+                {t('writtenBy', 'Written by: {{name}}, {{date}}', {
+                  name: v.recordedBy,
+                  date: formatDate(v.recordedAt),
+                })}
+              </div>
+            </div>
+          ))}
+        </div>
+      </ModalBody>
+    </>
+  );
+};
+
+export default NoteHistoryModal;
--- a/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.tsx
@@ -5,6 +5,7 @@ import { ModalHeader, ModalBody } from '@carbon/react';
 import { type PatientNote } from '../types';
 import styles from './styles.scss';
 import modalStyles from './note-history.modal.scss';
+import classNames from 'classnames';
 
 interface NoteHistoryModalProps {
   close: () => void;
@@ -27,7 +28,7 @@ const NoteHistoryModal: React.FC<NoteHistoryModalProps> = ({ close, note }) => {
       <ModalBody>
         <div className={styles.notesContainer}>
           {versions.map((v, i) => (
-            <div key={i} className={`${styles.noteTile} ${modalStyles.versionTile}`}>
+            <div key={i} className={classNames(styles.noteTile, modalStyles.versionTile)}>
               <div className={styles.noteBody}>{v.note}</div>
               <div className={styles.noteProviderName}>
                 {t('writtenBy', 'Written by: {{name}}, {{date}}', {
--- a/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note.component.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note.component.tsx
@@ -108,18 +108,20 @@ const InPatientNote: React.FC<InPatientNoteProps> = ({ note, mutatePatientNotes,
                   }}
                 />
               )}
-              <OverflowMenuItem
-                aria-label={t('viewEditHistory', 'View edit history')}
-                id={'view-history-' + note.encounterUuid}
-                className={styles.menuItem}
-                itemText={t('viewEditHistory', 'View edit history')}
-                onClick={() => {
-                  const dispose = showModal('note-history-modal', {
-                    close: () => dispose(),
-                    note,
-                  });
-                }}
-              />
+              {note.isEdited && (
+                <OverflowMenuItem
+                  aria-label={t('viewEditHistory', 'View edit history')}
+                  id={'view-history-' + note.encounterUuid}
+                  className={styles.menuItem}
+                  itemText={t('viewEditHistory', 'View edit history')}
+                  onClick={() => {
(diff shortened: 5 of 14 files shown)


packages/esm-ward-app/src/routes.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://json.openmrs.org/routes.schema.json",
   "backendDependencies": {
-    "webservices.rest": ">=2.2.0",
+    "webservices.rest": ">3.3.0 <4.0.0",
```
mogoodrich, line 4: This seems kind of weird (that it doesn't support 4.0.0 yet?)

Also if this requires RESTWS-1026, looks like that change won't be available until 3.4.0?

mogoodrich (commented):
Generally looks good, but had some questions about the dependencies... does this require 3.4.0-SNAPSHOT of REST WS and do we need to do a release?

packages/esm-ward-app/src/routes.json
chibongho: > This seems kind of weird (that it doesn't support 4.0.0 yet?)

We don't do [it](https://github.com/openmrs/openmrs-esm-patient-management/blob/53717bc8cd002bca0ef6bedb4aca7c2a5e1b4403/packages/esm-service-queues-app/src/routes.json) everywhere, but this is a safeguard against future breaking changes. Following semver convention, a major version bump corresponds to a breaking change. When we bump the REST module version to 4.0.0 in the future, we'll get a backend dependency warning in the ward app, which should prompt us to manually verify that it still works fine before we relax the upper bound to `<5.0.0`.

> Also if this requires RESTWS-1026, looks like that change won't be available until 3.4.0?

Either `3.4.0` or `3.4.0-SNAPSHOT`. I already made the corresponding version bump [in the refapp](https://github.com/openmrs/openmrs-distro-referenceapplication/pull/1005/changes). I'll do the same for PIH EMR.

mogoodrich (approved):

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.resource.ts
```diff
@@ -45,10 +45,20 @@ export function usePatientNotes(patientUuid: string, visitUuid: string, conceptU
...
                     conceptUuid: obs.concept.uuid,
                     encounterTypeUuid: encounter.encounterType.uuid,
+                    isEdited: (obs.previousVersions?.length ?? 0) > 0,
+                    lastEditedBy: obs.creator?.person?.display ?? obs.creator?.display ?? '',
+                    lastEditedAt: obs.obsDatetime,
+                    editHistory: (obs.previousVersions ?? []).map((v) => ({
+                      note: v.value as string,
+                      recordedAt: v.obsDatetime,
+                      recordedBy: v.creator?.person?.display ?? v.creator?.display ?? '',
+                    })),
```
denniskigen, line 59: Hmm, I think the edit timestamps here might not do what we want. `obsDatetime` is the clinical observation time, and when an obs is edited OpenMRS preserves it unchanged across versions: `Obs.newInstance()` copies it, and `saveExistingObs` nulls `creator`/`dateCreated` (so they get reset to the editor + now) but leaves `obsDatetime` alone.

So:
- `lastEditedAt: obs.obsDatetime` will show the original clinical time, not when the edit actually happened
- every `editHistory[i].recordedAt: v.obsDatetime` is identical across versions, so the modal timeline ends up with the same timestamp repeated

I think we'd want `dateCreated` for the edit time instead. Could you extend the custom rep to something like `obs:(...,dateCreated,previousVersions:(value,obsDatetime,dateCreated,creator),creator)` and use `dateCreated` for `lastEditedAt` and each `editHistory[i].recordedAt`?

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/types.ts
```diff
@@ -38,4 +46,20 @@ export interface ObsData {
...
+
+export interface Provider {
+  uuid: string;
+  display: string;
+  comments: string;
+  response?: string;
+  person: OpenmrsResource;
+  location: string;
+  serviceType: string;
 }
```
denniskigen, line 56: Small nit on the type name: `obs.creator` in OpenMRS is actually a `User`, not a `Provider`. The fields declared here (`comments`, `response`, `location`, `serviceType`) aren't on the User default rep that comes back. Since we only ever read `display` and `person.display`, could this just be inline, `creator?: { display: string; person?: { display: string } }`, to match `ObsData` above?

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.workspace.tsx
```diff
@@ -65,6 +65,7 @@ const WardPatientNotesWorkspace: React.FC<WardPatientWorkspaceDefinition> = ({
       const notePayload: EncounterPayload = {
         patient: patientUuid,
         location: locationUuid,
+        visit: wardPatient.visit?.uuid,
```
denniskigen: Nice catch that notes weren't being associated with a visit. Would you mind splitting this into its own commit (or its own PR)? It's a separate bug from edit history and would be easier to bisect/revert later if it stands alone.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note.component.tsx
```diff
@@ -115,6 +108,19 @@ const InPatientNote: React.FC<InPatientNoteProps> = ({ note, mutatePatientNotes,
                   }}
                 />
               )}
+              {note.isEdited && (
+                <OverflowMenuItem
```
denniskigen, line 112: Tiny consistency thing: the sibling `OverflowMenuItem`s for edit and delete both set `aria-label`. Could you add one here too? Something like `aria-label={t('viewEditHistory', 'View edit history')}`.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/notes-container.test.tsx
```diff
@@ -85,8 +113,29 @@ describe('PatientNotesHistory', () => {
...
+    });
+
+    mockedUsePatientNotes.mockReturnValue({
+      patientNotes: [mockEditedPatientNote],
+      isLoadingPatientNotes: false,
+      errorFetchingPatientNotes: undefined,
+      mutatePatientNotes: jest.fn(),
+    });
+
+    render(<PatientNotesHistory patientUuid={mockPatientUuid} promptBeforeClosing={jest.fn()} visitUuid={''} />);
```
denniskigen: Thanks for the "Last edited by" test. Would you be open to also adding coverage for:
- the "View edit history" menu item only appearing when `isEdited` is true
- the modal rendering the full version list when opened

Not a blocker, just would lock in the new behavior.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note.component.tsx
```diff
@@ -115,6 +108,19 @@ const InPatientNote: React.FC<InPatientNoteProps> = ({ note, mutatePatientNotes,
...
+                  id={'view-history-' + note.encounterUuid}
+                  className={styles.menuItem}
```
denniskigen, line 115: Noticed while looking at the overflow menu: the three `OverflowMenuItem`s all reference `className={styles.menuItem}`, but `.menuItem` isn't actually defined in `history/styles.scss`, so Carbon's default `max-width` applies and longer labels like "View edit history" can get truncated.

The convention across the rest of the monorepo (appointments, service queues, patient list management) is:

```scss
.menuItem {
  max-width: none;
}
```

Pre-existing gap, but worth fixing here since this PR introduces the first label that's noticeably longer than "Edit"/"Delete".

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/types.ts
mogoodrich: Oh, thanks, nice catch @denniskigen !

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.resource.ts
mogoodrich: And nice catch here as well, thanks @denniskigen , agree that this should be date created.

mogoodrich (changes requested):
Thanks for taking a closer look than I did @denniskigen !  Yes, @chibongho  let's fix the stuff up regarding dates and user/provider, thanks!

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.workspace.tsx
chibongho: Filed PR: https://github.com/openmrs/openmrs-esm-patient-management/pull/2506

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.resource.ts
chibongho: Changed them all to use `dateCreated` instead of obsDatetime.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/notes-container.test.tsx
chibongho: extended an e2e test to include editing an inpatient note.

chibongho:
There is one failing E2E test, I'm still investigating, but I think it's because the refapp image it's running against does not have the latest REST module (3.4.0-SNAPSHOT) for some reason. That said, I think the PR is ready for review again. Thanks.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/types.ts
```diff
@@ -38,4 +46,21 @@ export interface ObsData {
...
+}
+
+export interface Provider {
+  uuid: string;
+  display: string;
+  comments: string;
+  response?: string;
+  person: OpenmrsResource;
+  location: string;
+  serviceType: string;
```
mogoodrich: I know there was some talk of this being a User, not a Provider, but I'm not seeing where this is used?

mogoodrich (approved):
One small question but LGTM!

e2e/specs/ward-patient-notes.spec.ts
```diff
@@ -92,6 +92,48 @@ test('Add a patient note to an inpatient admission', async ({ page, api }) => {
...
+    await expect(page.getByText(/last edited by/i)).toBeVisible();
+  });
+
+  await test.step('And when I click on the note overflow menu and select "View edit history"', async () => {
+    await wardPage.selectViewEditHistoryOption();
+  });
+
+  await test.step('Then I should see the modal showing the edit history with 2 versions', async () => {
+    await expect(page.getByText('Note edit history')).toBeVisible();
+    // await expect(page.getByText('Sample patient note - edited')).toBeVisible();
```
denniskigen: This looks like a leftover commented-out assertion. OK to remove?

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.scss
```diff
@@ -0,0 +1,5 @@
+@use '@carbon/layout';
```
denniskigen:
```suggestion
```

Doesn't look like we're using this import. Safe to remove I think.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/types.ts
denniskigen: Yeah, we can remove this interface.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note-history.modal.tsx
```diff
@@ -0,0 +1,46 @@
...
+    ...note.editHistory.map((v) => ({ note: v.note, recordedAt: v.recordedAt, recordedBy: v.recordedBy })),
+  ];
+
+  return (
+    <>
+      <ModalHeader closeModal={close}>{t('noteEditHistory', 'Note edit history')}</ModalHeader>
+      <ModalBody>
+        <div className={styles.notesContainer}>
+          {versions.map((v, i) => (
+            <div key={i} className={`${styles.noteTile} ${modalStyles.versionTile}`}>
```
denniskigen: nit: This could be simplified using `classNames`:


```suggestion
            <div key={i} className={classNames(styles.noteTile, modalStyles.versionTile)}>
```

e2e/specs/ward-patient-notes.spec.ts
chibongho: Sorry, meant to change it to a different selector. Will fix.

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/styles.scss
```diff
@@ -72,3 +72,13 @@
...
 }
+
+.noteEditedBy {
+  font-size: 0.75rem;
+  margin-top: layout.$spacing-02;
+  color: $text-02;
+}
+
+.menuItem {
+  max-width: none;
```
denniskigen, line 84: nit: Noticed Delete's red hover background doesn't fill the full menu width on my end. Carbon sets `max-inline-size: 11.25rem` (the logical property, not `max-width`) on `.cds--overflow-menu-options__btn`, so the existing override doesn't always cascade over it. Adding the logical equivalent fixes the Delete hover gap:

```suggestion
.menuItem {
  max-width: none;
  max-inline-size: none;
}
```
 
CleanShot 2026-04-23 at 21 17 14@2x

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/history/note.component.tsx
```diff
@@ -115,6 +108,18 @@ const InPatientNote: React.FC<InPatientNoteProps> = ({ note, mutatePatientNotes,
...
+                id={'view-history-' + note.encounterUuid}
+                className={styles.menuItem}
+                itemText={t('viewEditHistory', 'View edit history')}
+                onClick={() => {
+                  const dispose = showModal('note-history-modal', {
+                    close: () => dispose(),
+                    note,
+                  });
+                }}
+              />
```
denniskigen: Small UX polish: this `OverflowMenuItem` renders unconditionally, so it shows on unedited notes too. Clicking it there opens the "Note edit history" modal with just a single tile of the current note, which feels a bit redundant.

Would you consider only showing it when there's actual history?

```suggestion
              {note.isEdited && (
                <OverflowMenuItem
                  aria-label={t('viewEditHistory', 'View edit history')}
                  id={'view-history-' + note.encounterUuid}
                  className={styles.menuItem}
                  itemText={t('viewEditHistory', 'View edit history')}
                  onClick={() => {
                    const dispose = showModal('note-history-modal', {
                      close: () => dispose(),
                      note,
                    });
                  }}
                />
              )}
```

packages/esm-ward-app/src/ward-workspace/ward-patient-notes/notes.resource.ts
ibacher: Hmmm... this is problematic. We actually want `obsDateTime` to represent the time the obs is valid. This is one of the ways we support RDE (because with `created` where recording when the obs came into existence even if its back-dated). The right fix here is to ensure we're always updated the `obsDatetime` when we update the note.

chibongho: Just to be clear, this mean that when we do `POST /obs/<some-obs-uuid>` to "update" an obs, we should create a new obs row that is a clone of `some-obs-uuid`, but with its `obsDatetime` set to current time instead of `some-obs-uuid.obsDatetime`?

mogoodrich: Hmmm... not sure I agree @ibacher . In HTML Form entry, if you edit an obs, the new obs that gets created gets the obs_datetime of initial obs (which is usually set to the encounter datetime when it was first created), it's only the date_created that is timestamped to the current time.

For this widget, which is primarily (only?) designed for real-time data entry, I think a typical use case would be, say, a clinician enters a note at 4:02pm that says "Patient has a hedache" (sic).  If the clinician comes back at 4:30pm to fix their typo, then the I would still expect the obs date to be 4:02pm... as that is still the clinically-relevant time for this obs.

That being said--and this may be what Ian is getting at--if the clinician came back and *completely edited* the note at 4:30pm based on new information the patient reports them then, then, yes, theoretically it should get a new obs datetime.  But I would suggest that an improper use of this feature... but then again we don't usually prevent that.

fyi @mseaton

chibongho:
Merging this in to get it into an upcoming release. Post-review suggestions welcome. (Sorry...)

jayasanka-sack:
I was going through why the E2E job is failing, and figured the following

1. Patient management wasn't getting deployed to the dev3 for a while, probably after reverting to the old github workflows here: https://github.com/openmrs/openmrs-esm-patient-management/commit/8eea19a7a9862a19aa6600c4e81406565faa799c. The bambo job is no longer in use, so the frontend won't get deployed. I'm not yet sure why we reverted.
2. webservices.rest module also hasn't get deployed for a while. The build keep failing, so the deployment is skipped. https://github.com/openmrs/openmrs-module-webservices.rest/actions/runs/24605774382

I'll try to get this fixed.

jayasanka-sack:
Alright, 

Regarding the point 1, we need to switch back to the common workflow, so it should fix the deployment issue.

point 2: 
I checked [jfrog](https://openmrs.jfrog.io/ui/packages/gav:%2F%2Forg.openmrs.module:webservices.rest-omod/3.4.0-20260416.204254-3?name=webservices.rest-omod&type=packages), the latest snapshot of rest module was updated on `17-04-26 02:14:48 +0530`, which is [@chibongho 's commit](https://github.com/openmrs/openmrs-module-webservices.rest/commit/083ac60a71ed68e2df47146bccee657c1ac8070d) on the backend change. 

However, the `Deploy SNAPSHOT to maven` job is failing since `14-04-26 03:20:00 +0530` ([job](https://github.com/openmrs/openmrs-module-webservices.rest/actions/runs/24368714416)). So having an updated snapshot in the jfrog is impossible, unless someone manually pushed it. Also, the dev3 backend contains these changes. I'm puzzled. Any idea on what's going on here @ibacher @wikumChamith ?    

Anyway moving to the main question, "why E2E jobs are failing?"

The reason is, E2E jobs are using a prefilled docker container to save time with demo data generation, We used [this bamboo plan](https://ci.openmrs.org/browse/O3-BP) to get those images populated, however, we no longer trigger the job. So the image we use for E2E is outdated by 2 months. That's why it doesn't contain the latest backend changes. I triggered a manual build for now, but we should create a github action to populate the image.

ibacher:
> Any idea on what's going on here @ibacher @wikumChamith ?

There are two branches in the REST module. `master` which is 4.x and `3.x` [which is where the SNAPSHOT came from](https://github.com/openmrs/openmrs-module-webservices.rest/actions/runs/24532856183/job/71720768291).

jayasanka-sack:
Ah... That explains @ibacher. The E2Es are working fine now as I manually updated the prefilled image, I'll create a github workflow soon.

chibongho:
@jayasanka-sack Thanks for investigating. I have a (untested) PR out for also building the nightly-with-data image in the github action: https://github.com/openmrs/openmrs-distro-referenceapplication/pull/1007

PR #2504, opened 2026-04-21
(feat) O3-5624: Add ward app configuration to only show discharge button in certain locations

chibongho:
## Requirements

## Summary

This PR add a way to configure the discharge button in the ward app workspace to only render at certain locations.

Manually tested by configuring my local instance to only show the button at a certain location (mother's dorm in PIH distribution) and verify it shows / hides the button based on the location.

## Related Issue

https://issues.openmrs.org/browse/O3-5624

### DIFF
--- a/packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail-config-schema.ts
+++ b/packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail-config-schema.ts
@@ -1,7 +1,7 @@
 import { Type } from '@openmrs/esm-framework';
 
 export const dischargeWorkspaceSiderailExtensionConfigSchema = {
-  onlyShowDischargeForLocations: {
+  allowedSessionLocationUuids: {
     _type: Type.Array,
     _default: [],
     _description:
@@ -13,5 +13,5 @@ export const dischargeWorkspaceSiderailExtensionConfigSchema = {
 };
 
 export interface DischargeWorkspaceSiderailConfig {
-  onlyShowDischargeForLocations: string[];
+  allowedSessionLocationUuids: string[];
 }
--- a/packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail-config-schema.ts
+++ b/packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail-config-schema.ts
@@ -1,0 +1,17 @@
+import { Type } from '@openmrs/esm-framework';
+
+export const dischargeWorkspaceSiderailExtensionConfigSchema = {
+  onlyShowDischargeForLocations: {
+    _type: Type.Array,
+    _default: [],
+    _description:
+      'If not empty, the discharge button will only show for users whose session location is included in this list. If empty, the discharge button will show for all locations.',
+    _elements: {
+      _type: Type.String,
+    },
+  },
+};
+
+export interface DischargeWorkspaceSiderailConfig {
+  onlyShowDischargeForLocations: string[];
+}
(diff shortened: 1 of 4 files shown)


mseaton (approved):

packages/esm-ward-app/src/index.ts
```diff
@@ -137,6 +139,7 @@ export const deleteNoteModal = getAsyncLifecycle(
 export function startupApp() {
   registerBreadcrumbs([]);
   defineConfigSchema(moduleName, configSchema);
+  defineExtensionConfigSchema('ward-patient-discharge', dischargeWorkspaceSiderailExtensionConfigSchema);
```
denniskigen, line 142: I think the name here should be `patient-discharge-siderail-button` (the actual extension name at [routes.json:49](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-ward-app/src/routes.json#L49)). `ward-patient-discharge` is a workspaceWindow name ([routes.json:195](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-ward-app/src/routes.json#L195)), so the schema gets registered under a key nothing is looking up.

Knock-on effect: the `_default: []` never gets applied, so `onlyShowDischargeForLocations` comes back `undefined` at runtime and the `.length` check in the component throws in any deployment that doesn't set this config explicitly. The manual test worked because you had the array set. The empty-default path just hasn't been exercised.

Swapping the name should sort it.

packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail-config-schema.ts
```diff
@@ -0,0 +1,17 @@
+import { Type } from '@openmrs/esm-framework';
+
+export const dischargeWorkspaceSiderailExtensionConfigSchema = {
+  onlyShowDischargeForLocations: {
```
denniskigen: Two small snags with `onlyShowDischargeForLocations`:

- `Locations` is ambiguous — could read as session, patient, ward, or encounter location.
- `only ... []` reads "only show for none" but means "show everywhere" — a gotcha for someone editing config at 2am.

Maybe something like `allowedSessionLocationUuids` reads better. The `Uuids` suffix also matches what the code compares against (`sessionLocation?.uuid`) and preempts "can I use a display name here?"

packages/esm-ward-app/src/action-menu-buttons/discharge-workspace-siderail.component.tsx
```diff
@@ -1,17 +1,25 @@
...
+    return (
+      <ActionMenuButton2
+        icon={(props) => <Exit {...props} />}
+        label={t('discharge', 'Discharge')}
+        workspaceToLaunch={{
+          workspaceName: 'patient-discharge-workspace',
+        }}
+      />
+    );
+  } else {
```
denniskigen: nit: This reads more naturally with early returns - no `else` branch needed.

denniskigen (changes requested):
Thanks for picking this up — per-location gating is useful for multi-ward deployments. I've left some feedback on a couple of things.

packages/esm-ward-app/src/index.ts
chibongho: That's not true. With the new workspace system, a window with an icon gets treated has an extension, with extension name set to the window's name. (See [here](https://github.com/openmrs/openmrs-esm-core/pull/1544/changes#diff-ac439765c15fec0fdec7e5820d2c07c0bef6263c7a1bd3ffee9e0facf886a47bR204-R206)). The intent is that we should not need to re-export the icon as an extension in order to configure it. I also verified by changing it to `patient-discharge-siderail-button` and seeing it throw an exception when I click on a patient card in the ward when the config is not set. 

The extension entry in [routes.json:49](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-ward-app/src/routes.json#L49) is old code from the old workspace system. I'll remove it as part of this PR.

chibongho:
Merging this in, as we are making a release soon. Post-commit feedback welcome.

PR #2515, opened 2026-04-28
(fix) Make the create appointments form translatable

ibacher:
## Requirements

## Summary

The main goal of this PR is to ensure that appointment types and appointment statuses are translated in the drop-downs on the appointment scheduling form (they are already translated on the table).

As part of this, I added validators to ensure that the config schema values are using the enums, added the translation keys alongside the constants for the configuration and slightly tweaked the `<SelectItem />` which had some unnecessary mark-up.


packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
```diff
@@ -835,7 +835,7 @@ const AppointmentsForm: React.FC<Workspace2DefinitionProps<AppointmentsFormProps
                     onBlur={onBlur}
                     ref={ref}
                     value={value}>
-                    <SelectItem text={t('chooseProvider', 'Choose a provider')} value="" />
+                    <SelectItem key="chooseProvider" text={t('chooseProvider', 'Choose a provider')} value="" />
                     {providers?.providers?.length > 0 &&
                       providers?.providers?.map((provider) => (
                         <SelectItem key={provider.uuid} text={provider.display} value={provider.uuid}>
```
denniskigen:
```suggestion
                        <SelectItem key={provider.uuid} text={provider.display} value={provider.uuid} />
```

Should we fix this up too?

denniskigen (commented):
LGTM. Thanks, @ibacher.

Tangential, but this diff adds `Requested` to the `AppointmentStatus` enum, but `canTransition` in `src/helpers/functions.ts` still has no sequence entry for it. That means requested appointments selected in batch status changes can be blocked by the frontend even though the backend would accept forward transitions. This is not strictly introduced here, since `Requested` was already in the prior config default, but this PR is the natural place to close the gap now that the enum formalizes the value.

denniskigen (approved):

packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
ibacher: Yeah... I must've missed this one.

PR #1369, opened 2024-11-12
(refactor) Refactor bed management app to match conventions

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [ ] My work includes tests or is validated by existing tests.

## Summary

This PR applies a set of changes to the bed management app to match the conventions of the other apps in the O3 ecosystem.

PR #2538, opened 2026-05-19
(fix) O3-5700: Prevent state updates after component unmount in useBedsGroupedByLocation

omeriinnocent:
## Requirements

## Summary
This PR fixes a React state update issue in useBedsGroupedByLocation by guarding async state updates with an isSubscribed check.

The fix prevents setResult, setError, setIsLoading, and setIsValidating from running after the component has unmounted while asynchronous bed fetch requests are still in progress.

This improves stability during rapid navigation between bed and patient management views by preventing stale updates, React warnings, and unintended UI behavior.
## Related Issue
[O3-5700](https://openmrs.atlassian.net/browse/O3-5700)

## Other

[O3-5700]: https://openmrs.atlassian.net/browse/O3-5700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


omeriinnocent:
Hi @jayasanka-sack @denniskigen , may i have a little of your time to review this

omeriinnocent:
@ELVIS-KATO, may you help review this

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -109,21 +111,18 @@ export function useBedsGroupedByLocation() {
...
+          setError(error);
         })
         .finally(() => {
-          if (isSubscribed) {
-            setIsLoading(false);
-            setIsValidating(false);
-          }
+          if (!isSubscribed) return;
+          setIsLoading(false);
+          setIsValidating(false);
```
denniskigen, line 125: These guards already exist on `main`; they came in with [#1369](https://github.com/openmrs/openmrs-esm-patient-management/pull/1369). All four setters are currently wrapped in `if (isSubscribed)`, so changing those checks into early returns preserves the same behavior. I ran the same unmount-mid-flight test against both implementations and got identical results.

We’re also on React 18.3.1, and [React 18 removed the unmounted-component warning](https://github.com/facebook/react/pull/22114) described in O3-5700. There is no warning for this change to prevent.

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -95,10 +95,12 @@ export function useBedsGroupedByLocation() {
     if (!isLoadingAdmissionLocations && admissionLocations && isValidating) {
       const fetchData = async () => {
         const promises = admissionLocations.map(async (location) => {
+          if (!isSubscribed) return null;
```
denniskigen, line 98: This check cannot fire. `admissionLocations.map()` invokes each async callback synchronously until its first `await`, and this effect’s cleanup cannot run until the effect body returns. Every `openmrsFetch` call has therefore already started while `isSubscribed` is still `true`.

The check after the `await` is reachable, but it only discards data that would otherwise reach `setResult`, which is already guarded. If the goal is to stop these requests after navigation, that would require cancelling them rather than adding another subscription check.

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -109,21 +111,18 @@ export function useBedsGroupedByLocation() {
...
+        if (!isSubscribed) return;
+        setResult(updatedWards);
       };
       fetchData()
         .catch((error) => {
-          if (isSubscribed) {
-            setError(error);
-          }
+          if (!isSubscribed) return;
+          setError(error);
```
denniskigen, line 120: There is a separate reproducible issue here: once a revalidation fails, `error` is set and never cleared. I reproduced this by letting the initial bed request succeed and then injecting a failure into the next `GET /bed?locationUuid=...` revalidation.

The existing rows are then replaced by `<ErrorState />`, and the error branch also removes the row actions and Add bed button, leaving no retry path in the UI. Even a later successful revalidation does not restore the table because the old error remains set.

If you’d like to retarget this PR, can we create a new ticket and update the hook so a background failure keeps the existing rows visible and a successful retry clears the error? It’s probably better to consider moving this aggregation onto SWR, since the other hooks in this file already use it, but the regression behavior is the important part.

denniskigen (changes requested):
Thanks for taking a crack at this, @omeriinnocent, and nice work following through on the regression coverage in [#2450](https://github.com/openmrs/openmrs-esm-patient-management/pull/2450). I couldn't reproduce O3-5700 on `main`, and the existing guards, React 18 behavior, and matching unmount tests show that this diff has no observable effect, so I don't think we can merge it as-is; I've requested changes and left the details inline, including a separate reproducible bug in the same hook that we could track under a new ticket and retarget this PR toward.

packages/esm-bed-management-app/src/summary/summary.resource.ts
omeriinnocent: Thanks @denniskigen , let me look into that

PR #2539, opened 2026-05-19
(feat) O3-5699: Make choosing a provider when creating an appointment optional.

ahmihassu:
## Requirements

## Summary
I have added a configuration called requireProvider that can be true or false. The default value is true(checked) so that the current way of creating an appointment won't be affected. If a user changes the value to false(unchecked) then the section that let's the user choose a provider doesn't appear in the UI and a user can create an appointment without choosing a provider. 
## Screenshots
[Screencast from 2026-05-19 15-09-08.webm](https://github.com/user-attachments/assets/bd80b37f-c57a-462e-b83a-a34bcc825ad9)


packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
```diff
@@ -228,6 +237,12 @@ const AppointmentsForm: React.FC<Workspace2DefinitionProps<AppointmentsFormProps
     ? new Date(appointment?.dateAppointmentScheduled)
     : new Date();
 
+  const defaultProviderUuid = requireProvider
+    ? appointment?.providers?.find((p) => p.response === 'ACCEPTED')?.uuid ??
+      session?.currentProvider?.uuid ??
+      ''
+    : (appointment?.providers?.find((p) => p.response === 'ACCEPTED')?.uuid ?? '');
```
chibongho, line 244: can we just default it to '' when `requireProvider` is false?

packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
```diff
@@ -269,6 +281,13 @@ const AppointmentsForm: React.FC<Workspace2DefinitionProps<AppointmentsFormProps
 
   useEffect(() => setValue('formIsRecurringAppointment', isRecurringAppointment), [isRecurringAppointment, setValue]);
 
+  useEffect(() => {
```
chibongho, line 284: I don't think this useEffect is needed.

chibongho (commented):
Thanks. Looks good, just a few minor comments.

ahmihassu:
I will address the comments and get back to you shortly. Thank you.

packages/esm-appointments-app/src/form/appointments-form.workspace.tsx
chibongho: Never mind, I think your way is correct. In edit mode, we should have it be populated by the existing appointment's providers.

PR #2540, opened 2026-05-20
(fix) O3-5665: Fix vitals form and triage form not opening

UjjawalPrabhat:
## Requirements

## Summary

The **Vitals form** and **Triage form** buttons in the Service Queues "Current visit" panel were broken — they redirected users to the patient summary page instead of opening the form. Clinic staff had to manually navigate to find and open the form, slowing down patient intake.

Now both buttons open the form directly in a side panel, so vitals and triage notes can be recorded without leaving the queue.

## Screenshots

https://github.com/user-attachments/assets/6f1d1745-98a8-49a4-a8b2-28de7dc5e474

## Related Issue

[O3-5665](https://openmrs.atlassian.net/browse/O3-5665)

## Other

[O3-5665]: https://openmrs.atlassian.net/browse/O3-5665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


UjjawalPrabhat:
@ibacher @NethmiRodrigo @denniskigen

jwnasambu (commented):
@UjjawalPrabhat Thanks for the progress on this project. Kindly ensure the pull request title includes the correct Jira ticket ID as outlined in the linked guidelines. https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips.

packages/esm-service-queues-app/src/current-visit/visit-details/triage-note.component.tsx
```diff
@@ -42,8 +44,20 @@ const TriageNote: React.FC<TriageNoteProps> = ({ notes, patientUuid, diagnoses }
...
+                'service-queues-patient-clinical-forms-workspace',
+                {},
+                {
+                  formEntryWorkspaceName: 'service-queues-patient-form-entry-workspace',
+                  patient,
+                  patientUuid,
+                  visitContext: visit,
+                },
+              )
+            }
```
NethmiRodrigo: If you look at [what gets rendered](https://github.com/openmrs/openmrs-esm-patient-management/blob/232ead1930911059c7def70e38a899a182895411/packages/esm-service-queues-app/src/current-visit/visit-details/current-visit-details.component.tsx#L25) in the Triage note section, you’d see that it appears to diagnoses, and the visit note. So I doubt that launching the clinical form workspace is the right step to do here. The visit note workspace seems more appropriate.

packages/esm-service-queues-app/src/current-visit/visit-details/vitals.component.tsx
```diff
@@ -144,8 +146,15 @@ const Vitals: React.FC<VitalsComponentProps> = ({ vitals, patientUuid, visitType
...
+                disabled={!patient}
                 renderIcon={(props) => <ArrowRight size={16} {...props} />}
-                onClick={() => navigate({ to: `\${openmrsSpaBase}/patient/${patientUuid}/chart` })}
+                onClick={() =>
+                  launchWorkspace2('service-queues-patient-vitals-workspace', {
+                    patient,
+                    patientUuid,
+                    visitContext: visit,
+                  })
+                }
```
NethmiRodrigo: Based on your screen recording, a user would now need to refresh the page to see the updated vitals. It needs to be so that when it gets submitted the swr cache here gets invalidated and the vitals get updated.

packages/esm-service-queues-app/translations/en.json
```diff
@@ -216,6 +215,7 @@
   "urgent": "Urgent",
   "use12HourFormat": "Please use 12-hour format with AM/PM (01–12)",
   "useMidnightFormat": "For midnight use 12:00 AM, for noon use 12:00 PM",
+  "visitNoteForm": "Visit note form",
```
jwnasambu, line 218:
```suggestion
  "triageForm": "Triage form"
```

packages/esm-service-queues-app/src/current-visit/visit-details/visit-note.component.tsx
```diff
@@ -42,10 +44,17 @@ const TriageNote: React.FC<TriageNoteProps> = ({ notes, patientUuid, diagnoses }
...
-            {t('triageForm', 'Triage form')}
+            onClick={() =>
+              launchWorkspace2('service-queues-visit-notes-workspace', { formContext: 'creating' }, null, {
+                patient,
+                patientUuid,
+                visitContext: visit,
+              })
+            }
+            iconDescription={t('visitNoteForm', 'Visit note form')}>
+            {t('visitNoteForm', 'Visit note form')}
```
jwnasambu, line 56:
```suggestion
            iconDescription={t('triageForm', 'Triage form')}>
             {t('triageForm', 'Triage form')}
```
    Kindly the specifications requires the buttons to be labeled "Vitals form →" and "Triage form →".
     Reverting from "Visit note form" ensures the feature meets the business requirements.

packages/esm-service-queues-app/src/routes.json
```diff
@@ -196,6 +206,14 @@
...
       "name": "add-queue-room",
       "group": "service-queues"
+    },
+    {
+      "name": "service-queues-patient-vitals",
+      "group": "service-queues"
+    },
+    {
+      "name": "service-queues-visit-note",
+      "group": "service-queues"
```
jwnasambu:
```suggestion
     { "name": "service-queues-visit-note" }
```
I want to believe O3, the framework injects the group's context into the workspace because the service-queues group does not have a patient context, it injects null into the workspace's groupProps overriding any manual data you send via launchWorkspace2. By removing the group association prevents this
     override.

packages/esm-service-queues-app/src/current-visit/visit-details/triage-note.component.tsx
```diff
@@ -42,10 +44,17 @@ const TriageNote: React.FC<TriageNoteProps> = ({ notes, patientUuid, diagnoses }
           <Button
             size="sm"
             kind="ghost"
+            disabled={!patient}
             renderIcon={(props) => <ArrowRight size={16} {...props} />}
-            onClick={() => navigate({ to: `\${openmrsSpaBase}/patient/${patientUuid}/chart` })}
-            iconDescription={t('triageForm', 'Triage form')}>
-            {t('triageForm', 'Triage form')}
+            onClick={() =>
+              launchWorkspace2('service-queues-visit-notes-workspace', { formContext: 'creating' }, null, {
```
jwnasambu: To ensure stability for launching workspaces, I recommend updating this launchWorkspace2 call to prevent a potential runtime exception.

```suggestion
       ```
       launchWorkspace2('service-queues-visit-notes-workspace',
      { formContext: 'creating' }
     {},
    { patient, patientUuid } 
     > );
```
```
Kindly this the reason for the above changes:
 1. The esm-patient-notes-app workspace component is designed to destructure
  patient data from the 4th argument (groupProps). Passing this data in the 2nd argument instead
  will cause the component to receive null for its context resulting in a TypeError: Cannot read
  properties of null (reading 'patientUuid').

2. Using an empty object {} instead of null for the 3rd argument
  (windowProps) is the preferred pattern to ensure the framework correctly maps the subsequent
  shared context (the 4th argument).

 3. Since the Notes app automatically resolves the patient’s active visit
  using the patientUuid, providing visitContext is redundant. Removing it helps keep the prop
  interface clean and aligned with what the component actually uses."

cc @NethmiRodrigo , @chibongho , @denniskigen, @ibacher Kindly feel free to correct me.

File: packages/esm-service-queues-app/src/routes.json (hunk 5 above)
Comment (UjjawalPrabhat) [reply packages/esm-service-queues-app/src/routes.json]: Thanks @jwnasambu. I looked into this against the framework code and I don't think we can drop the group here. `launchWorkspace2` requires every window to belong to a group - `getGroupByWindowName` returns `undefined` and it throws `Workspace window ... is not registered to a workspace group`, so removing the group would break the launch entirely. On the override concern: group props passed to `launchWorkspace2` actually win (`props: groupProps ?? openedGroup?.props ?? null`), and `arePropsCompatible` treats any nullish side as compatible, so the `service-queues` group doesn't clobber the `{ patient, patientUuid }` we pass. I've confirmed the notes workspace receives them correctly.

File: packages/esm-service-queues-app/src/current-visit/visit-details/triage-note.component.tsx (hunk 6 above)
Comment (UjjawalPrabhat) [reply packages/esm-service-queues-app/src/current-visit/visit-details/triage-note.component.tsx]: Good catch on `visitContext` — `visitNotesFormWorkspace` never reads it, removed. `patient`/`patientUuid` are already in the 4th arg (groupProps), which is where the component destructures them. `{}` vs `null` for windowProps doesn't cause the TypeError (both compatible), but happy to use `{}`.

File: packages/esm-service-queues-app/src/current-visit/visit-details/visit-note.component.tsx (hunk 4 above)
Comment (UjjawalPrabhat) [reply packages/esm-service-queues-app/src/current-visit/visit-details/visit-note.component.tsx:56]: The button now opens the Visit Note form (per @NethmiRodrigo's suggestion [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2540#discussion_r3299135123)), so "Visit note form" reflects what actually launches. So, shouldn't we keep the label and the workspace it opens stay consistent?

cc @NethmiRodrigo

Comment (jwnasambu) [reply packages/esm-service-queues-app/src/current-visit/visit-details/visit-note.component.tsx:56]: Thanks @UjjawalPrabhat I had not got a chance to lot at the prior discussion. Thanks for the point out.

File: packages/esm-service-queues-app/src/current-visit/visit-details/current-visit-details.component.tsx [hunk 7]
```diff
@@ -28,33 +31,29 @@ const CurrentVisitDetails: React.FC<CurrentVisitProps> = ({ patientUuid, encount
 
     // Iterating through every Encounter
     encounters?.forEach((enc: Encounter) => {
-      // Check for Visit Diagnoses and Notes
-      if (enc.encounterType?.display === 'Visit Note') {
```
NethmiRodrigo, line 32: I don't understand the reason for removing this condition?

packages/esm-service-queues-app/src/current-visit/visit-details/current-visit-details.component.tsx
```diff
@@ -28,33 +31,29 @@ const CurrentVisitDetails: React.FC<CurrentVisitProps> = ({ patientUuid, encount
...
-              provider: {
-                name: enc.encounterProviders.length ? enc.encounterProviders[0].provider.person.display : '',
-                role: enc.encounterProviders.length ? enc.encounterProviders[0].encounterRole.display : '',
-              },
-              time: formatTime(parseDate(obs.obsDatetime)),
-              concept: obs.concept,
-            });
+      enc.obs?.forEach((obs: Observation) => {
+        if (obs.concept?.uuid === concepts.visitDiagnosesConceptUuid) {
+          const problemList = obs.groupMembers?.find((mem) => mem.concept.display === 'PROBLEM LIST');
```
NethmiRodrigo: Shouldn’t this too use the config concept like `config.concepts.problemListConceptUuid`?

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
```diff
@@ -1,18 +1,53 @@
...
 import { Tag, DataTableSkeleton } from '@carbon/react';
+import { getGlobalStore } from '@openmrs/esm-framework';
 import { useVisit } from './current-visit.resource';
 import CurrentVisitDetails from './visit-details/current-visit-details.component';
 import styles from './current-visit.scss';
 
+const VISIT_REFRESHING_WORKSPACES = new Set([
+  'service-queues-visit-notes-workspace',
+  'service-queues-patient-vitals-workspace',
+]);
```
NethmiRodrigo: If either workspace `name` in routes.json (service-queues-visit-notes-workspace / service-queues-patient-vitals-workspace) is later renamed but this Set is not updated, the close-detection silently never matches, mutate() never fires, and the visit summary stops refreshing after the form is submitted.

packages/esm-service-queues-app/src/routes.json
NethmiRodrigo: We should probably give these two a separate workspace group since group props are shared by every window in the group. The visit-note form passes {patient, patientUuid} as group props, so patient-specific context ends up in a group shared with the queue-management workspaces, which don't use a patient.

This works today because the queue management workspaces launch with null group props, which is compatible with anything. But when launching into a group already open with incompatible non-null group props, launchWorkspace2 prompts the user to close the group and discard unsaved changes instead of opening the form. If any queue-management workspace starts using group props, clicking "Visit note form" would trigger that prompt.

packages/esm-service-queues-app/src/current-visit/visit-details/visit-note.component.tsx
```diff
@@ -42,10 +43,16 @@ const TriageNote: React.FC<TriageNoteProps> = ({ notes, patientUuid, diagnoses }
...
-            onClick={() => navigate({ to: `\${openmrsSpaBase}/patient/${patientUuid}/chart` })}
-            iconDescription={t('triageForm', 'Triage form')}>
-            {t('triageForm', 'Triage form')}
+            onClick={() =>
+              launchWorkspace2('service-queues-visit-notes-workspace', { formContext: 'creating' }, null, {
+                patient,
+                patientUuid,
+              })
+            }
+            iconDescription={t('visitNoteForm', 'Visit note form')}>
```
NethmiRodrigo, line 55: Let’s rename the component as well to say `visit-note.component.tsx`

packages/esm-service-queues-app/src/routes.json
UjjawalPrabhat: @NethmiRodrigo Done! Added a `service-queues-patient-context` group and moved both windows to it, so patient-specific group props no longer share a group with the `queue-management` workspaces. 

One note: since only one group opens at a time, launching the form will close an open queue-management workspace group (prompting if there are unsaved changes). Seems acceptable since they're distinct contexts, flag if you'd prefer otherwise.

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
UjjawalPrabhat: Agreed. The names are now single-sourced via constants shared with the `launchWorkspace2` calls, so they can't drift silently, a code rename updates both and a routes.json mismatch throws at launch instead of quietly skipping the refresh. Kept the close-detection itself since the vitals/notes forms are external and don't share the `useVisit` SWR key (added a comment explaining that).

packages/esm-service-queues-app/src/current-visit/visit-details/current-visit-details.component.tsx
UjjawalPrabhat: The obs are now matched by concept UUID (`visitDiagnosesConceptUuid` / `generalPatientNoteConceptUuid`) instead of display strings. Once we match by UUID, the outer `encounterType.display === 'Visit Note'` gate (also a hardcoded English string) seemed redundant, those concepts only live on visit-note encounters. So I figured it's a move off display-string matching, not a behavioral change.

NethmiRodrigo: Is there a guarantee that those concepts only live on visit-note encounters? Especially since they are configurable. It is a behavioural change, since we now no longer explicitly look for obs that are under visit note encounter type. I think we should keep that guard.

UjjawalPrabhat: @NethmiRodrigo Aha, my bad.

Also instead of restoring the hardcoded `'Visit Note'` check, I've gated this on `encounterType.uuid` using a new configurable `visitNoteEncounterTypeUuid`. That preserves the guard while avoiding a locale-dependent display string.

Does this approach seem preferable or would you rather stick with the original check?

NethmiRodrigo: This is good

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat! LGTM

PR #2541, opened 2026-05-20
(fix) O3-5679: De-duplicate patient notes in previous visit tab

UjjawalPrabhat:
## Requirements

## Summary

Fixes patient notes appeared twice in the Service Queues → Previous visit → Notes sub-tab.

### Before
image

### After 
image

## Related Issue

[O3-5679](https://openmrs.atlassian.net/browse/O3-5679)

## Other

[O3-5679]: https://openmrs.atlassian.net/browse/O3-5679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwnasambu (commented):
Thanks so much @UjjawalPrabhat for this. Kindly add a specific test case to verify that "Visit Note" encounters does not produce duplicate UI elements.

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
jwnasambu:
```suggestion
      const processObservations = (observations: Obs[], encounter: Encounter, useObsTime = false) => {
```

Kindly shouldn't we pass the encounter object into the function since it is no longer "inside" the loop scope where encounter is defined?

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
```diff
@@ -108,14 +108,8 @@ const PastVisitSummary: React.FC<PastVisitSummaryProps> = ({ encounters, patient
...
 
-      // Process general observations
       if (encounter?.obs) {
-        processObservations(encounter.obs);
-      }
-
-      // Process Visit Note observations with obs-specific timing
-      if (encounter.encounterType?.display === 'Visit Note' && encounter.obs) {
-        processObservations(encounter.obs, true);
+        processObservations(encounter.obs, encounter.encounterType?.display === 'Visit Note');
```
jwnasambu:
```suggestion
        processObservations(encounter.obs, encounter, encounter.encounterType?.display === 'Visit Note');
```
Pass the current encounter as the new second argument to match the updated parameter list.

Kindly the reason for the above requested changes above is because processObservations is redefined for every single encounter in the array which causes unnecessary memory allocation and
  garbage collection. By moving it outside the loop but still within useMemo ensures it is defined only once per render.
  
  cc @denniskigen kindly feel free to correct me please!

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
```diff
@@ -72,50 +106,8 @@ const PastVisitSummary: React.FC<PastVisitSummaryProps> = ({ encounters, patient
...
-
-      // Process general observations
       if (encounter?.obs) {
-        processObservations(encounter.obs);
-      }
-
-      // Process Visit Note observations with obs-specific timing
-      if (encounter.encounterType?.display === 'Visit Note' && encounter.obs) {
-        processObservations(encounter.obs, true);
+        processObservations(encounter.obs, encounter, encounter.encounterType?.display === 'Visit Note');
```
denniskigen: Really nice catch on the root cause, and the test pins the duplication well.

One timing issue to flag: for Visit Note encounters this now routes through `useObsTime`, which reads `obs.obsDatetime`. The `usePastVisits` custom representation doesn't request `obsDatetime` (the obs fields are `uuid, concept, display, groupMembers, value`), and the REST API only returns requested fields, so `obs.obsDatetime` is `undefined` at runtime. `parseDate` is `dayjs(x).toDate()`, and `dayjs(undefined)` returns the current time, so these notes end up showing "now" rather than the actual note time. Before this change the duplicate pair still included a note with the correct encounter time, so after de-duping only the `obsDatetime` one survives and the displayed time regresses.

The test passes only because the fixture obs carries an `obsDatetime` the real query never returns.

Since `obsDatetime` isn't fetched, the simplest fix is to drop the third argument so `useObsTime` defaults to `false` and the encounter time is used:

```suggestion
        processObservations(encounter.obs, encounter);
```

That keeps the de-duplication and restores the correct time. The `useObsTime` branch then becomes dead code and could be removed for tidiness. If per-obs timing is genuinely wanted, the alternative is adding `obsDatetime` to the obs representation, but that seems unnecessary here.

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
```diff
@@ -57,6 +57,36 @@ const PastVisitSummary: React.FC<PastVisitSummaryProps> = ({ encounters, patient
...
+              : String(diagnosisValue || '');
+
+          diagnoses.push({
+            diagnosis,
+          });
+        } else if (config.concepts.generalPatientNoteConceptUuid === obs?.concept?.uuid) {
+          notes.push({
+            note: String(obs.value || ''),
+            provider: {
+              name: encounter.encounterProviders.length ? encounter.encounterProviders[0].provider.person.display : '',
```
NethmiRodrigo: No optional chaining on encounterProviders[0].provider.person.display or encounterRole.display — a null provider or encounterRole crashes the entire useMemo block, wiping notes, medications, and diagnoses for the whole visit.

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
```diff
@@ -57,6 +57,36 @@ const PastVisitSummary: React.FC<PastVisitSummaryProps> = ({ encounters, patient
...
+              role: encounter.encounterProviders.length ? encounter.encounterProviders[0].encounterRole.display : '',
+            },
+            time: encounter.encounterDatetime ? formatTime(parseDate(encounter.encounterDatetime)) : '',
```
NethmiRodrigo, line 83: A clinician creates a Visit Note encounter at 09:00 and later addends a note observation at 23:55 (obs.obsDatetime = 23:55). The displayed time is 09:00 (encounter start), not 23:55. The old code used obs.obsDatetime specifically for Visit Note encounters; this is an intentional change per the commit message but is a user-visible timestamp regression for any site where obs and encounter datetimes diverge.

NethmiRodrigo: CC @denniskigen

denniskigen: Thanks Nethmi, good thing to scrutinize. I think dropping the obs-time path is actually safe here. The catch is that `usePastVisits` doesn't request `obsDatetime` in its custom representation (the obs fields are just `uuid, concept, display, groupMembers, value`), and OpenMRS REST only returns the fields you ask for. 

If we do want true per-obs timing for addended notes later, I think we'd need to add `obsDatetime` to the representation and bring the obs-time branch back so it has real data to read. Happy for that to be spun out as a follow-up if it's useful, but for this view the encounter time looks right to me.

packages/esm-service-queues-app/src/past-visit/past-visit-details/past-visit-summary.component.tsx
denniskigen: Valid catch, a null `provider`/`person`/`encounterRole` would throw and take the whole `useMemo` down. Worth noting it predates this PR though: the same unguarded `encounterProviders[0].provider.person.display` access is in the medications mapping just below (lines 97-100), and it was in the original closure before it got hoisted. Since these note lines are already in the diff, I'm fine with adding optional chaining here, but it'd be good to harden the medications spot in the same pass so they stay consistent, @UjjawalPrabhat.

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #2543, opened 2026-05-21
(fix) O3-5682: Default the visit queue number attribute so the service queues screen renders

UjjawalPrabhat:
## Requirements

## Summary

The Service Queues screen (`/spa/home/service-queues/screen`) is blank in default installs because the Call→serve flow depends on a "Visit Queue Number" visit attribute. The frontend has long had a `visitQueueNumberAttributeUuid` config knob, but it defaulted to `null` — so `generateVisitQueueNumber` silently no-ops, visits never get a ticket number, and the assignticket POST 200s without storing anything.

This points `visitQueueNumberAttributeUuid._default` at the canonical `Visit Queue Number` UUID (`c0c579b0-8e59-401d-8a4a-976a0b183519`), which is shipped to every install via the demo content package (openmrs-content-referenceapplication-demo). The initializer module creates the VisitAttributeType from that content package on startup, so the default now resolves to a real attribute. Updates one test assertion to match.

Distributions that use their own UUID can still override via config.

## Screenshots

https://github.com/user-attachments/assets/4956c41a-08ed-4dd6-bdd7-a47ce0fe32af

## Related Issue

[O3-5682](https://openmrs.atlassian.net/browse/O3-5682)

## Other

Related [PR](https://github.com/openmrs/openmrs-content-referenceapplication-demo/pull/74)

[O3-5682]: https://openmrs.atlassian.net/browse/O3-5682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


denniskigen:
Confirmed from the backend: [openmrs-module-queue#113](https://github.com/openmrs/openmrs-module-queue/pull/113) adds a Liquibase changeset that inserts the `Visit Queue Number` visit attribute type with exactly this UUID (`c0c579b0-8e59-401d-8a4a-976a0b183519`). The changeset is idempotent (guarded by a `sqlCheck` precondition), so the UUID itself lines up with the paired backend PR.

The release/version piece still needs coordination, though. This frontend now defaults to metadata that is not present in any released queue module version yet: the current latest release, `queue-3.0.0`, does not include that changeset, while this app still declares `queue >=2.4.0 <4.0.0` in `routes.json`. On those currently-supported backend versions, `/queue-entry-number?...visitAttributeType=c0c579b0...` will not find a `VisitAttributeType`, so it silently returns an empty queue number and the queue screen flow remains broken unless the deployment has custom config/metadata.

Could we release the queue module change first and bump this app’s minimum queue dependency to that release as part of the same window, or otherwise hold this frontend change until the backend metadata is available in a released queue module?

NethmiRodrigo:
> Confirmed from the backend: [openmrs-module-queue#113](https://github.com/openmrs/openmrs-module-queue/pull/113) adds a Liquibase changeset that inserts the `Visit Queue Number` visit attribute type with exactly this UUID (`c0c579b0-8e59-401d-8a4a-976a0b183519`). The changeset is idempotent (guarded by a `sqlCheck` precondition), so the UUID itself lines up with the paired backend PR.
> 
> The release/version piece still needs coordination, though. This frontend now defaults to metadata that is not present in any released queue module version yet: the current latest release, `queue-3.0.0`, does not include that changeset, while this app still declares `queue >=2.4.0 <4.0.0` in `routes.json`. On those currently-supported backend versions, `/queue-entry-number?...visitAttributeType=c0c579b0...` will not find a `VisitAttributeType`, so it silently returns an empty queue number and the queue screen flow remains broken unless the deployment has custom config/metadata.
> 
> Could we release the queue module change first and bump this app’s minimum queue dependency to that release as part of the same window, or otherwise hold this frontend change until the backend metadata is available in a released queue module?

That change made to the queue module isn’t the right approach. It should rather be done in the content package. I’ve left a comment on the queue module PR as well

UjjawalPrabhat:
@NethmiRodrigo @denniskigen I've made the correct fix via the initializer instead of the backend change [here](https://github.com/openmrs/openmrs-content-referenceapplication-demo/pull/74). So, could you take another look at this whenever possible?
Thanks!

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -411,7 +411,7 @@ export const configSchema = {
   },
   visitQueueNumberAttributeUuid: {
     _type: Type.String,
-    _default: null,
+    _default: 'c0c579b0-8e59-401d-8a4a-976a0b183519',
```
jwnasambu, line 414: Kindly I noticed that `visitQueueNumberAttributeUuid` is still set to null in other parts of the schema and helper objects:
           * defaultColumnConfig (line 82)
           * Inside columnDefinitions schema (line 315)
To ensure consistency and avoid confusion consider updating these to use the new default or
 explicitly reference the top level default. If null is intended to signal a fallback to the top level config  ensure the code correctly handles this fallback.

denniskigen (approved):
LGTM, @UjjawalPrabhat. Thanks for fixing this end-to-end. 

Process note @NethmiRodrigo, @wikumChamith: since the attribute now ships via the [demo content package](https:/github.com/openmrs/openmrs-content-referenceapplication-demo#74) rather than the queue module, the release to coordinate is the content package v1.8.0. Can we remember to cut that?

packages/esm-service-queues-app/src/config-schema.ts
UjjawalPrabhat: Thanks for flagging. However these two `null`s play different roles, so I'd leave both as-is:

- **`defaultColumnConfig` (line 82)** — dead value. Always overwritten by the top-level config on spread: [`columns.resource.ts:35-41`](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts#L35-L41).
- **Per-column `_default: null` (line 329)** — intentional cascade signal: `null` means *"inherit from the top-level UUID"*, handled by the `??` at [`columns.resource.ts:66-69`](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-service-queues-app/src/queue-table/cells/columns.resource.ts#L66-L69). Hardcoding the canonical UUID here would break that inheritance for implementers overriding the top-level.

So I think the code already correctly handles this fallback.

jwnasambu (approved):

packages/esm-service-queues-app/src/config-schema.ts
jwnasambu: @UjjawalPrabhat Thank for the clear explanation.  I am so happy to learn something new from you too.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -411,7 +411,7 @@ export const configSchema = {
   },
   visitQueueNumberAttributeUuid: {
     _type: Type.String,
```
denniskigen: `Type.UUID` fits better than `Type.String`. It adds UUID-format validation and matches how this file types its other non-concept UUIDs.

```suggestion
    _type: Type.UUID,
```

denniskigen (approved):

PR #2545, opened 2026-05-22
(chore) Use reusable build and release GHA workflows

NethmiRodrigo:
## Requirements

## Summary

Replaces inline CI job definitions in .github/workflows/ci.yml with calls to reusable workflows from [openmrs/openmrs-contrib-gha-workflows](https://github.com/openmrs/openmrs-contrib-gha-workflows).

## Other

The translations seem to be from a left over clean up


.github/workflows/ci.yml
```diff
@@ -10,134 +10,21 @@ on:
...
-        with:
-          name: packages
-          path: |
-            packages/**/dist
-          overwrite: true
+    uses: openmrs/openmrs-contrib-gha-workflows/.github/workflows/build-frontend-module.yml@main
+    with:
+      verify_command: yarn verify
 
   release:
```
denniskigen, line 23: Thanks for moving this over to the shared workflow. Could we preserve the old manual-run behavior here? Previously, `workflow_dispatch` only ran the build job, while prerelease publishing and RefApp deployment were limited to `push` on `main`. The reusable release workflow also runs `pre_release` and deployment for `workflow_dispatch`, so manually running this workflow on `main` would now publish `next` packages and trigger the RefApp build. Adding an `if` to skip this job for `workflow_dispatch`, or removing the manual trigger if it is no longer needed, would avoid an accidental release.

NethmiRodrigo: Thanks for flagging it! Added `if: github.event_name != 'workflow_dispatch'` to the `release` job — manual runs will now only execute the build, matching the previous behavior.

denniskigen (approved):
Nice cleanup. Verified locally that `yarn turbo extract-translations` against `main` produces exactly the en.json deletions in this PR, so the translation removals are safe (orphaned keys; `translateFrom` fallbacks handle runtime). The workflow consolidation is a clear maintainability win.

Two non-blocking heads-ups for whoever watches the first run on main:
- Make sure `OMRS_BOT_GH_TOKEN` is set at the repo/org level — the new RefApp trigger uses it via `BOT_PAT` and (unlike the old Bamboo step) isn't `continue-on-error`.
- The reusable runs build before verify, where the old job ran verify first. Same outcome in most cases, just a behavior change worth knowing.

PR #2554, opened 2026-05-29
(feat) O3-5714: Improve performance with multi-paged patient search results

ibacher:
## Requirements

## Summary

Several notable fixes for the compact-patient-search component:

* Previously, each call to `useSWRInfinite()` was, in essence, returning new copies of existing data, which lead to unnecessary re-renders. This has been fixed by setting `keepPreviousData` to `true`
* Previously, when `useSWRInfinite()` set `isLoading` to `true`, all results were replaced with placeholder. Since this gets set to `true` even when, e.g., loading a second page of results, this resulted in unnecessary unmounts and recreations, which was both a memory leak and lead to a stampede of requests.
* All search results were previously rendered and retained. This leads to stampedes of secondary queries (e.g., for patient photos, queue status, visit status, etc.). Instead, this PR moves us to using list virtualization so that roughly 10-18 items are retained in a React query. Large scrolls backwards or forwards have some slight glitches as components are re-rendered, but the easiest way to fix this is in the components rendering the secondary queries themselves.
* Recent patients, however, are not virtualized at all, as we rely on this being a limited list that is safe to load.
* A mistake in the call to `useArrowNavigation()` resulted in the arrow keys only being usable for up to the number of `recentPatients`
* This PR adds additional component-level memoization where it makes sense to speed up rendering. In addition, in the context of search results, we re-tune SWR to more aggressively cache so that there's less need to re-try, e.g., patient photo queries
* Additionally, this PR makes two fixes the service queues banner tag. First, if it is rendered outside the patient chart, a wrapper component is used so no query needs to be issued. Second, the query that was used overfetched data, especially as the queue details in the modal are basically reloaded, so the representation has been updated to minimize the data requested. This is one of the biggest contributing components.

## Screenshots

Patient Search Screenshot

This illustrates the trade-off here. Note the white patches while rows are reloading:

patient-search

## Related Issue

https://issues.openmrs.org/browse/O3-5714

## Other

Follow-ups: 

* ~~New billing PAID bill needs same wrapper treatment as queues app~~ [PR 749](https://github.com/openmrs/openmrs-esm-billing-app/pull/749)
* Active Visit tag needs SWR tweak
* Might do small rework on central components in styleguide

### DIFF
--- a/packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
+++ b/packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
@@ -1,5 +1,7 @@
-import React, { forwardRef, useCallback, useMemo } from 'react';
+import React, { forwardRef, useEffect, useImperativeHandle, useMemo, useRef, useState } from 'react';
 import classNames from 'classnames';
+import { Loading } from '@carbon/react';
+import { useVirtualizer } from '@tanstack/react-virtual';
 import {
   ConfigurableLink,
   getPatientName,
@@ -8,11 +10,33 @@ import {
   PatientPhoto,
   useConfig,
 } from '@openmrs/esm-framework';
-import type { SearchedPatient } from '../types';
 import { type PatientSearchConfig } from '../config-schema';
+import Loader from './loader.component';
 import { usePatientSearchContext } from '../patient-search-context';
 import { mapToFhirPatient } from '../utils/fhir-mapper';
+import type { SearchedPatient } from '../types';
 import styles from './compact-patient-banner.scss';
+
+// OpenMRS base assumption
+const remInPixels = 16;
+
+// Estimated row height used before a row is measured. Rows are measured individually and may grow
+// (roughly 5.625rem–7.5rem), so this only needs to be close.
+const estimatedRowHeight = 5.625 * remInPixels;
+
+// Rows rendered above and below the viewport. With the ~4-row viewport this keeps roughly 16
+// patients mounted as a scroll buffer. This is independent of pagination: see the fetch trigger,
+// which keys off the visible range, so buffering more rows doesn't pull an extra page over the wire.
+const overscan = 6;
+
+// Start loading the next page when the visible range reaches within this many rows of the end, so
+// the next page arrives just before the user hits the bottom rather than stalling there.
+const fetchLookAhead = 2;
+
+export interface CompactPatientBannerHandle {
+  /** Scroll the result at `index` into view and move focus to it (for arrow-key navigation). */
+  focusSearchResult: (index: number) => void;
+}
 
 interface ClickablePatientContainerProps {
   children: React.ReactNode;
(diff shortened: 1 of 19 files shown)


packages/esm-patient-search-app/src/compact-patient-search/patient-search.component.tsx
```diff
@@ -89,12 +59,23 @@ const PatientSearch = React.forwardRef<HTMLDivElement, PatientSearchProps>(
...
-              </div>
-            )}
+            {/* Set SWRConfig to minimize revalidations of, e.g., the patient photo */}
+            <SWRConfig
+              value={{
+                revalidateIfStale: false,
+                revalidateOnFocus: false,
+                revalidateOnReconnect: false,
+                dedupingInterval: 180_000, // 3 minutes
+                keepPreviousData: true,
```
denniskigen: This nested `keepPreviousData`, combined with row reuse, can briefly show the wrong patient's data. The virtualizer has no `getItemKey`, so rows are keyed by index: on a query refinement the row at a given index is the same instance with a new `patient` prop rather than a remount. The per-row queries underneath inherit `keepPreviousData` from this `SWRConfig` (the photo via `usePatientPhoto`, and tags like Active Visit), so when a row's patient changes they return the previous patient's response while the new one loads. The row updates the name immediately but keeps the prior patient's photo/visit tag for a beat.

Could we give the virtualizer `getItemKey: (i) => patients[i].uuid` so rows remount per patient, or drop `keepPreviousData` from this nested config (keeping it on the search query itself)? One thing to keep in mind either way: this config governs the whole banner subtree, so anything added to `patient-banner-tags-slot` / `patient-banner-bottom-slot` inherits the 3-minute dedupe and no-revalidate too.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -83,6 +83,7 @@ export function useInfinitePatientSearch(
   const { data, isLoading, isValidating, setSize, error, size } = useSWRInfinite<InfinitePatientSearchResponse, Error>(
     shouldFetch ? getUrl : null,
     openmrsFetch,
+    { keepPreviousData: true },
```
denniskigen, line 86: Good regression coverage for preserving results while an active query loads. There is one inactive-query case left: with `keepPreviousData`, once the search is cleared (`shouldFetch` false, key null) the hook keeps returning the last `data`, `hasMore`, and `totalResults` (I confirmed this with a quick test: after clearing it still reports 10 results, `hasMore: true`, `totalResults: 100`). In the compact extension, arrow navigation is sized off `patients?.length` and is not gated on `showSearchResults`, so after clearing, an arrow press plus Enter can select a now-hidden stale result. The sibling `compact-patient-search.component.tsx` is safe here because it gates the count on `hasSearchTerm`.

Could we gate the returned data and metadata on `shouldFetch` in the hook (which covers every consumer), or only keep previous data while loading another non-empty query?

packages/esm-patient-search-app/src/compact-patient-search/compact-patient-banner.component.tsx
```diff
@@ -8,41 +10,195 @@ import {
...
+ * Renders the patient search results as a list of compact banners. By default the list is
+ * virtualized and infinitely-scrolling, mounting only the rows near the viewport.
+ */
+const CompactPatientBanner = forwardRef<CompactPatientBannerHandle, CompactPatientBannerProps>(
+  ({ patients, hasMore = false, isValidating = false, fetchMore, virtualize = true }, ref) => {
+    const scrollContainerRef = useRef<HTMLDivElement>(null);
+    const [pendingFocusIndex, setPendingFocusIndex] = useState<number | null>(null);
+    // Rows we've already shown as real banners, so a resumed scroll doesn't replace the rows that
+    // are still on screen with skeletons
+    const seenIndices = useRef<Set<number>>(new Set());
```
denniskigen: Minor, and related to the row keying: `seenIndices` is tracked by row index and never reset. Because `keepPreviousData` keeps this banner mounted across query refinements, indices marked "seen" in one query stay seen for the next, so `showSkeleton` stays false for those positions and a fast scroll on a refined search mounts real rows (firing their photo queries) rather than deferring until scroll settles. It is bounded by the virtual window and the per-uuid cache, so it does not bring back the full stampede, and a full clear of the search resets it, so this is low priority.

Clearing the set when the result set changes (or tracking seen rows by uuid) would restore the deferral. Worth noting this path is not exercised by tests, since the virtualizer mock hardcodes `isScrolling: false`.

ibacher:
@denniskigen Thanks for the thorough review! I think I've addressed the items you raised.

denniskigen (approved):

PR #2555, opened 2026-06-01
(fix) O3-5680: Make calling status and drug order type filters configurable

UjjawalPrabhat:
## Requirements

## Summary

Two status/filter values in the queues UI were hardcoded in the frontend, so deployments couldn't adapt them to a different signage label or to renamed/localized order types. This PR makes both configurable while keeping defaults that preserve existing behavior.

## Related Issue

[O3-5680](https://openmrs.atlassian.net/browse/O3-5680)

## Other

[O3-5680]: https://openmrs.atlassian.net/browse/O3-5680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwnasambu (approved):

jwnasambu (commented):
This L.G.T.M @UjjawalPrabhat  my only recommendation is to ensure that the unit tests
  are updated to verify that the components correctly react to changes in these new configuration values and also ensuring that
  future updates don't inadvertently revert to hardcoded behavior.

Once again thank You!

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -121,6 +122,14 @@ export const configSchema = {
     },
   },
   biometrics: biometricsConfigSchema,
+  callingStatus: {
+    _type: Type.String,
+    _default: 'calling',
+    _description:
+      'The status string sent to the queueutil/assignticket endpoint when a patient is called, ' +
+      'and matched on the queue screen to trigger the blinking ticket display. Change to match ' +
+      'what your digital signage expects (e.g. "Now serving").',
```
denniskigen: Optional: the queue backend reserves `completed` as a signal to clear a ticket from the board, so it's the one value this should never be set to. Might be worth noting it here, e.g. appending: `Avoid "completed", which the queue backend treats as a signal to remove the ticket.`

denniskigen:
Nice, focused change @UjjawalPrabhat. I traced both values end to end and the defaults preserve current behavior: `callingStatus` defaults to `calling`, and `131168f4-15f5-102d-96e4-000c29c2a5d7` is core's Drug Order type (the one whose display is `Drug Order`), so the new uuid match is equivalent to the old display match on standard data, and more robust against localization or renamed order types. The extraction is complete too: the only `calling` literal left is the schema default, and the call action and the queue-screen matcher now both read the same config key.

I also checked the queue module backend. `queueutil/assignticket` stores whatever `status` string you send verbatim and `active-tickets` echoes it back unchanged, so an overridden `callingStatus` round-trips correctly and the blinking match still holds. One nuance worth guarding against: `completed` is special-cased server-side (it clears the ticket from the board), so that's the one value `callingStatus` should never be set to. Small inline suggestion on that below.

The only real follow-up is the one @jwnasambu raised: neither change is covered by a test. A queue-screen test asserting that a ticket whose status equals `callingStatus` gets the blinking class, plus a medication test asserting that an order with the configured `orderType.uuid` renders, would lock the behavior in. Not blocking, but worthwhile given the PR is specifically about config-driven behavior.

packages/esm-active-visits-app/src/visits-summary/visits-components/medications-summary.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+      <MedicationSummary
+        medications={[
+          buildOrderItem('drug-order', configDefaults.drugOrderTypeUuid, 'aspirin'),
+          buildOrderItem('lab-order', labOrderTypeUuid, 'complete-blood-count'),
+        ]}
+      />,
+    );
+
+    expect(screen.getByText('Aspirin')).toBeInTheDocument();
+    expect(screen.queryByText('Complete-blood-count')).not.toBeInTheDocument();
```
NethmiRodrigo:
```suggestion
    expect(screen.getByText(/aspirin/i)).toBeInTheDocument();
    expect(screen.queryByText(/complete-blood-count/i)).not.toBeInTheDocument();
```

NethmiRodrigo (approved):
Minor nit: but otherwise LGTM

NethmiRodrigo (approved):

PR #2564, opened 2026-06-05
(fix) O3-5681: Queue-screen widget polls regardless of user activity

UjjawalPrabhat:
## Requirements

## Summary

The Service Queues app polled at fixed intervals regardless of user activity, the queue screen hit `active-tickets` every 3s nonstop (1,200 req/hr per tab), even on an unattended waiting-room monitor.

Polling is now activity-aware: each surface has a configurable active rate and an idle rate that kicks in after 60s without interaction.

| Surface | Before | Active | Idle |
| --- | --- | --- | --- |
| Queue screen | 3s fixed | 5s | 10s |
| Dashboards | 60s fixed | 60s | 120s |

No surface polls faster than before; the queue screen drops ~70% of its requests (1,200 → 360 req/hr) while staying board-fresh.

## Related Issue

[O3-5681](https://openmrs.atlassian.net/browse/O3-5681)

## Other

[O3-5681]: https://openmrs.atlassian.net/browse/O3-5681?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/activityAwareRefreshInterval.ts
```diff
@@ -0,0 +1,48 @@
+export const ACTIVE_REFRESH_INTERVAL = 5_000;
+export const IDLE_REFRESH_INTERVAL = 30_000;
+const IDLE_TIMEOUT = 60_000;
```
NethmiRodrigo: I feel like these are values we might want to make configurable because implementations might be aware of these intervals for themselves and can tweak them to their requirements if need be

packages/esm-service-queues-app/src/activityAwareRefreshInterval.ts
```diff
@@ -0,0 +1,48 @@
+export const ACTIVE_REFRESH_INTERVAL = 5_000;
+export const IDLE_REFRESH_INTERVAL = 30_000;
```
NethmiRodrigo: Also, if the queue screen is a wall-mounted, unattended board, with no interaction it settles to the 30s idle interval (was a fixed 3s). ~30s worst-case latency for a called ticket to appear may be too slow for a waiting-room display. `revalidateOnFocus` doesn't help an unattended board.

packages/esm-service-queues-app/src/activityAwareRefreshInterval.ts
```diff
@@ -0,0 +1,48 @@
...
+    isActive = false;
+  }
+};
+
+const initialize = () => {
+  if (initialized) {
+    return;
+  }
+  initialized = true;
+  interactionEvents.forEach((event) => window.addEventListener(event, markActive));
```
NethmiRodrigo: Minor: these listeners never call `preventDefault`, so registering them with `{ passive: true }` avoids any chance of scroll-jank from the `scroll`/`touchstart`/`mousemove` handlers.

packages/esm-service-queues-app/src/activityAwareRefreshInterval.ts
UjjawalPrabhat: Agreed. And since we can't reliably tell an unattended wall board from someone actively watching, I kept the queue screen activity-aware but gave it its own, much tighter idle floor: **5s active / 10s idle** (configurable). So an unattended board now sits at 10s rather than the old fixed 3s — still board-fresh, ~70% less load and never falls back to the dashboard's 60s. `revalidateOnFocus` stays as a bonus for the attended case.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -255,6 +255,35 @@ export const configSchema = {
     _default: defaultDrugOrderTypeUuid,
     _description: 'The UUID of the "Drug Order" order type, used to filter medications in the previous-visit view.',
   },
+  refreshIntervals: {
+    dashboard: {
+      active: {
+        _type: Type.Number,
+        _default: 30000,
```
NethmiRodrigo: The current default active interval is `60000`, so this would double the request rate. Lets leave this to what it was and increase the idle to say `120000`?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -255,6 +255,35 @@ export const configSchema = {
     _default: defaultDrugOrderTypeUuid,
     _description: 'The UUID of the "Drug Order" order type, used to filter medications in the previous-visit view.',
   },
+  refreshIntervals: {
+    dashboard: {
+      active: {
+        _type: Type.Number,
```
NethmiRodrigo, line 294: Also, since these are bare millisecond numbers, it'd be nice to guard against someone writing seconds by mistake. `queueScreen.active: 5` would give us a ~200 req/s loop, which is a lot worse than the 3s we're fixing. validators is already imported here:

```suggestion
        _type: Type.Number,
        _validators: [validators.inRange(1000, 3600000)],
```

packages/esm-service-queues-app/src/queue-screen/useActiveTickets.ts
```diff
@@ -1,11 +1,21 @@
...
+  } = useConfig<ConfigObject>();
+  const refreshInterval = useMemo(
+    () => getActivityAwareRefreshInterval(queueScreen.active, queueScreen.idle),
+    [queueScreen.active, queueScreen.idle],
+  );
   const { data, isLoading, error, mutate } = useSWR<{ data: Record<string, { status: string; ticketNumber: string }> }>(
     `${restBaseUrl}/queueutil/active-tickets`,
     openmrsFetch,
-    { refreshInterval: 3000 },
+    { refreshInterval, revalidateOnFocus: true },
```
NethmiRodrigo: Do we need `revalidateOnFocus` here, since an unattended board never gets a focus event? It's already refetching every 10s on the idle interval. Also  in a normal tab, since the framework's `defaultSwrConfig` sets `focusThrottleInterval`: `1800000` and we don't override it, SWR would trigger it once every 30 minutes.

packages/esm-service-queues-app/src/activity-aware-refresh-interval.test.ts
```diff
@@ -0,0 +1,61 @@
...
+    expect(refreshInterval()).toBe(IDLE_REFRESH_INTERVAL);
+  });
+
+  it('returns to the active interval when the user interacts again', () => {
+    vi.advanceTimersByTime(IDLE_TIMEOUT + 1);
+    expect(refreshInterval()).toBe(IDLE_REFRESH_INTERVAL);
+
+    window.dispatchEvent(new Event('mousemove'));
+    expect(refreshInterval()).toBe(ACTIVE_REFRESH_INTERVAL);
+  });
```
NethmiRodrigo, line 41: `markActive` does two things: mark active, and restart the countdown, but the tests only check the first. If you deleted the `clearTimeout` line, all five tests would still pass, but the app would break. 

This test would make sure that an active user wouldn’t be treated as idle.

```suggestion
  });
  
  it('keeps the active interval when interaction keeps resetting the idle countdown', () => {
     vi.advanceTimersByTime(IDLE_TIMEOUT / 2);
     window.dispatchEvent(new Event('mousemove'));
     vi.advanceTimersByTime(IDLE_TIMEOUT / 2 + 1);

     expect(refreshInterval()).toBe(ACTIVE_REFRESH_INTERVAL);
   });
```

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -255,6 +255,35 @@ export const configSchema = {
...
+        _default: 30000,
+        _description:
+          'How often (in ms) the queue dashboards poll for updates while the user is actively interacting with a visible tab.',
+      },
+      idle: {
+        _type: Type.Number,
+        _default: 60000,
+        _description: 'How often (in ms) the queue dashboards poll for updates once the tab is idle or hidden.',
```
NethmiRodrigo: Is the `hidden` part correct? It gives the assumption that a tab in the background would refresh, but  SWR skips the fetch entirely if a tab is in the background.

packages/esm-service-queues-app/src/activityAwareRefreshInterval.ts
NethmiRodrigo: Since both the callers now directly use the config values, lets remove this

packages/esm-service-queues-app/src/activity-aware-refresh-interval.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const initialize = () => {
+  if (initialized) {
+    return;
+  }
+  initialized = true;
+  interactionEvents.forEach((event) => window.addEventListener(event, markActive, { passive: true }));
+  document.addEventListener('visibilitychange', handleVisibilityChange);
+  markActive();
+};
```
NethmiRodrigo, line 40: As it is, this would make a simple mouse move fire resetting the timer way too many times: 
```suggestion
const throttledMarkActive = throttle(markActive, 1000, { leading: true, trailing: false });

const initialize = () => {
  if (initialized) {
    return;
  }
  initialized = true;
  interactionEvents.forEach((event) => window.addEventListener(event, throttledMarkActive, { passive: true }));
  document.addEventListener('visibilitychange', handleVisibilityChange);
  markActive();
};

```

packages/esm-service-queues-app/src/views/queue-table-by-status-view.component.tsx
```diff
@@ -12,7 +13,11 @@ interface QueueTableByStatusViewProps {
 const QueueTableByStatusView: React.FC<QueueTableByStatusViewProps> = ({ queueUuid }) => {
   const { queue, isLoading: isLoadingQueue, error } = useQueue(queueUuid);
```
NethmiRodrigo, line 21: Since this is not inside the SWR wrapper, it won’t get the configured refresh interval. It does currently because `PatientQueueHeader` is inside the wrapper and fetches the same URL.

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat! I left a starter feedback, I’m still going through, sorry for the delay

packages/esm-service-queues-app/src/swr-config.component.tsx
```diff
@@ -0,0 +1,23 @@
...
+const SwrConfig: React.FC<PropsWithChildren> = ({ children }) => {
+  const {
+    refreshIntervals: { dashboard },
+  } = useConfig<ConfigObject>();
+  const refreshInterval = useMemo(
+    () => getActivityAwareRefreshInterval(dashboard.active, dashboard.idle),
+    [dashboard.active, dashboard.idle],
+  );
+
+  return <SWRConfig value={{ refreshInterval }}>{children}</SWRConfig>;
```
denniskigen: One case still needs covering here. When a hidden dashboard tab becomes visible, its pending poll timer remains scheduled. Since polling is skipped while hidden and `revalidateOnFocus` is currently disabled, the user can continue seeing stale cached data for up to `dashboard.idle` (120 seconds) after returning. `markActive` changes the rate, but SWR only reads the interval function when scheduling the next tick. Before this PR, the maximum wait after returning was 60 seconds.

The linked ticket also explicitly requires `revalidateOnFocus: true` for this case. The framework's 30-minute `focusThrottleInterval` would allow the first focus refresh but suppress subsequent ones during that window, so could we override both values here?

```suggestion
  return (
    <SWRConfig value={{ refreshInterval, revalidateOnFocus: true, focusThrottleInterval: 10_000 }}>
      {children}
    </SWRConfig>
  );
```

denniskigen (commented):
This is very close, @UjjawalPrabhat. Nice work addressing all of @NethmiRodrigo's feedback.

denniskigen (approved):

PR #2577, opened 2026-06-11
(fix) O3-5730: Automatically check in patient after starting a visit

NethmiRodrigo:
## Requirements

## Summary

When a scheduled patient has no active visit, clicking **Check in** opens the start visit workspace instead of checking them in directly. Previously, once the visit was created and the workspace closed, the appointment status was **not** updated — so the user had to click **Check in** a second time (now that an active visit exists) to actually check the patient in.

This PR wires the start visit workspace's `onVisitStarted` callback to set the appointment status to `CheckedIn` automatically. The patient is now checked in as soon as the visit is created, with no second click.

### Guarding against a double check-in

When patient-chart's `showUpcomingAppointments` is enabled, the visit form renders an "upcoming appointments" card. If the user selects the appointment there, the form checks it in via its own `onVisitCreatedOrUpdated` callback **before** `onVisitStarted` fires. Our auto check-in would then try to set `CheckedIn` a second time, which the backend rejects — the `DefaultAppointmentStatusChangeValidator` in `openmrs-module-appointments` disallows a `CheckedIn -> CheckedIn` transition (`toStatus.getSequence() <= currentStatus.getSequence()`).

To avoid this, the auto check-in first reads the appointment's current status (`GET /appointment?uuid={uuid}`) and skips the status change if it is already `CheckedIn`. If the status cannot be read, it attempts the check-in anyway rather than silently skipping it.

Implementation notes:
- Extracted the status-change + snackbar logic into a small local `checkIn(successSubtitle)` helper, reused by both the existing-active-visit path and the new auto-check-in path.
- The `customUrl` override and the existing-active-visit behavior are unchanged.
- Because the visit form only invokes `onVisitStarted` after a **successful** visit creation, a failed visit creation leaves the appointment status unchanged.

Added `checkin-button.test.tsx` covering:
- `customUrl` configured → navigates, does not change status or launch the workspace
- active visit exists → checks in directly, no workspace
- no active visit → launches the workspace but does not check in yet
- visit creation fails (`onVisitStarted` never fires) → appointment status unchanged
- visit started successfully (status not `CheckedIn`) → appointment status changes to `CheckedIn` and visits are mutated
- visit form already checked the appointment in (status `CheckedIn`) → auto check-in is skipped
- appointment status cannot be read → check-in is attempted anyway

## Screenshots

Before:

https://github.com/user-attachments/assets/eefe2e7b-b2cc-4cbe-a20f-39f2ced20b9d

After:

https://github.com/user-attachments/assets/050f5ed5-4866-4477-961f-37e5a6863ed9

## Related Issue
<!-- No Jira ticket. —>
https://openmrs.atlassian.net/browse/O3-5730


packages/esm-appointments-app/src/appointments/common-components/checkin-button.component.tsx
```diff
@@ -35,39 +61,33 @@ const CheckInButton: React.FC<CheckInButtonProps> = ({ appointment, patientUuid,
...
+          onVisitStarted: async () => {
+            mutateVisits();
+            // The visit form's "upcoming appointments" card may have already checked this appointment
+            // in. The backend rejects a CheckedIn -> CheckedIn transition, so only check in if the
+            // appointment isn't already checked in. If the status can't be read, attempt the check-in
+            // anyway rather than silently skipping it.
+            try {
+              if ((await getAppointmentStatus(appointment.uuid)) === 'CheckedIn') {
+                return;
+              }
```
denniskigen, line 84: Nice work, @NethmiRodrigo! 

One small suggestion: when we skip the auto check-in here, the appointments table only refreshes because the visit form's upcoming appointments card happens to call `mutateAppointments()` itself after checking the appointment in. If the appointment reached `CheckedIn` some other way (say, another user or tab while the form was open), the table would keep showing the stale status until the next revalidation. Mutating before the early return would keep the table consistent in both cases:

```suggestion
              if ((await getAppointmentStatus(appointment.uuid)) === 'CheckedIn') {
                mutateAppointments?.();
                return;
              }
```

What do you think?

denniskigen (dismissed):

packages/esm-appointments-app/src/appointments/common-components/checkin-button.component.tsx
NethmiRodrigo: Done!

NethmiRodrigo:
@chibongho can you give this a review :)

ibacher (approved):

chibongho:
> @chibongho can you give this a review :)

Sorry for the late review; I was off last Friday. LGTM.

PR #2578, opened 2026-06-11
(fix) O3-5731: De-duplicate appointments occurring later today in the upcoming appointments card

NethmiRodrigo:
## Requirements

## Summary

The patient chart's "Upcoming appointments" card sometimes rendered the same appointment twice.

`usePatientAppointments` derives `todaysAppointments` and `upcomingAppointments` from the same source array using overlapping filters:

- `upcomingAppointments` → `startDateTime` `isAfter(now)`
- `todaysAppointments` → `startDateTime` `isToday()`

An appointment scheduled **later today** (after the current moment) satisfies *both* conditions, so it appears in both lists. The card concatenated the two lists without deduping (`todaysAppointments.concat(futureAppointments)`), causing such appointments to render twice. Appointments dated tomorrow or earlier today were unaffected, which is why duplication only showed up "some of the time for some of the appointments".

This PR dedupes the combined list by `uuid` so each appointment renders exactly once. The fix is scoped to the card component; the resource hook's semantics are left unchanged because `patient-appointments-detailed-summary.extension.tsx` relies on today's appointments appearing in `upcomingAppointments` for its "Upcoming" tab.

## Screenshots

I can’t seem to visually replicate this anymore for some reason.

## Related Issue

[No Jira ticket.](https://openmrs.atlassian.net/browse/O3-5731)

## Other

Added `patient-upcoming-appointments-card.test.tsx` with two cases:
1. An appointment later today (present in both `todaysAppointments` and `upcomingAppointments`) renders only once. Verified this case fails against the pre-fix code.
2. Two genuinely distinct appointments both render — guards against over-deduping.


packages/esm-appointments-app/src/patient-appointments/patient-upcoming-appointments-card.component.tsx
```diff
@@ -93,6 +93,9 @@ const PatientUpcomingAppointmentsCard: React.FC<PatientUpcomingAppointmentsProps
 
   const appointments = todaysAppointments
     .concat(futureAppointments)
+    // An appointment later today satisfies both `isToday` and `isAfter(now)`, so it appears in
+    // both lists; dedupe by uuid to avoid rendering it twice.
+    .filter((appointment, index, self) => index === self.findIndex((a) => a.uuid === appointment.uuid))
```
ibacher: Rather than refiltering, wouldn't it make more sense to make the definition of `isAfter()` mean "is greater than today"?

NethmiRodrigo: Done — moved the fix to the source: `upcomingAppointments` now compares against `endOf('day')` instead of `now`, so later-today appointments stay only in `todaysAppointments`. Dropped the component dedup. This also fixes the same duplication in the detailed-summary tabs.

jayasanka-sack (approved):
Thanks for getting this done @NethmiRodrigo!

PR #2583, opened 2026-06-13
(fix) O3-5693: Trim queue entry representation to what the table renders

UjjawalPrabhat:
## Requirements

## Summary

`useQueueEntries` requested the full visit → encounters → obs/diagnoses/providers graph, deep `queue`/`status`/`priority` concept reps, and the full `person` — ~95% of which the queue table never displays. Because these are lazy associations, requesting them forces them to load on the server.

This trims the representation to only the fields the table and its actions/modals actually read, cutting both the payload and the server-side lazy-load cascade.

## Related Issue

[O3-5693](https://openmrs.atlassian.net/browse/O3-5693)

## Other

Relate PRs' [here](https://github.com/openmrs/openmrs-module-queue/pull/114) and [here](https://github.com/openmrs/openmrs-module-queue/pull/117)

[O3-5693]: https://openmrs.atlassian.net/browse/O3-5693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
```diff
@@ -6,7 +6,7 @@ import { type QueueEntry, type QueueEntrySearchCriteria } from '../types';
 const queueEntryBaseUrl = `${restBaseUrl}/queue-entry`;
 
 const repString =
-  'custom:(uuid,display,queue,status,patient:(uuid,display,person,identifiers:(uuid,display,identifier,identifierType)),visit:(uuid,display,startDatetime,encounters:(uuid,display,diagnoses,encounterDatetime,encounterType,obs,encounterProviders,voided),attributes:(uuid,display,value,attributeType)),priority,priorityComment,sortWeight,startedAt,endedAt,locationWaitingFor,queueComingFrom,providerWaitingFor,previousQueueEntry)';
+  'custom:(uuid,display,queue:(uuid,display,name),status:(uuid,display),patient:(uuid,display,person:(uuid,display,age,birthdate,gender),identifiers:(uuid,identifier,identifierType:(uuid,display))),visit:(uuid,startDatetime,attributes:(uuid,value,attributeType:(uuid))),priority:(uuid,display),priorityComment,sortWeight,startedAt,endedAt,queueComingFrom:(uuid,display),previousQueueEntry:(uuid,startedAt,status:(uuid,display)))';
```
NethmiRodrigo, line 9: The trimmed rep in `useQueueEntries` no longer fetches display on each identifier (it now requests `identifiers:(uuid,identifier,identifierType:(uuid,display))`, dropping the display the old rep had). This `<Tag>{identifier.display}</Tag>` will render empty here, since this modal receives the queue-table row entry sourced from that rep. Could you either add display back to the identifiers sub-rep, or render a field that is still fetched (e.g. `identifier.identifier`)? The `Identifer` type still declares display, so TypeScript doesn't flag it.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
NethmiRodrigo, line 1: Since the default rep no longer fetches `locationWaitingFor`, `providerWaitingFor`, or `visit.encounters`/`visitType`, the `QueueEntry` type now overstates what's present at runtime. Not blocking (no live reader today), but narrowing the type would have caught the identifier display issue at compile time.

UjjawalPrabhat: @NethmiRodrigo Fixed the blocking one. For the type, I made `Identifer.display` optional since that was the only dropped field with a live reader. So, the compiler now catches unguarded `identifier.display` reads.

Left the rest (`locationWaitingFor`, `providerWaitingFor`, `visit.encounters`/`visitType`) as-is because `QueueEntry` is shared with the single-entry rep (`useQueueEntry`/`useLatestQueueEntry`) that still fetches them. Fully narrowing would mean splitting into per-rep types. What do you think?

jwnasambu (approved):
LGTM. Thanks @UjjawalPrabhat

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
denniskigen, line 9: Thanks for carefully tightening this rep. Could we add a focused test to protect it? The initial `identifier.display` regression passed the existing suite: the queue entry mocks aren't shaped like this representation, and the call modal test doesn't assert identifier rendering, so nothing failed when a rendered field stopped being fetched. A minimal fixture shaped exactly like this custom rep, run through the configured columns and the Call and Undo actions, would catch any future trim that drops a field something still reads.

denniskigen (approved):
LGTM! Thanks, @UjjawalPrabhat. Verified the trimmed rep against dev3: every field the table columns, action modals, and extension slots read is still requested, `previousQueueEntry` still returns explicit nulls (which the undo action visibility checks rely on), and the payload for the same 10 entries drops from 92KB to 6.8KB.

packages/esm-service-queues-app/src/hooks/useQueueEntries.ts
denniskigen: One more reason for a narrower type: the extension column contract. `queueTableExtensionColumn` passes `state={{ queueEntry }}` to any configured `queue-table-*-slot`, so downstream extensions see whatever this rep fetches, and it no longer supplies `locationWaitingFor`, `providerWaitingFor`, `queue.location`, `queue.service`, or `visit.encounters`. I couldn't find an O3 extension reading those fields, so this doesn't block the optimization. But a `QueueEntrySummary` type used by `useQueueEntries` and the extension slot state would document what extensions can rely on.

packages/esm-service-queues-app/src/hooks/useQueueEntries.rep.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+import CallQueueEntryModal from '../modals/call-modal/call-queue-entry.modal';
+import UndoTransitionQueueEntryModal from '../modals/undo-transition-queue-entry.modal';
+import QueueTable from '../queue-table/queue-table.component';
+
+// `identifierTypeUuid` default of `defaultColumnConfig`, so the built-in patient-identifier column renders this identifier.
+const openmrsIdTypeUuid = '05a29f94-c0ed-11e2-94be-8c13b969e334';
+const visitQueueNumberAttributeUuid = 'queue-number-visit-attr-type-uuid';
+
+// Shaped exactly like the custom representation fetched by useQueueEntries: no fields beyond what it requests.
+// Rendering it through the columns and row actions fails if a future trim drops a field something still reads.
```
denniskigen: Thanks for adding this, @UjjawalPrabhat. The rendering coverage is useful, and the fixture currently matches the representation. One gap is that nothing ties the fixture to `repString` itself. I removed `birthdate` from the representation and all three tests stayed green because the fixture is maintained separately and `vi.mock` replaces the real module. A future trim could therefore reintroduce the regression this test is intended to catch.

Could we make the relationship self-enforcing by exporting `repString` and asserting that the fixture's key tree matches the parsed representation? I verified this locally: all four tests pass, and removing `birthdate` from `repString` produces a failure identifying the missing field.

The hook mock can preserve the real exports while replacing only `useMutateQueueEntries`:

```ts
vi.mock('../hooks/useQueueEntries', async (importOriginal) => ({
  ...((await importOriginal()) as object),
  useMutateQueueEntries: () => ({ mutateQueueEntries: vi.fn() }),
}));
```

Then import `repString` and add the parsed-representation/key-tree assertion before the rendering tests.

denniskigen: Thanks, @UjjawalPrabhat! Nice touch with the regex parser.

denniskigen (approved):

PR #2571, opened 2026-06-09
(feat) O3-5727: Create useAppointmentsByDate SWR hook for per-date appointment fetching

Pratikdeb07:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a label. See existing PR titles for inspiration.
- [ ] My work is based on designs, which are linked or shown either in the Jira ticket or the description below.
- [x] My work includes tests or is validated by existing tests.

## Summary

Adds a new SWR-based data-fetching hook `useAppointmentsByDate(isoDate)` that returns full `Appointment[]` objects for a given date from `GET /appointments?forDate=`. Also extracts a shared `buildAppointmentsUrl` helper to eliminate duplication with the existing `useAppointmentList` hook — both now delegate URL construction to the same function.

## Screenshots
N/A — data layer only, no UI changes.

## Related Issue
https://openmrs.atlassian.net/browse/O3-5727

## Other

This is a foundational PR. The weekly view , daily view , and day-detail modal all will depend on this hook.

PR #2586, opened 2026-06-17
(feat) O3-5747: Build DayAppointmentsModal with service grouping and status filters

Pratikdeb07:
## Requirements

## Summary

Creates `DayAppointmentsModal` : an overlay modal showing all appointments for a selected day, grouped by service with status filter chips. This is the interaction bridge between the overview views (monthly/weekly) and the detail view (daily).

## Screenshots
EXPECTED VIEW :

Screenshot 2026-06-13 154253

## Related Issue
https://openmrs.atlassian.net/browse/O3-5747

## Other

Depends on [PR](https://github.com/openmrs/openmrs-esm-patient-management/pull/2571)

### DIFF
--- a/packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
+++ b/packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
@@ -1,0 +1,163 @@
+import React, { useMemo, useState, useEffect, useRef } from 'react';
+import { Tag, InlineLoading, Button } from '@carbon/react';
+import { Close, Launch } from '@carbon/react/icons';
+import { useTranslation } from 'react-i18next';
+import { formatAMPM } from '../../helpers/functions';
+import { type Appointment } from '../../types';
+import { useAppointmentsByDate } from '../../hooks/useAppointmentsByDate';
+import { STATUS_TAG_TYPES, DEFAULT_STATUS_TAG_TYPE, getServiceColor } from '../utils/calendar-colors';
+import styles from './day-appointments-modal.scss';
+
+const LOCALE_MAP: Record<string, string> = {
+  gregory: 'en-US',
+  ethiopic: 'am-ET',
+  islamic: 'ar-SA',
+  persian: 'fa-IR',
+};
+
+interface DayAppointmentsModalProps {
+  isoDate: string;
+  calKey: string;
+  onClose: () => void;
+  onDrillDown: (mode: 'daily', isoDate: string) => void;
+}
+
+const DayAppointmentsModal: React.FC<DayAppointmentsModalProps> = ({ isoDate, calKey, onClose, onDrillDown }) => {
+  const { t } = useTranslation();
+  const [statusFilter, setStatusFilter] = useState('All');
+  const { appointments, isLoading } = useAppointmentsByDate(isoDate);
+  const locale = LOCALE_MAP[calKey] ?? 'en-US';
+
+  const onCloseRef = useRef(onClose);
+  onCloseRef.current = onClose;
+
+  useEffect(() => {
+    const handler = (e: KeyboardEvent) => {
+      if (e.key === 'Escape') onCloseRef.current();
+    };
+    document.addEventListener('keydown', handler);
+    return () => document.removeEventListener('keydown', handler);
+  }, []);
+
+  const displayDate = useMemo(() => {
+    const d = new Date(isoDate + 'T00:00:00');
+    return new Intl.DateTimeFormat(locale, {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+      calendar: calKey,
+    }).format(d);
+  }, [isoDate, locale, calKey]);
+
+  const statuses = useMemo(() => ['All', ...Array.from(new Set(appointments.map((a) => a.status)))], [appointments]);
+
+  useEffect(() => {
+    if (statusFilter !== 'All' && !statuses.includes(statusFilter)) {
+      setStatusFilter('All');
+    }
+  }, [statuses, statusFilter]);
+
+  const filtered = useMemo(
+    () => (statusFilter === 'All' ? appointments : appointments.filter((a) => a.status === statusFilter)),
+    [appointments, statusFilter],
+  );
+
+  const byService = useMemo(() => {
+    const map = new Map<string, Appointment[]>();
+    filtered.forEach((a) => {
+      const k = a.service?.name ?? t('unknownService', 'Unknown Service');
+      map.set(k, [...(map.get(k) ?? []), a]);
+    });
+    return Array.from(map.entries());
+  }, [filtered, t]);
+
+  return (
+    <div className={styles.backdrop} onClick={onClose}>
+      <div className={styles.modal} onClick={(e) => e.stopPropagation()} role="dialog" aria-modal="true">
+        <div className={styles.header}>
+          <div>
+            <p className={styles.headerLabel}>{t('appointments', 'Appointments')}</p>
+            <h2 className={styles.headerTitle}>{displayDate}</h2>
+            <p className={styles.headerSub}>
+              {appointments.length} {t('appointmentsTotal', 'total')}
+            </p>
+          </div>
+          <div className={styles.headerActions}>
+            <Button kind="ghost" size="sm" renderIcon={Launch} onClick={() => onDrillDown('daily', isoDate)}>
+              {t('dayView', 'Day View')}
+            </Button>
+            <Button
+              hasIconOnly
+              kind="ghost"
+              size="sm"
+              renderIcon={Close}
+              iconDescription={t('close', 'Close')}
+              onClick={onClose}
+            />
+          </div>
+        </div>
+
+        {!isLoading && appointments.length > 0 && (
+          <div className={styles.filters}>
+            {statuses.map((s) => {
+              const ttype = s === 'All' ? 'gray' : (STATUS_TAG_TYPES[s] ?? DEFAULT_STATUS_TAG_TYPE);
+              const count = s === 'All' ? appointments.length : appointments.filter((a) => a.status === s).length;
+              return (
+                <Tag
+                  key={s}
+                  type={statusFilter === s ? ttype : 'gray'}
+                  tabIndex={0}
+                  onClick={() => setStatusFilter(s)}
+                  onKeyDown={(e) => {
+                    if (e.key === 'Enter' || e.key === ' ') {
+                      e.preventDefault();
+                      setStatusFilter(s);
+                    }
+                  }}
+                  className={`${styles.filterChip} ${statusFilter === s ? styles.filterChipActive : ''}`}>
+                  {s} ({count})
+                </Tag>
+              );
+            })}
+          </div>
+        )}
+
+        <div className={styles.body}>
+          {isLoading ? (
+            <InlineLoading description={t('loadingAppointments', 'Loading appointments…')} />
+          ) : filtered.length === 0 ? (
+            <p className={styles.empty}>{t('noAppointmentsFound', 'No appointments found')}</p>
+          ) : (
+            byService.map(([svcName, appts]) => (
+              <div key={svcName} className={styles.serviceGroup}>
+                <div className={styles.serviceHeader} style={{ borderBottomColor: `${getServiceColor(svcName)}40` }}>
+                  <span className={styles.serviceDot} style={{ background: getServiceColor(svcName) }} />
+                  <span className={styles.serviceName}>{svcName}</span>
+                  <span
+                    className={styles.serviceCount}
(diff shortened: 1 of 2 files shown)


Josephkagimu1 (commented):
Hi @Pratikdeb07 thanks for this task , there are few thing to look at :
 **Must Fix**
No tests — PR description claims "My work includes tests" but no test file is included. Add a day-appointments-modal.test.tsx.
Custom modal instead of Carbon's ComposedModal — The codebase uses Carbon Design System, but this rolls a completely custom backdrop + modal with hardcoded z-index: 9999. Use Carbon's ComposedModal or the O3 framework's showModal() utility to stay consistent and get free focus-trapping, scroll-lock, and accessibility.
Missing focus trap — aria-modal="true" is set but there's no focus trap. Tab will escape the modal into background content. This is an a11y violation (WCAG 2.4.3).
byService grouping is O(n²) — map.set(k, [...(map.get(k) ?? []), a]) copies the array on every insertion. Use push() instead:

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  islamic: 'ar-SA',
+  persian: 'fa-IR',
+};
+
+interface DayAppointmentsModalProps {
+  isoDate: string;
+  calKey: string;
+  onClose: () => void;
+  onDrillDown: (mode: 'daily', isoDate: string) => void;
+}
```
denniskigen, line 23: Where will the wiring live, this PR, #2636, or a follow-up? Nothing renders this component yet, and #2636 currently leaves day cell clicks unwired, so the two need to converge. I wired it locally and the flow works well, so it's close. Whichever PR does the wiring should add a test covering open, close, and the Day View drill-down.

Nit: `onDrillDown`'s `mode` argument can only ever be `'daily'`, so it can be dropped. Also, the `Launch` icon on Day View suggests opening a new window, but the action stays in place; a directional icon (or none) would read better.

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+    filtered.forEach((a) => {
+      const k = a.service?.name ?? t('unknownService', 'Unknown Service');
+      map.set(k, [...(map.get(k) ?? []), a]);
+    });
+    return Array.from(map.entries());
+  }, [filtered, t]);
+
+  return (
+    <div className={styles.backdrop} onClick={onClose}>
+      <div className={styles.modal} onClick={(e) => e.stopPropagation()} role="dialog" aria-modal="true">
```
denniskigen, line 76: @Josephkagimu1 already called this out, and it must be fixed before merge:

> Use Carbon's ComposedModal or the O3 framework's showModal() utility to stay consistent and get free focus-trapping, scroll-lock, and accessibility.

To make that concrete, this app already shows the pattern in [end-appointment.modal.tsx](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/appointments/common-components/end-appointment.modal.tsx): rename this file to `day-appointments-modal.modal.tsx`, register it under `"modals"` in `routes.json`, launch it with `showModal()`, and build the contents with Carbon's `ModalHeader`/`ModalBody` (see the [o3-docs modal guide](https://o3-docs.openmrs.org/docs/frontend-modules/modals)).

Testing locally, the current version shows exactly what this fixes: focus stays on the page after the modal opens, the dialog has no accessible name, closing doesn't return focus to the cell that opened it, and Tab escapes the modal. The conversion also deletes the backdrop div, the `stopPropagation`, the Escape listener, and the `z-index: 9999`, so it's less code, not more. (The `onCloseRef.current = onClose` write during render is also a pattern the [React docs advise against](https://react.dev/reference/react/useRef#caveats); it goes away with the conversion.)

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+          ) : (
+            byService.map(([svcName, appts]) => (
+              <div key={svcName} className={styles.serviceGroup}>
+                <div className={styles.serviceHeader} style={{ borderBottomColor: `${getServiceColor(svcName)}40` }}>
+                  <span className={styles.serviceDot} style={{ background: getServiceColor(svcName) }} />
+                  <span className={styles.serviceName}>{svcName}</span>
+                  <span
+                    className={styles.serviceCount}
+                    style={{ background: `${getServiceColor(svcName)}18`, color: getServiceColor(svcName) }}>
+                    {appts.length}
```
denniskigen, line 139: I tested this locally and the service colors don't render at all. Two problems:

1. `getServiceColor()` returns values like `var(--cds-teal-60)`, but Carbon doesn't define those variables, so the dot is invisible.
2. Appending `40` or `18` produces invalid CSS (`var(--cds-teal-60)40`), so the border and count background are dropped too.

The backend already solves this: each service has a `color` field configured by implementers, and it's already on the `AppointmentService` type. Could we use `a.service.color` with a visible fallback? The broken palette also affects the merged weekly and daily views, so fixing `calendar-colors.ts` itself can be a follow-up ticket.

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  );
+
+  const byService = useMemo(() => {
+    const map = new Map<string, Appointment[]>();
+    filtered.forEach((a) => {
+      const k = a.service?.name ?? t('unknownService', 'Unknown Service');
+      map.set(k, [...(map.get(k) ?? []), a]);
+    });
+    return Array.from(map.entries());
+  }, [filtered, t]);
```
denniskigen, line 72: Could we group by `service.uuid` and use the name only for display? Service names aren't unique, so two different services with the same name currently merge into one group.

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  isoDate: string;
+  calKey: string;
+  onClose: () => void;
+  onDrillDown: (mode: 'daily', isoDate: string) => void;
+}
+
+const DayAppointmentsModal: React.FC<DayAppointmentsModalProps> = ({ isoDate, calKey, onClose, onDrillDown }) => {
+  const { t } = useTranslation();
+  const [statusFilter, setStatusFilter] = useState('All');
+  const { appointments, isLoading } = useAppointmentsByDate(isoDate);
```
denniskigen, line 28: Could we handle the hook's `error`? Right now a failed request looks like an empty day: I forced a 404 on a day with four appointments and the modal showed "0 total" and "No appointments found". The framework's `ErrorState` is the standard fix, e.g. `<ErrorState error={error} headerTitle={...} />`.

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+
+  const displayDate = useMemo(() => {
+    const d = new Date(isoDate + 'T00:00:00');
+    return new Intl.DateTimeFormat(locale, {
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+      calendar: calKey,
+    }).format(d);
+  }, [isoDate, locale, calKey]);
```
denniskigen, line 50: This re-implements the framework's `formatDate`, which already picks the user's locale and calendar. As written, `LOCALE_MAP` sends every Gregorian locale to `en-US`, so a French user sees the date in English. One line does it all:

```tsx
const displayDate = useMemo(() => formatDate(new Date(isoDate + 'T00:00:00'), { time: false }), [isoDate]);
```

That also lets you delete `LOCALE_MAP` and the `calKey` prop from this component. Note that #2636 is centralizing these maps into `calendar-utils.ts`, so whichever PR lands second should reuse the shared helper.

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+  return (
+    <div className={styles.backdrop} onClick={onClose}>
+      <div className={styles.modal} onClick={(e) => e.stopPropagation()} role="dialog" aria-modal="true">
+        <div className={styles.header}>
+          <div>
+            <p className={styles.headerLabel}>{t('appointments', 'Appointments')}</p>
+            <h2 className={styles.headerTitle}>{displayDate}</h2>
+            <p className={styles.headerSub}>
+              {appointments.length} {t('appointmentsTotal', 'total')}
+            </p>
```
denniskigen, line 83: Could we translate the status values? With the app in French, the chips and rows still show `Scheduled`, `All`, and so on. The app already does this elsewhere: see `t(appointment.status)` in [appointments-table.component.tsx](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx), and the status keys are already in `en.json`. Three related fixes:

- `'All'` needs its own key.
- `{appointments.length} {t('appointmentsTotal', 'total')}` only reads correctly in English. Use one interpolated key, e.g. `t('appointmentsTotalCount', '{{count}} total', { count: appointments.length })`.
- The new keys (`dayView`, `noAppointmentsFound`, `unknownService`, `loadingAppointments`) are missing from `translations/en.json`. Run `yarn turbo run extract-translations` and commit the result.

The times could also follow the user's locale via the framework's `formatTime()` instead of `formatAMPM` (its other usages predate this PR and can be a follow-up).

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+    }).format(d);
+  }, [isoDate, locale, calKey]);
+
+  const statuses = useMemo(() => ['All', ...Array.from(new Set(appointments.map((a) => a.status)))], [appointments]);
+
+  useEffect(() => {
+    if (statusFilter !== 'All' && !statuses.includes(statusFilter)) {
+      setStatusFilter('All');
+    }
+  }, [statuses, statusFilter]);
```
denniskigen, line 58: Nit: this effect renders one frame with the stale filter, then corrects it. Deriving the value avoids that:

```tsx
const effectiveFilter = statuses.includes(statusFilter) ? statusFilter : 'All';
```

See [You Might Not Need an Effect](https://react.dev/learn/you-might-not-need-an-effect).

packages/esm-appointments-app/src/calendar/day-appointments-modal/day-appointments-modal.component.tsx
```diff
@@ -0,0 +1,163 @@
...
+                  type={statusFilter === s ? ttype : 'gray'}
+                  tabIndex={0}
+                  onClick={() => setStatusFilter(s)}
+                  onKeyDown={(e) => {
+                    if (e.key === 'Enter' || e.key === ' ') {
+                      e.preventDefault();
+                      setStatusFilter(s);
+                    }
+                  }}
+                  className={`${styles.filterChip} ${statusFilter === s ? styles.filterChipActive : ''}`}>
```
denniskigen, line 117: Could these use Carbon's `SelectableTag`? The active filter is currently shown only by color, so screen readers can't tell which one is selected. `SelectableTag` gives you `aria-pressed` and keyboard handling for free. (While here: `ttype` reads better as `tagType`, and the class join could use `classnames` like the other calendar components.)

denniskigen (changes requested):
Thanks @Pratikdeb07. The modal looks good and matches the ticket: appointments grouped by service, with status filters. I wired it up locally against #2636 and the whole flow works, clicking a day opens the modal and Day View drills into the daily view.

Four things must be fixed before merge: move to the standard O3 modal pattern (as @Josephkagimu1 already asked, see my comment on the backdrop for the concrete steps), the service colors don't render, services with the same name get merged into one group, and failed requests look like empty days. The rest of the comments are smaller.

Pratikdeb07:
Hi @denniskigen @Josephkagimu1 , thanks for your valuable review!

Unfortunately, @jayasanka-sack mentioned that this type of modal won't be used going forward. Because of that, I removed some of the related code, so parts of it no longer work with #2636 and `main`.

I already have a Carbon Design modal with most of the functionality ready in my local branch. My original plan was to merge the modal first and then wire everything together in #2636.

@jayasanka-sack will share the new modal design/approach, which will behave differently for the monthly, weekly, and daily views.

cc: @Bawanthathilan @jayasanka-sack

PR #2593, opened 2026-06-21
(fix) Remove nonexistent over-fetched properties from Appointment interface

Abhishek2005-ard:
…erface

## Requirements

## Summary

This PR resolves a TypeScript over-fetching issue where the `Appointment` interface included `voided`, `extensions`, and `teleconsultationLink` properties. 

These fields are not actually returned by the Bahmni Appointments API runtime responses. Having them present in the definition created misleading assumptions for developers consuming this type. Along with deleting them from `packages/esm-appointments-app/src/types/index.ts`, I updated all associated unit test mocks (like `appointments-form.test.tsx` and `appointment-details.test.tsx`) to remove these properties and prevent TS compiler errors.

## Screenshots

N/A

## Related Issue

N/A

## Other

N/A

### DIFF
--- a/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.test.tsx
+++ b/packages/esm-appointments-app/src/appointments/common-components/appointments-actions.test.tsx
@@ -48,9 +48,7 @@ const appointment: Appointment = {
   additionalInfo: null,
   providers: [{ uuid: '24252571-dd5a-11e6-9d9c-0242ac150002', display: 'Dr James Cook' }],
   recurring: false,
-  voided: false,
-  teleconsultationLink: null,
-  extensions: [],
+
   endDateTime: null,
   dateAppointmentScheduled: null,
 };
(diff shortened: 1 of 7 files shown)


Abhishek2005-ard:
Pls review the pr

jayasanka-sack:
@Abhishek2005-ard I'm not sure which endpoint you're referring to. From what I can see, the response already contains the values you mentioned are missing:

https://github.com/Bahmni/openmrs-module-appointments/blob/21c427d7d461508eb7519a4a4614d066b5ec2c98/omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentDefaultResponse.java#L8

Could you please provide a bit more context, such as the endpoint you're calling and an example response? That would help us better understand the issue.

Abhishek2005-ard:
@jayasanka-sack 
Thanks for pointing that out and sharing the link! 

You are completely correct that those properties exist in the backend's `AppointmentDefaultResponse`. My motivation for removing `voided`, `extensions`, and `teleconsultationLink` from the TypeScript `Appointment` interface was because they are not currently utilized anywhere within the `esm-appointments-app` frontend components.

For context, the endpoints the frontend is calling are:
- `GET /ws/rest/v1/appointment/all`
- `POST /ws/rest/v1/appointments/search`

While the backend JSON response does include these fields, I was attempting to eliminate "type over-fetching" by slimming down the frontend interface to only include the properties the UI strictly needs. 

However, if the project's convention is to have the TypeScript interfaces strictly mirror the backend's default REST response even if some fields are unused in the frontend, I'd be more than happy to revert the removal of those specific properties. Let me know what you prefer!

PR #2600, opened 2026-06-24
(feat) O3-5763: Unify the Current visit and Previous visit views

UjjawalPrabhat:
## Requirements

## Summary
Unifies the Service Queues **Current visit** and **Previous visit** tabs onto the patient chart's standard visit summary. Replaces the queue's bespoke visit widgets (vitals, notes, medications, encounter lists) with the patient chart's `visit-summary` extension, consumed via an `ExtensionSlot` — the same pattern the queue already uses for `patient-vitals-info`.

Requires the companion [patient-chart PR](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3419), which provides the `visit-summary` extension.

### Before:
image

### After:
image

## Related Issue
[O3-5763](https://openmrs.atlassian.net/browse/O3-5763)

## Other
Merge order: **chart -> queue**.

[O3-5763]: https://openmrs.atlassian.net/browse/O3-5763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
```diff
@@ -1,12 +1,17 @@
...
 
+const vitalsInfoSlot = 'service-queues-current-visit-vitals-slot';
+attach(vitalsInfoSlot, 'patient-vitals-info');
+
+const visitSummarySlot = 'service-queues-visit-summary-slot';
+attach(visitSummarySlot, 'visit-summary');
+
 // External workspaces that don't share the useVisit SWR key, so we revalidate the visit when
 // they close rather than relying on the form to mutate it.
 const VISIT_REFRESHING_WORKSPACES = new Set([serviceQueuesVisitNotesWorkspace, serviceQueuesPatientVitalsWorkspace]);
```
NethmiRodrigo, line 18: `serviceQueuesVisitNotesWorkspace` is now only referenced here (for revalidate-on-close) — nothing launches it anymore, since the deleted `visit-note.component.tsx` was its only launcher and the new `visit-summary` extension renders notes read-only. So this change quietly removes the ability to create/edit a visit note from the Current visit tab (the vitals authoring path is preserved via `launchCustomVitalsForm`). Is that intended? If so, the `serviceQueuesVisitNotesWorkspace` constant (`constants.ts:10`) and its `routes.json` registration (line 192) are now dead and could be removed, and this entry in `VISIT_REFRESHING_WORKSPACES` becomes moot. If not, we'll want to keep a note launcher.

packages/esm-service-queues-app/src/current-visit/current-visit.resource.ts
```diff
@@ -1,11 +1,9 @@
 import useSWR from 'swr';
 import { openmrsFetch, restBaseUrl, type Visit } from '@openmrs/esm-framework';
-import { type ConceptMetadata } from './hooks/useVitalsConceptMetadata';
-import { type ObsMetaInfo } from '../types/index';
 
 export function useVisit(visitUuid?: string) {
   const customRepresentation =
-    'custom:(uuid,encounters:(uuid,encounterDatetime,' +
+    'custom:(uuid,encounters:(uuid,diagnoses:(uuid,display,rank,diagnosis,certainty,voided),encounterDatetime,' +
```
NethmiRodrigo: The Past visit rep requests `form:(uuid,display)` but this Current visit rep doesn't. The shared `VisitSummary` reads `encounter.form` for its Completed-Forms tab / form detection, so that section will always be empty on the Current visit tab. Worth adding `form:(uuid,display)` here to match Past visit. (Separately: neither rep requests `form.resources`, which JSON-schema-form detection relies on — so that detection is effectively always false in both tabs.)

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat! A couple of points inline — mainly whether losing the in-queue visit-note authoring path is intended, plus a rep parity gap and a test suggestion.

packages/esm-service-queues-app/src/current-visit/current-visit-summary.test.tsx
```diff
@@ -1,13 +1,16 @@
...
 import { render, screen } from '@testing-library/react';
-import { useConfig } from '@openmrs/esm-framework';
 import { mockPastVisit } from '__mocks__';
 import { useVisit } from './current-visit.resource';
 import CurrentVisit from './current-visit-summary.component';
 
 const useVisitMock = vi.mocked(useVisit);
-vi.mocked(useConfig).mockReturnValue({ concepts: {} });
+
+vi.mock('@openmrs/esm-framework', async (importOriginal) => ({
```
NethmiRodrigo: Since `usePatient` is now mocked here, consider also mocking `ExtensionSlot` (as `past-visit.test.tsx` does) and asserting both slots mount with the expected state — including that the vitals slot receives `launchCustomVitalsForm`. That slot wiring is the main new logic in this component and is currently untested (the existing cases only assert the header text).

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
UjjawalPrabhat: @NethmiRodrigo Yes, this is intended coz we want the Current visit tab to be consistent with the patient chart, where the shared `visit-summary` renders notes read-only. Cleaned up the now-dead wiring accordingly.

That said, since this does remove in-queue note authoring, I'd like to confirm whether we'd prefer to keep a note-authoring launcher on this tab. If so, we can restore it as a follow-up. 
cc @ibacher @denniskigen

ibacher: I think we would like to retain the ability to right clinical notes in the service queues. It maybe worth discussing with the community whether this should be the ability to create a visit note or to update the patient sticky note / the notes in the wards app instead (those features didn't exist when this landed), but we do want to keep this behaviour.

UjjawalPrabhat: Thanks @ibacher. The workspace registration is back and the Current visit tab header now has a Visit note form launcher, so in-queue note authoring is preserved.

While wiring it up I found the launcher was silently broken even before this PR: the notes form reads its visit from `usePatientChartStore`, which Service Queues never populates, so visit was dropped from the payload and the backend's assignment handler picked a visit instead, notes could land on a different visit than the one on screen. openmrs/openmrs-esm-patient-chart#3509 exposes the form the way the vitals form already is and the launcher here passes `visitContext` explicitly. Verified against dev3.

Also raised the Talk thread [here](https://talk.openmrs.org/t/service-queues-which-kind-of-clinical-note-belongs-on-the-current-visit-tab/49832). This keeps today's behaviour working in the meantime.

image

UjjawalPrabhat: @NethmiRodrigo @ibacher Let's go ahead with visit notes then?...Ref [here](https://talk.openmrs.org/t/service-queues-which-kind-of-clinical-note-belongs-on-the-current-visit-tab/49832/3?u=ujjawalprabhat).

packages/esm-service-queues-app/src/config-schema.ts
NethmiRodrigo, line 1: There’s a couple of orphaned configs left:
```
generalPatientNoteConceptUuid,
historicalObsConceptUuid,
problemListConceptUuid,
visitDiagnosesConceptUuid
visitNoteEncounterTypeUuid
```

packages/esm-service-queues-app/src/current-visit/current-visit.resource.ts
NethmiRodrigo: We need the `display` for group members otherwise the app breaks for forms and obs that have group members.
Also, the encounter type’s editPrivilege and viewPrivilege wasn’t requested before which gates the overflow menu items. 
```suggestion
    'value:(uuid,display),display),value),encounterType:(uuid,display,viewPrivilege,editPrivilege),’ +
```

packages/esm-service-queues-app/src/config-schema.ts
NethmiRodrigo: Also, the matching `ConfigObject` entries, and `FormattedEncounter` in `types/index.ts`

packages/esm-service-queues-app/src/past-visit/past-visit.resource.ts
```diff
@@ -20,7 +23,9 @@ export function usePastVisits(patientUuid: string) {
   );
 
   const previousVisit = data?.data?.results
-    ?.filter((result) => dayjs(result.startDatetime).isBefore(dayjs().startOf('day')))
+    ?.filter(
```
NethmiRodrigo, line 15: Can we add a test for the resource file to test out the filter? E.g - `excludes the current visit when it started before today` and `returns the most recent past visit when it is not the current visit`

packages/esm-service-queues-app/src/past-visit/past-visit.component.tsx
```diff
@@ -1,21 +1,28 @@
...
   const { t } = useTranslation();
-  const { visits, isLoading } = usePastVisits(patientUuid);
+  const { visits, isLoading } = usePastVisits(patientUuid, currentVisitUuid);
 
   if (isLoading) {
-    return <StructuredListSkeleton />;
+    return (
+      <div role="progressbar">
+        <StructuredListSkeleton />
+      </div>
```
NethmiRodrigo, line 26: In the current visit component, we passed the role directly to the skeleton
```suggestion
        <StructuredListSkeleton role=“progressbar" />
```

NethmiRodrigo (changes requested):
Thanks @UjjawalPrabhat! A couple of suggestions and a blocker. Also the app breaks when trying to edit an encounter (like editing a visit note) with ` null is not an object (evaluating 'e.groupProps.patientUuid’)`. I suspect it's because chart's encounters table opens both workspaces without group props, so they come through as null outside the chart. Can we add an e2e test so that we can ensure this doesn’t break in the future with updates to the chart?
There’s also the issue of not being able to expand certain encounters, but that should be fixed by getting the `display` property for group members.
Another thing is that the styles have gotten a bit off in the table.

NethmiRodrigo:
Here’s the screen recording of it failing when expanding (you can also see the styling issue in the table)-

https://github.com/user-attachments/assets/526b645d-d7ac-40cf-a8f5-b4fcb7160bca

packages/esm-service-queues-app/src/past-visit/past-visit.component.tsx
UjjawalPrabhat: So, applying role on `StructuredListSkeleton` actually fails typecheck; Carbon declares only `className` and `rowCount` on its props (it spreads `...rest` at runtime, so only `tsc` catches it). `DataTableSkeleton` does declare them, which is why I think it works for current visit works. We could switch Previous visit to `DataTableSkeleton` as well but that might lead to a visual change so left as it is for now.

UjjawalPrabhat:
> Also the app breaks when trying to edit an encounter (like editing a visit note) with ` null is not an object (evaluating 'e.groupProps.patientUuid’)`. I suspect it's because chart's encounters table opens both workspaces without group props, so they come through as null outside the chart. 

Thanks for testing, @NethmiRodrigo!! Your hunch was right: those errors are the group's window icons, not the form, and the `patient-chart` group is `scopePattern`, scoped to chart URLs with props only the chart populates. 

Raised openmrs/openmrs-esm-patient-chart#3569 that adds an optional `onEditEncounter` and passes a handler that launches our own workspace registrations. 

Let me know your thoughts on the approach :)

NethmiRodrigo:
So we’ve had an issue where the pre-releases hadn’t gotten published, so I’m guessing the e2e tests are failing because the patient chart version that has your fix isn’t available yet...

e2e/specs/queue-entry-visit-summary.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  });
+  queueEntryUuid = queueEntry.uuid;
+});
+
+// Void what we created, newest first — the `patient` fixture cannot delete a patient that still has visits.
+test.afterEach(async ({ api }) => {
+  await api.delete(`queue-entry/${queueEntryUuid}`);
+  await api.delete(`encounter/${encounter.uuid}`);
+  await api.delete(`visit/${activeVisit.uuid}`);
+  await api.delete(`visit/${pastVisit.uuid}`);
```
NethmiRodrigo: If the setup fails partway, this throws on encounter.uuid and the earlier visits leak, which then fails the patient fixture's teardown
```suggestion
   for (const path of [
     queueEntryUuid && `queue-entry/${queueEntryUuid}`,
     encounter?.uuid && `encounter/${encounter.uuid}`,
     activeVisit?.uuid && `visit/${activeVisit.uuid}`,
     pastVisit?.uuid && `visit/${pastVisit.uuid}`,
   ]) {
     if (path) {
       await api.delete(path);
     }
   }
```

e2e/specs/queue-entry-visit-summary.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+const encounterNote = 'Note recorded on the previous visit';
+const omrsDatetime = 'YYYY-MM-DDTHH:mm:ss.SSSZZ';
+
+let pastVisit: Visit;
+let activeVisit: Visit;
+let encounter: Encounter;
+let queueEntryUuid: string;
+
+// Posts and fails loudly — a silently rejected setup call surfaces much later as a missing table row.
+async function post(api: APIRequestContext, path: string, data: Record<string, unknown>) {
```
NethmiRodrigo: We keep API setup helpers like this in `e2e/commands` (`startVisit` already covers the active visit below, and it asserts `res.ok()` the same way). Let's move the past visit and queue entry creation there too, so the next queue spec can reuse them.

e2e/specs/queue-entry-visit-summary.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+  await test.step('And then I expand that encounter', async () => {
+    await encounters.getByRole('button', { name: /expand current row/i }).click();
+    await expect(encounters).toContainText(/text of encounter note/i);
+  });
+
+  await test.step('When I choose "Edit this encounter"', async () => {
+    await page
+      .getByRole('button', { name: /edit this encounter/i })
+      .first()
+      .click();
```
NethmiRodrigo: Can’t we scope this to `encounters` so that we don’t have to do a page-wide `first()`?
```suggestion
    await encounters.getByRole('button', { name: /edit this encounter/i }).click();
```

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
```diff
@@ -1,12 +1,19 @@
...
```
NethmiRodrigo, line 19: One issue here, deleting an encounter doesn’t refresh the current visit tab because the delete encounter triggers just a modal and not a workspace. I suppose we have to add a workaround that lets the host hand its mutate to the delete path, the way `onEditEncounter` does for edits?

packages/esm-service-queues-app/src/edit-encounter.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Builds the `onEditEncounter` handler for the shared `visit-summary` extension. It launches our own
+ * registrations of the chart's edit workspaces, since the chart's belong to its `patient-chart` workspace
+ * group, which is scoped to chart URLs. The two read the patient context from different places: the visit
+ * notes form takes it as workspace props, the form entry workspace as window props.
+ */
+export function getEditEncounterHandler({ patient, patientUuid, visit, mutateVisit }: EditEncounterContext) {
+  return (encounter: EditableEncounter, isVisitNote: boolean) => {
```
NethmiRodrigo, line 28: A click before `usePatient` resolves launches the form with `patient` undefined, the Visit note form button guards this with `disabled={!patient}`
```suggestion
  return (encounter: EditableEncounter, isVisitNote: boolean) => {
     if (!patient) {
       return;
     }
```

packages/esm-service-queues-app/src/past-visit/past-visit.component.tsx
```diff
@@ -24,15 +33,17 @@ const PastVisit: React.FC<PastVisitProps> = ({ patientUuid }) => {
         <div className={styles.header}>
           <h4 className={styles.visitType}>{visits?.visitType?.display}</h4>
           <p className={styles.date}>
-            {visits?.startDatetime
-              ? (() => {
-                  const parsedDate = parseDate(visits.startDatetime);
-                  return parsedDate && !isNaN(parsedDate.getTime()) ? formatDatetime(parsedDate) : '--';
-                })()
-              : '--'}
+            {visits?.startDatetime ? formatDatetime(parseDate(visits.startDatetime)) : '--'}
```
NethmiRodrigo: `formatDatetime` throws a `RangeError` on an invalid date, so lets keep the validity check
```suggestion
            {visits?.startDatetime && !isNaN(parseDate(visits.startDatetime).getTime())
               ? formatDatetime(parseDate(visits.startDatetime))
               : '--'}
```

packages/esm-service-queues-app/src/past-visit/past-visit.test.tsx
```diff
@@ -1,80 +1,66 @@
...
-    expect(vitalsTab).toBeInTheDocument();
-    expect(screen.getByRole('tab', { name: /notes/i })).toBeInTheDocument();
-    expect(screen.getByRole('tab', { name: /medications/i })).toBeInTheDocument();
-    await user.click(vitalsTab);
-    expect(vitalsTab).toHaveAttribute('aria-selected', 'true');
+    expect(screen.getByText(pastVisit.visitType.display)).toBeInTheDocument();
+    expect(mockExtensionSlot.mock.calls.map(([props]) => props)).toContainEqual(
+      expect.objectContaining({
+        name: 'service-queues-past-visit-summary-slot',
+        state: expect.objectContaining({ visit: pastVisit, patientUuid: mockPatient.id }),
```
NethmiRodrigo: Dropping `onEditEncounter` here would only be caught by the e2e
```suggestion
        state: expect.objectContaining({
           visit: pastVisit,
           patientUuid: mockPatient.id,
           onEditEncounter: expect.any(Function),
         }),
```

packages/esm-service-queues-app/src/past-visit/past-visit.resource.test.tsx
NethmiRodrigo, line 1: Can we add a case where every visit gets filtered out? E.g - returns nothing when the only visits are today's or the current one

packages/esm-service-queues-app/src/routes.json
```diff
@@ -210,6 +215,10 @@
     {
       "name": "service-queues-visit-note",
       "group": "service-queues-patient-context"
+    },
+    {
+      "name": "service-queues-patient-form-entry",
+      "group": "service-queues-patient-context"
```
NethmiRodrigo: The chart opens this same workspace `extra-wide`, without a width this window defaults to `narrow`
```suggestion
      "name": "service-queues-patient-form-entry",
       "group": "service-queues-patient-context",
       "width": "extra-wide",
       "canMaximize": true
```

NethmiRodrigo:
@UjjawalPrabhat there’s a misalignment in the header row and the other rows in the encounters and bills table - 
Screenshot 2026-08-24 at 6 30 49 PM

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
```diff
@@ -61,21 +69,54 @@ const CurrentVisit: React.FC<CurrentVisitProps> = ({ patientUuid, visitUuid }) =
...
+              patient,
+              visitContext: visit,
+            })
+          }>
+          {t('visitNoteForm', 'Visit note form')}
+        </Button>
       </div>
       <div className={styles.visitContainer}>
-        <CurrentVisitDetails encounters={visit.encounters} patientUuid={patientUuid} visit={visit} />
+        <ExtensionSlot name={vitalsInfoSlot} state={vitalsSlotState} />
```
denniskigen:
```suggestion
        {patient && <ExtensionSlot name={vitalsInfoSlot} state={vitalsSlotState} />}
```

`patient-vitals-info` requires `patient` and calls `shouldShowBmi(patient, ...)` before its loading fallback. With a nonzero `biometrics.bmiMinimumAge`, `getPatientAge` reads `patient.birthDate`, so this can hit the extension error boundary while `usePatient` is still loading. The launcher also forwards the same missing patient. Can we cover this in the existing patient-loading test?

UjjawalPrabhat:
> One issue here, deleting an encounter doesn’t refresh the current visit tab because the delete encounter triggers just a modal and not a workspace. I suppose we have to add a workaround that lets the host hand its mutate to the delete path, the way onEditEncounter does for edits?

> there’s a misalignment in the header row and the other rows in the encounters and bills table -

@NethmiRodrigo Addressed both [here](https://github.com/openmrs/openmrs-esm-patient-chart/pull/3595).

NethmiRodrigo:
@UjjawalPrabhat can you resolve the merge conflicts?

packages/esm-service-queues-app/src/current-visit/current-visit-summary.component.tsx
```diff
@@ -61,21 +69,58 @@ const CurrentVisit: React.FC<CurrentVisitProps> = ({ patientUuid, visitUuid }) =
...
       </div>
       <div className={styles.visitContainer}>
-        <CurrentVisitDetails encounters={visit.encounters} patientUuid={patientUuid} visit={visit} />
+        {/* The vitals extension reads `patient` during render, before its own loading fallback. */}
+        {patient && <ExtensionSlot name={vitalsInfoSlot} state={vitalsSlotState} />}
+        <ExtensionSlot
+          name={visitSummarySlot}
+          state={{
+            visit,
+            patientUuid,
```
NethmiRodrigo, line 115: Following your fix on the patient chart,
```suggestion
            patientUuid,
            patient,
```

packages/esm-service-queues-app/src/current-visit/current-visit-summary.test.tsx
```diff
@@ -32,8 +57,78 @@ describe('CurrentVisit', () => {
     expect(screen.getByText('Scheduled for today')).toBeInTheDocument();
     expect(screen.getByText('On time')).toBeInTheDocument();
   });
+
+  it('mounts the vitals and visit-summary slots and wires the vitals form launcher', async () => {
+    render(<CurrentVisit patientUuid={patientUuid} visitUuid={visitUuid} />);
+
+    expect(getSlotState(visitSummarySlotName)).toMatchObject({ visit, patientUuid });
```
NethmiRodrigo: To make sure it gets this function,
```suggestion
    expect(getSlotState(visitSummarySlotName)).toMatchObject({
       visit,
       patientUuid,
       mutateVisitContext: expect.any(Function),
     });
```

packages/esm-service-queues-app/src/past-visit/past-visit.component.tsx
```diff
@@ -1,38 +1,51 @@
...
-              : '--'}
-          </p>
+          <p className={styles.date}>{startDate && !isNaN(startDate.getTime()) ? formatDatetime(startDate) : '--'}</p>
         </div>
-        <PastVisitSummary encounters={visits.encounters} patientUuid={patientUuid} />
+        <ExtensionSlot
+          name={visitSummarySlot}
+          state={{
+            visit: visits,
+            patientUuid,
```
NethmiRodrigo, line 44:
```suggestion
            patientUuid,
            patient,
```

packages/esm-service-queues-app/src/past-visit/past-visit.test.tsx
```diff
@@ -1,80 +1,70 @@
...
-    await user.click(vitalsTab);
-    expect(vitalsTab).toHaveAttribute('aria-selected', 'true');
+    expect(screen.getByText(pastVisit.visitType.display)).toBeInTheDocument();
+    expect(mockExtensionSlot.mock.calls.map(([props]) => props)).toContainEqual(
+      expect.objectContaining({
+        name: 'service-queues-past-visit-summary-slot',
+        state: expect.objectContaining({
+          visit: pastVisit,
+          patientUuid: mockPatient.id,
+          onEditEncounter: expect.any(Function),
```
NethmiRodrigo, line 48:
```suggestion
          onEditEncounter: expect.any(Function),
          mutateVisitContext: expect.any(Function),
```

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #2603, opened 2026-06-25
(fix) Keep patient chart siderail visible when editing an appointment

denniskigen:
## Requirements

## Summary

Editing an appointment from the **Appointments** dashboard in the patient chart made the chart's right-hand siderail (Clinical forms, Orders, Notes, Tasks, Patient lists) disappear.

The row **Edit** action launched `appointments-form-workspace`, which belongs to `appointments-group` rather than the `patient-chart` group. The workspaces2 system renders only one group's action menu at a time, so this switched the active group away from `patient-chart` and dropped its siderail. Because `appointments-group` defines no window icons of its own, nothing rendered in its place.

`PatientAppointmentsDetailedSummary` already receives a `launchAppointmentForm` callback that the patient chart sets to launch the chart-scoped `patient-chart-appointments-form-workspace`, but it was only wired to the **Add** button. This threads the callback through `PatientAppointmentsTable` to `PatientAppointmentsActionMenu` so the **Edit** action uses it too. When no callback is provided (standalone appointments app), it falls back to the existing `appointments-form-workspace` behavior.

Same class of bug as openmrs/openmrs-esm-billing-app#779.

### Before

https://github.com/user-attachments/assets/3eba60ba-16f5-4ee1-9f6b-836aae7127f8

### After

https://github.com/user-attachments/assets/6934d2e6-3dcf-4f74-8edc-c31523b873fd

### DIFF
--- a/packages/esm-appointments-app/src/patient-appointments/patient-appointments-action-menu.component.tsx
+++ b/packages/esm-appointments-app/src/patient-appointments/patient-appointments-action-menu.component.tsx
@@ -8,14 +8,29 @@ import styles from './patient-appointments-action-menu.scss';
 interface appointmentsActionMenuProps {
   appointment: Appointment;
   patientUuid: string;
+
+  /**
+   * Optional callback to launch the appropriate appointments form workspace, depending
+   * on which app is using this component. If not provided, defaults to launching the
+   * standalone appointments form workspace.
+   */
+  launchAppointmentForm?(patientUuid: string, appointment?: Appointment): void;
 }
 
-export const PatientAppointmentsActionMenu = ({ appointment, patientUuid }: appointmentsActionMenuProps) => {
+export const PatientAppointmentsActionMenu = ({
+  appointment,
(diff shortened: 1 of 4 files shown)


VeronicaMuthee (approved):
@denniskigen good catch and the fix. Why are the e2e tests failing?

denniskigen:
Thanks @VeronicaMuthee! 🙏

The e2e was red because this fix has a companion change in `esm-patient-chart`. Editing an appointment from the in-chart list now routes through the chart's `launchAppointmentForm` callback, so the form opens as a window in the `patient-chart` workspace group and the side rail stays visible. That callback wasn't forwarding the appointment, so edit opened the form in create mode and the "Appointment edited" assertion timed out.

The chart-side fix is now on main and published to `next` (openmrs/openmrs-esm-patient-chart@dda5104a). So the e2e tests did a sterling job yet again!

ibacher (approved):

PR #2606, opened 2026-06-26
SL-1361: add Inborn/Outborn tag to ward card

cioan:
@chibongho , please find below a prototype that implements the generated design for the Inborn\Outborn tag. Before we refactor this PR further I have a few questions:

- Can we wire this new tag into the PIH openmrs-esm-pihemr? This seems like a PIH specific custom requirement?
- If the above is not possible, how should we go about configuring this new tag that only shows up for our implementation, or anyone else who choses to add this feature?
- Is there a better way to implement this request?

Screenshot 2026-06-26 at 9 34 29 AM

> We built a NICU Inborn/Outborn visit-attribute badge for the ward app, matching your Claude design as a top-right pill wired into the maternal ward cards.
>What I created

> packages/esm-ward-app/src/ward-patient-card/card-rows/inborn-outborn-tag-row.component.tsx
> 
>   It mirrors coded-obs-tags-row.component.tsx, but instead of fetching observations it reads the visit attribute that's already on the visit object:
> 
>   - Finds the attribute whose attributeType.uuid === '86f716fc-5e26-4eb1-9484-46370cff28f0'.
>   - If its value is truthy (true / "true") → renders a blue Inborn tag; otherwise a purple Outborn tag.
>   - If the visit has no such attribute, it renders nothing (origin is unknown — see note below).
> 
>   The blue/purple choice comes straight from your Claude design (NICU Inborn-Outborn badge.html): its .origin--inborn / .origin--outborn pills use Carbon's --tag-bg-blue/--tag-bg-purple tokens — exactly what Carbon's
>   <Tag type="blue|purple"> renders, so the colors match the design while staying consistent with how the rest of the ward card draws tags.
> 
>   I also added inborn / outborn keys to translations/en.json.
> - Hits the visit's attribute subresource — GET /ws/rest/v1/visit/{visitUuid}/attribute?v=custom:(uuid,value,attributeType:

> (uuid,display)) — which does include value, unlike the default
>   visit rep.
>   - Built on useOpenmrsFetchAll (same primitive as useObs), so it gets SWR caching, isLoading/error, and paging handling for free.
>   - Returns the matching attribute and a convenience value, plus the SWR fields. No request is made when visitUuid is falsy.
> 
>   Component — inborn-outborn-tag-row.component.tsx
> 
>   - Dropped the local visit.attributes lookup (and its now-unused interface) and instead calls useVisitAttribute(visit?.uuid, INBORN_OUTBORN_ATTRIBUTE_TYPE_UUID).
>   - isInborn = value === true (strict boolean, per the boolean datatype).
>


chibongho:
> * Can we wire this new tag into the PIH openmrs-esm-pihemr? This seems like a PIH specific custom requirement?

Yeah, in `maternal-ward-patient-card.component.tsx`, we can replace `<InbornOutbornTagRow>` with `<ExtensionSlot>` instead, and define the extension in openmrs-esm-pihemr. 

> * Is there a better way to implement this request?

I think this is mostly fine given what we have.

- I can imagine that if we generalize the component to show a configurable visit attribute, we can keep it in patient-management. That said, with how we'll need to configure + localize the display strings ("Inborn" and "Outborn") it's probably not worth it. 
- We already hard-code the elements, some of them specific to our use case, in the maternal ward patient card in `maternal-ward-patient-card.component.tsx`, so it doesn't feel clean that we still need to use an ExtensionSlot to wire in a custom component. But without the ability of us to export components from this repo for use in openmrs-esm-pihemr, and therefore the ability to define `maternal-ward-patient-card.component.tsx` in pihemr instead of here, I think this is the best we can do.

cioan:
Thanks @chibongho ! I have moved this InbornOutborn component into the esm-pihemr:
https://github.com/PIH/openmrs-esm-pihemr/pull/38

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.scss
```diff
@@ -92,6 +92,14 @@
 .wardPatientCardHeader {
   @extend .dotSeparatedChildren;
   border-top: 0;
+
+  // Reserve space for the absolutely-positioned Inborn/Outborn badge (top-right) when present,
+  // so header content never slides under it. The badge is contributed by an extension slot, so it
+  // renders nested inside the slot wrapper that immediately follows the header.
+  &:has(+ :global(.ward-inborn-outborn-badge)),
```
chibongho, line 99: This doesn't feel clean as it's referring to a class / component that's PIH specific. Maybe we can work around it by wrapping the extension with a positioned div, and apply style onto that div instead? That should also simplify the CSS [here](https://github.com/PIH/openmrs-esm-pihemr/pull/38/changes).

chibongho:
The github checks are failing because they require that the PR description follow the PR template. Conventionally, we create another tracking ticket in the O3 JIRA project for changes in this repo.

PR #2611, opened 2026-06-30
(feat) O3-5769: Combine the Move and Transition row actions into one

UjjawalPrabhat:
## Requirements

## Summary

Move and Transition were never really two operations — both opened the same shared modal and hit the same `transitionQueueEntry()` endpoint, differing only in which picker they showed, so this collapses them into one action. 

I kept move as the survivor because the action-cell fallback promotes the first overflow action to an inline button on non-"Waiting" rows, and move was already first — so existing inline-button behavior is preserved.

## Screenshots

image
image

## Related Issue

[O3-5769](https://openmrs.atlassian.net/browse/O3-5769)

## Other

[O3-5769]: https://openmrs.atlassian.net/browse/O3-5769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/modals/move-queue-entry.modal.tsx
jwnasambu: @UjjawalPrabhat Kindly the confirmation text reads "Queue entry transitioned" / "Error transitioning queue entry". Since the action is now called "Move", what do you think if we updated these to say "moved" / "Error moving queue entry" and rename the translation keys accordingly?

Also fix orphaned translation keys in other locales — transition and transitionPatient were removed from en.json but remain in ar, am, es, es_MX, en_US, he, hi, pt_BR, sq, uz_UZ, etc.  Kindly clean those up to avoid dead keys.

packages/esm-service-queues-app/src/modals/move-queue-entry.test.tsx
```diff
@@ -105,4 +105,17 @@ describe('MoveQueueEntryModal', () => {
...
+      queueEntry: mockQueueEntryAlice,
+      error: null,
+      isLoading: false,
+    });
+
+    renderWithSwr(<MoveQueueEntryModal queueEntry={mockQueueEntryAlice} closeModal={closeModal} />);
+
+    expect(screen.getByText('Service location')).toBeInTheDocument();
+    expect(screen.getByText('Status')).toBeInTheDocument();
+  });
```
jwnasambu, line 126: This test only verifies both pickers render. Kindly have you consider adding a test that changing the status triggers the submit action correctly and that disableSubmit enables the button when only priority changes?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -20,7 +20,7 @@ const columnTypes = [
 ] as const;
 type ColumnType = (typeof columnTypes)[number];
 
-const queueEntryActions = ['move', 'call', 'edit', 'transition', 'remove', 'delete', 'undo'] as const;
+const queueEntryActions = ['move', 'call', 'edit', 'remove', 'delete', 'undo'] as const;
```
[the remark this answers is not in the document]
denniskigen: I confirmed this compatibility issue and found an additional failure mode. If `transition` reaches the fallback `.find()` before a known action, `actionPropsByKey[actionKey]` is undefined and reading `showIf` throws, taking down the action cell. A null guard here would prevent the crash, but the configured capability would still disappear. Could we retain `transition` as a deprecated alias for `move`, and add a focused test using legacy action configuration? A defensive null check here would still be useful for any genuinely unknown action.

denniskigen:
The combined modal works well. One non-blocking discoverability question for the O3-5769 design: will someone who only wants to change a patient's status, such as Waiting to In Service, know to open the action labelled "Move"? The status and priority controls are clear after opening the modal, but the action label does not advertise them. Could the design consider a broader label or supporting text where the action is presented?

Desktop combined Move modal

UjjawalPrabhat:
> will someone who only wants to change a patient's status, such as Waiting to In Service, know to open the action labelled "Move"? The status and priority controls are clear after opening the modal, but the action label does not advertise them. Could the design consider a broader label or supporting text where the action is presented?

Thanks for testing it, @denniskigen. 

We've a few options:
1. Rename the row label + modal copy - e.g. "Move or update". Note "Edit queue entry" is out, we already have a separate "Edit" action; and since this label also renders as an inline row button, longer strings cost column width.
2. Keep "Move", add a line in the modal - "Change the queue, status, or priority for this patient.", but might not help discoverability from the row.
3. Or both.

or any other suggestions?

CC: @ibacher @NethmiRodrigo @VeronicaMuthee

NethmiRodrigo:
> Rename the row label + modal copy - e.g. "Move or update". Note "Edit queue entry" is out, we already have a separate "Edit" action; and since this label also renders as an inline row button, longer strings cost column width.
 
This seems like the most obvious choice to me. We can always update it to something else based on feedback

packages/esm-service-queues-app/src/modals/move-queue-entry.modal.tsx
```diff
@@ -100,9 +100,9 @@ const MoveQueueEntryModal: React.FC<MoveQueueEntryModalProps> = ({ queueEntry, c
       modalParams={{
         modalTitle: t('movePatient', 'Move {{patient}}', { patient: freshEntry.display }),
         submitButtonText: t('move', 'Move'),
-        submitSuccessTitle: t('queueEntryTransitioned', 'Queue entry transitioned'),
-        submitSuccessText: t('queueEntryTransitionedSuccessfully', 'Queue entry transitioned successfully'),
-        submitFailureTitle: t('queueEntryTransitionFailed', 'Error transitioning queue entry'),
+        submitSuccessTitle: t('queueEntryMoved', 'Queue entry moved'),
+        submitSuccessText: t('queueEntryMovedSuccessfully', 'Queue entry moved successfully'),
+        submitFailureTitle: t('queueEntryMoveFailed', 'Error moving queue entry'),
```
NethmiRodrigo: Since this is the message shown for either a move or transition action, lets go for a neutral message like


```suggestion
        submitSuccessTitle: t('queueEntryUpdated', 'Queue entry updated'),
         submitSuccessText: t('queueEntryUpdatedSuccessfully', 'Queue entry updated successfully'),
         submitFailureTitle: t('queueEntryUpdateFailed', 'Error updating queue entry'),
```

packages/esm-service-queues-app/src/queue-table/cells/queue-table-action-cell.component.tsx
```diff
@@ -208,19 +221,19 @@ function ActionOverflowMenuItem({ actionKey, queueEntry }: { actionKey: QueueEnt
 }
 
 export const queueTableActionColumn: QueueTableColumnFunction = (key, header, config: ActionsColumnConfig) => {
+  const buttons = normalizeActions(config.actions.buttons, 'actions.buttons');
+  const overflowMenu = normalizeActions(config.actions.overflowMenu, 'actions.overflowMenu');
```
NethmiRodrigo: Existing configs that have move in the buttons and transition in the overflow menu would resolve both to move now and render it twice. 

```suggestion
  const overflowMenu = normalizeActions(config.actions.overflowMenu, 'actions.overflowMenu').filter(
     (actionKey) => !buttons.includes(actionKey),
   );
```

packages/esm-service-queues-app/src/modals/move-queue-entry.test.tsx
```diff
@@ -105,4 +111,65 @@ describe('MoveQueueEntryModal', () => {
...
+    const user = userEvent.setup();
+    renderWithSwr(<MoveQueueEntryModal queueEntry={mockQueueEntryAlice} closeModal={closeModal} />);
+
+    const submitButton = screen.getByRole('button', { name: 'Move' });
+    expect(submitButton).toBeDisabled();
+
+    // Alice's current priority is "Urgent"; switching to "Non urgent" is the only change.
+    await user.click(screen.getByRole('radio', { name: mockPriorityNonUrgent.display }));
+
+    expect(submitButton).toBeEnabled();
```
NethmiRodrigo, line 174: We should assert that the new priority reaches the function too.
```suggestion
    expect(submitButton).toBeEnabled();
    
    mockTransitionQueueEntry.mockResolvedValue({ status: 200 });
     await user.click(submitButton);

     expect(mockTransitionQueueEntry).toHaveBeenCalledWith(
       expect.objectContaining({ newPriority: mockPriorityNonUrgent.uuid }),
     );
```

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat!

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #2615, opened 2026-07-02
(feat) O3-5760: Rework Service Queues dashboard to match design

UjjawalPrabhat:
## Requirements

## Summary

- The header now carries the current **location · service** with an inline **Change** link, so the old right-side Location/Service dropdowns are gone — selection moved into the Change modal.
- Default location resolves to the login location if queue-tagged, else its nearest queue-tagged ancestor, else **All**. It's recomputed only when the login location changes, so a manual pick sticks across reloads.
- "Appointments today" and the Expected appointments tab come from the **appointment-scheduling** module (the queue backend has no appointments endpoint). Location filtering is client-side since `/appointments?forDate=` takes no location param.
- Average wait time switched to `queue-entry-metrics` so it works with a service selected (the old `queue-metrics` returned an unparsed payload).

**Deferred to follow-up:** configurable location `tag` + `restrictToVisitLocationDescendants` (pending [Talk-thread](https://talk.openmrs.org/t/seeking-community-input-multi-facilities-server-instances/49590) discussion)

## Screenshots

https://github.com/user-attachments/assets/b5917b37-ff7b-467c-a12e-d790f48717bc

## Related Issue

[O3-5760](https://openmrs.atlassian.net/browse/O3-5760)

## Other

Related Design [here](https://zpl.io/gggoEBN)

[O3-5760]: https://openmrs.atlassian.net/browse/O3-5760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


UjjawalPrabhat:
@ibacher @VeronicaMuthee @NethmiRodrigo @denniskigen @chibongho @jayasanka-sack would love your thoughts here.

packages/esm-service-queues-app/src/patient-queue-header/queue-location-header-title.component.tsx
```diff
@@ -0,0 +1,79 @@
...
+// and the default is only recomputed when the login location changes.
+const LOGIN_LOCATION_STORAGE_KEY = 'queueLoginLocationUuid';
+
+const QueueLocationHeaderTitle: React.FC = () => {
+  const { t } = useTranslation();
+  const { dashboardTitle } = useConfig<ConfigObject>();
+  const { queueLocations, isLoading, error } = useQueueLocations();
+  const userSession = useSession();
+  const sessionLocationUuid = userSession?.sessionLocation?.uuid;
+  const { ancestry, isLoading: isLoadingAncestry } = useSessionLocationAncestry(sessionLocationUuid);
```
NethmiRodrigo: This line takes `ancestry` and `isLoading` from `useSessionLocationAncestry` but not its `error`. That matters: this hook walks up the location tree to auto-pick the default queue location, and it only runs once per login. If that FHIR call fails, `ancestry` comes back `[]`, so the default resolves to "All" — and because the effect still writes the "already resolved" flag to session storage, it never retries even after the network recovers. The user is silently stuck on "All" for the rest of the session.

Could we destructure `error: ancestryError`, add it to the guard on line 45, and skip writing the storage flag when resolution is degraded so it can retry?

Related: the hook's docstring says it returns the location itself when `_include:iterate` is unsupported, but the empty/error path actually returns `[]` — so that fallback doesn't happen. Either fix the comment, or (better) fall back to `[locationUuid]` when there's no data, which would also fix the "stuck on All" problem above.

packages/esm-service-queues-app/src/metrics/metrics-cards/appointments-today.extension.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { MetricsCard, MetricsCardBody, MetricsCardHeader, MetricsCardItem } from './metrics-card.component';
+import { useExpectedAppointments } from '../../hooks/useExpectedAppointments';
+import { useServiceQueuesStore } from '../../store/store';
+
+export default function AppointmentsTodayExtension() {
+  const { t } = useTranslation();
+  const { selectedQueueLocationUuid } = useServiceQueuesStore();
+  const { appointments, isLoading } = useExpectedAppointments(selectedQueueLocationUuid);
```
NethmiRodrigo: This ignores the `error` from `useExpectedAppointments`, so a failed request renders `0` — which reads as a real empty day. Could we show `--` (and `logError`) when `error` is set? The Expected appointments tab already distinguishes error from empty, so this keeps the two consistent.

packages/esm-service-queues-app/src/modals/change-queue-location/change-queue-location.modal.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+const ChangeQueueLocationModal: React.FC<ChangeQueueLocationModalProps> = ({ closeModal }) => {
+  const { t } = useTranslation();
+  const { queueLocations, isLoading, error } = useQueueLocations();
+  const { selectedQueueLocationUuid, selectedServiceUuid } = useServiceQueuesStore();
+
+  const [locationUuid, setLocationUuid] = useState(selectedQueueLocationUuid ?? ALL);
+  const [serviceUuid, setServiceUuid] = useState(selectedServiceUuid ?? ALL);
+
+  const { queues, isLoading: isLoadingQueues } = useQueues(locationUuid === ALL ? undefined : locationUuid);
```
NethmiRodrigo: `useQueues` also returns an `error` here. If it fails, the Service dropdown silently shows only "All" and submit stays enabled, so the user applies "All" without knowing anything broke. Could we surface that error like the locations error is handled?

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -36,17 +25,21 @@ function DefaultQueueTable() {
...
   const [searchTerm, setSearchTerm] = useState('');
 
+  // Waiting list shows only waiting patients; in-service ones live in the Attending cards.
   const searchCriteria = useMemo(() => {
     return {
       service: selectedServiceUuid,
       location: selectedQueueLocationUuid,
       isEnded: false,
-      status: selectedQueueStatusUuid,
+      status: defaultStatusConceptUuid,
```
NethmiRodrigo: With the status dropdown removed, the waiting list is fixed to `defaultStatusConceptUuid` and Attending shows `defaultTransitionStatus`. Entries in other/custom statuses won't appear anywhere on the dashboard. Is that intended for deployments that use additional statuses?

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat!
`useQueueStatuses.ts` and the `selectedQueueStatus*` fields + `updateSelectedQueueStatus` in `store/store.ts` look orphaned (no remaining writers/consumers).

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
```diff
@@ -8,17 +8,19 @@ import { type ConfigObject } from '../../config-schema';
 
 export default function AverageWaitTimeExtension() {
   const { t } = useTranslation();
-  const { selectedServiceUuid } = useServiceQueuesStore();
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
   const {
     concepts: { defaultStatusConceptUuid },
   } = useConfig<ConfigObject>();
-  const { waitTime } = useAverageWaitTime(selectedServiceUuid, defaultStatusConceptUuid);
+  const { waitTime } = useAverageWaitTime(selectedServiceUuid, selectedQueueLocationUuid, defaultStatusConceptUuid);
```
NethmiRodrigo: Two things here: (1) this only reads `waitTime`, not `error`/`isLoading`, so a failed `queue-entry-metric` call renders the same `--` as a genuinely empty queue and nothing is logged — same silent-failure pattern as the Appointments-today card. (2) The SWR key is now unconditional (it used to be null-guarded), so it fires even with no service/location selected. The backend accepts that, but is an unfiltered average meaningful to show? Worth confirming.

packages/esm-service-queues-app/src/hooks/useExpectedAppointments.ts
```diff
@@ -0,0 +1,28 @@
...
+export function useExpectedAppointments(locationUuid?: string) {
+  const startOfDay = dayjs().startOf('day').format('YYYY-MM-DDTHH:mm:ss.SSSZZ');
+  const url = `${restBaseUrl}/appointments?forDate=${encodeURIComponent(startOfDay)}`;
+
+  const { data, error, isLoading } = useSWR<{ data: Array<ExpectedAppointment> }, Error>(url, openmrsFetch, {
+    errorRetryCount: 2,
+  });
+
+  const appointments = (data?.data ?? [])
+    .filter((appointment) => !locationUuid || appointment.location?.uuid === locationUuid)
```
NethmiRodrigo, line 28: This filters on an exact location UUID match, so an appointment booked against a child/room location of the selected queue location gets dropped from both the count and the Expected tab. Was ancestry-aware matching considered (or is exact-match the intended assumption)? Related question: should this also exclude Cancelled/Completed/Missed appointments, or is showing all statuses for the day intended?

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,96 @@
...
+    </div>
+  );
+}
+
+function GenderIndicator({ gender }: { gender?: string }) {
+  if (!gender) {
+    return null;
+  }
+  const normalized = gender.charAt(0).toUpperCase();
+  const symbol = normalized === 'F' ? '♀' : normalized === 'M' ? '♂' : '⚧';
```
NethmiRodrigo: `GenderIndicator` hardcodes `♀/♂/⚧` and renders an untranslated value. Could we use a translated label (or the existing gender presentation used elsewhere in the app) so this is i18n-friendly?

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+function AttendingPatientCard({ queueEntry }: { queueEntry: QueueEntry }) {
+  const { t } = useTranslation();
+  const { customPatientChartUrl, priorityConfigs } = useConfig<ConfigObject>();
+  const { person } = queueEntry.patient;
+  const age = person?.birthdate ? dayjs().diff(dayjs(person.birthdate), 'years') : null;
+
+  const demographics = [
+    age != null ? t('ageYearsOld', '{{age}} years old', { age }) : null,
+    person?.birthdate ? dayjs(person.birthdate).format('DD-MM-YYYY') : null,
```
NethmiRodrigo: Birthdate is formatted as a hardcoded `DD-MM-YYYY`. Consider a locale-aware format (e.g. `formatDate` with the user locale) so it matches the rest of the app.

NethmiRodrigo (commented):
A few more, lower-severity items (polish + test coverage).

Can we add unit tests for the new non-trivial logic — the default-location resolution in `QueueLocationHeaderTitle` (the sessionStorage-keyed "resolve once per login" behaviour + ancestry), `ChangeQueueLocationModal`, `useSessionLocationAncestry`, `useExpectedAppointments`, and the new Attending/Expected tab components. The resolution logic is the riskiest piece and is exactly the kind of thing worth a focused test — e.g. that a manual pick sticks across a reload and that a failed ancestry fetch doesn't permanently strand the default.

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
UjjawalPrabhat: @NethmiRodrigo So, the pre-rework the table showed all statuses via a `StatusDropdownFilter` that defaulted to "Any". The new dashboard intentionally replaces that with the fixed Waiting + Attending split following the [design](https://app.zeplin.io/project/6650717a9275db731847fb32/screen/686fc816670976b093f53ec4), so custom/other statuses no longer surface. 

For deployments using additional statuses, do we want a follow-up surface for them, or is the two-section model sufficient?

CC: @ibacher @denniskigen

packages/esm-service-queues-app/src/hooks/useExpectedAppointments.ts
UjjawalPrabhat: This is new functionality. For reference, the appointments module ([home-appointments.component.tsx](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/home/home-appointments.component.tsx)) excludes only Cancelled, and there's no existing ancestry-aware location matching to follow. Before I add filtering: (1) should we exclude terminal statuses (Cancelled/Completed/Missed) so the count reflects what's still expected, and (2) do we want descendant-location matching, or is exact-match the right assumption? @ibacher @NethmiRodrigo @denniskigen

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
UjjawalPrabhat: @NethmiRodrigo Fixed (1).

On (2): So is it even possible to achieve the "nothing selected" state coz the default resolves to a location or to "All" (null service + null location), and the metrics should reflect whatever the current screen is scoped to. When it's "All", the all-queues average is the meaningful number for that view, and the backend supports it. 

What do you think?

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
ibacher: > Is that intended for deployments that use additional statuses?

I'm sort of fine with saying the service-queues model supports fewer states than the queues module might...

packages/esm-service-queues-app/src/expected-appointments/expected-appointments.component.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useExpectedAppointments } from '../hooks/useExpectedAppointments';
+import { useServiceQueuesStore } from '../store/store';
+import { type ConfigObject } from '../config-schema';
+import styles from './expected-appointments.scss';
+
+const ExpectedAppointments: React.FC = () => {
+  const { t } = useTranslation();
+  const { customPatientChartUrl } = useConfig<ConfigObject>();
+  const { selectedQueueLocationUuid } = useServiceQueuesStore();
+  const { appointments, isLoading, error } = useExpectedAppointments(selectedQueueLocationUuid);
```
ibacher, line 25: routes.json needs to be updated to declare the dependency on the appointments module...

packages/esm-service-queues-app/src/expected-appointments/expected-appointments.component.tsx
```diff
@@ -0,0 +1,114 @@
...
+  );
+
+  const rows = useMemo(
+    () =>
+      appointments.map((appointment) => ({
+        id: appointment.uuid,
+        patientUuid: appointment.patient.uuid,
+        name: appointment.patient.name,
+        service: appointment.service?.name ?? '--',
+        time: appointment.startDateTime ? dayjs(appointment.startDateTime).format('hh:mm A') : '--',
```
ibacher: Generally, do not define custom date formats. Use the frameworks date formatting tools (this assumes 12-hour time which isn't universal).

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
ibacher: I'd probably just leave out these symbols. I don't think we use them anywhere else.

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,105 @@
...
+function GenderIndicator({ gender }: { gender?: string }) {
+  const { t } = useTranslation();
+  if (!gender) {
+    return null;
+  }
+  const normalized = gender.charAt(0).toUpperCase();
+  const symbol = normalized === 'F' ? '♀' : normalized === 'M' ? '♂' : '⚧';
+  const label =
+    normalized === 'F'
+      ? t('female', 'Female')
```
ibacher: Use the core translations for these.

ibacher (dismissed):

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
ibacher: I tend to think the average wait time across all queues and locations is a silly thing to display, but I suppose people will want to look at it, so we may as well keep it this way. It would be nice if the `null` values weren't directly passed to the backend though.

packages/esm-service-queues-app/src/hooks/useExpectedAppointments.ts
ibacher: > (1) should we exclude terminal statuses (Cancelled/Completed/Missed) so the count reflects what's still expected

Yes

> (2) do we want descendant-location matching, or is exact-match the right assumption?

This is probably a better question for talk, but my suspicion is the answer is going to be something like we'll want to tag locations as "Appointment Locations" (or "Queue Locations") and we'll want to include sub-locations that are not so tagged or something. That probably sounds nonsensical, but the idea is that appointments should be counted for sub-locations unless there's a different queue managing that location.

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
```diff
@@ -8,17 +8,33 @@ import { type ConfigObject } from '../../config-schema';
...
     <MetricsCard>
-      <MetricsCardHeader title={t('averageWaitTime', 'Average wait time today')} />
+      <MetricsCardHeader title={t('avgWaitTime', 'Avg. wait time')} />
       <MetricsCardBody>
-        <MetricsCardItem label={t('minutes', 'Minutes')} value={waitTime ? waitTime.averageWaitTime : '--'} />
+        <MetricsCardItem
+          value={
+            isLoading || error || !waitTime
+              ? '--'
+              : `${Math.round(waitTime.averageWaitTime * 100) / 100} ${t('minsUnit', 'mins')}`
```
denniskigen: The backend divides by zero when no matching entries have an `endedAt` ([QueueUtils.computeAverageWaitTimeInMinutes](https://github.com/openmrs/openmrs-module-queue/blob/main/api/src/main/java/org/openmrs/module/queue/utils/QueueUtils.java#L60)) and returns `{"averageWaitTime":"NaN"}`, so any location or service with no completed waits renders "NaN mins".

CleanShot 2026-07-17 at 23 33 14@2x

Might be worth a queue-module ticket for the backend fix, but let's guard here too:

```suggestion
            isLoading || error || !waitTime || !Number.isFinite(waitTime.averageWaitTime)
              ? '--'
              : `${Math.round(waitTime.averageWaitTime * 100) / 100} ${t('minsUnit', 'mins')}`
```

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+import styles from './attending-patients.scss';
+
+// Renders patients currently being attended (queue entries with an "In Service" status) as cards.
+const AttendingPatients: React.FC = () => {
+  const { t } = useTranslation();
+  const {
+    concepts: { defaultTransitionStatus },
+  } = useConfig<ConfigObject>();
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
+  const { queueEntries } = useQueueEntries({
```
denniskigen: Could we handle `isLoading` and `error` here? On a failed request `queueEntries` is an empty array, so the section silently disappears, indistinguishable from nobody being attended to.

packages/esm-service-queues-app/src/expected-appointments/expected-appointments.component.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  const rows = useMemo(
+    () =>
+      appointments.map((appointment) => ({
+        id: appointment.uuid,
+        patientUuid: appointment.patient.uuid,
+        name: appointment.patient.name,
+        service: appointment.service?.name ?? '--',
+        time: appointment.startDateTime ? formatTime(new Date(appointment.startDateTime)) : '--',
+        status: appointment.status,
```
denniskigen: The appointments app translates these ([appointments-table.component.tsx](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/appointments/common-components/appointments-table.component.tsx#L121)):
```suggestion
        status: t(appointment.status),
```
(`t` also needs adding to the memo deps.)

packages/esm-service-queues-app/src/home.component.tsx
```diff
@@ -1,14 +1,54 @@
...
+  } = useConfig<ConfigObject>();
+  const { totalCount } = useQueueEntries({
+    service: selectedServiceUuid,
+    location: selectedQueueLocationUuid,
+    status: defaultStatusConceptUuid,
+    isEnded: false,
+  });
+  const waitingListLabel = isNaN(totalCount)
+    ? t('waitingList', 'Waiting list')
+    : `${t('waitingList', 'Waiting list')} (${totalCount})`;
```
denniskigen: Nit: use one key so translators control the whole string:
```suggestion
  const waitingListLabel = isNaN(totalCount)
    ? t('waitingList', 'Waiting list')
    : t('waitingListWithCount', 'Waiting list ({{count}})', { count: totalCount });
```

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,106 @@
...
+      </div>
+    </div>
+  );
+};
+
+function AttendingPatientCard({ queueEntry }: { queueEntry: QueueEntry }) {
+  const { t } = useTranslation();
+  const { customPatientChartUrl, priorityConfigs } = useConfig<ConfigObject>();
+  const { person } = queueEntry.patient;
+  const age = person?.birthdate ? dayjs().diff(dayjs(person.birthdate), 'years') : null;
```
denniskigen: Nit: the framework's `age()` helper handles infants and localization; this renders "0 years old" for a baby.

denniskigen (commented):
Thanks for working through several iterations of this, @UjjawalPrabhat. I've left some comments about a bug I found while testing this locally, a question on the wait-time thread, and a few pernickety bits.

A few questions arose from reviewing the Zeplin design as well:

1. The design header shows three segments (ART Clinic · Wellbody Clinic · Triage) but the PR implements two (location · service). Which is intended?
2. The design shows two Attending cards; the implementation renders every matching entry, unbounded. Do we want a cap with a "View all", and is there an intended ordering?
3. Should the Attending section hide when empty (current behavior) or say explicitly that nobody is being attended?
4. Are we comfortable showing full name, age, and birthdate on the Attending cards, or should that be configurable?

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
denniskigen: I don't think any nulls reach the backend. The hook omits unset params, so the unfiltered case sends just `?metric=averageWaitTime&status=<uuid>`. Did you mean we should skip the request entirely when nothing is selected?

Related: the query has no date bound, so this is the all-time average of completed waits. Dev3 currently shows "1206 mins" here. If we want "today", the endpoint supports `startedOnOrAfter`. Either is fine, but we should pick one deliberately.

ibacher: We should definitely be time-bound to today for these metrics, especially as the service queues are outpatient management. (And, yeah, I didn't actually dig through the code to answer my own question).

ibacher:
My stab at answering these (as always, feel free to disagree with me)

> The design header shows three segments (ART Clinic · Wellbody Clinic · Triage) but the PR implements two (location · service). Which is intended?

There's some intention in the designs for a "hospital name" (which is what "Wellbody Clinic" is in the designs), but I don't think we've worked out how to capture that (something to discuss on Talk).

> The design shows two Attending cards; the implementation renders every matching entry, unbounded. Do we want a cap with a "View all", and is there an intended ordering?

We should definitely have a cap. Maybe 2-3 with a view all. Ideally, they'd be ordered in a way that reflects where they are being seen (i.e., Room 1 is always first), but I don't think we have a fixed concept for "Room 1".

> Should the Attending section hide when empty (current behavior) or say explicitly that nobody is being attended?

I think we should be explicit and say "No one is being attended"

> Are we comfortable showing full name, age, and birthdate on the Attending cards, or should that be configurable?

For these attending cards (i.e., which are facing the clerk), it's probably ok. For things projected onto a second screen, we would not want that (but I think that's a different problem altogether).

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
ibacher: Hmmm... We should also fix this on the backend I think, to return nothing or `null` rather than `NaN`.

UjjawalPrabhat: Raised the backend PR [here](https://github.com/openmrs/openmrs-module-queue/pull/120).

NethmiRodrigo:
@UjjawalPrabhat could you resolve the conflicts on this PR? And can you also update the screenshots in the description if you’ve addressed the review feedback 😁

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+  getCoreTranslation,
+  parseDate,
+  PatientPhoto,
+  useConfig,
+} from '@openmrs/esm-framework';
+import { useQueueEntries } from '../hooks/useQueueEntries';
+import { useServiceQueuesStore } from '../store/store';
+import QueuePriority from '../queue-table/components/queue-priority.component';
+import { type ConfigObject } from '../config-schema';
+import { type QueueEntry } from '../types';
```
NethmiRodrigo, line 19: Minor nit - components before hooks
```suggestion
import QueuePriority from '../queue-table/components/queue-priority.component';
import { useQueueEntries } from '../hooks/useQueueEntries';
import { useServiceQueuesStore } from '../store/store';
import { type ConfigObject } from '../config-schema';
import { type QueueEntry } from '../types';
```

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -0,0 +1,131 @@
...
+import styles from './attending-patients.scss';
+
+const collapsedCardCount = 3;
+
+// Renders patients currently being attended (queue entries with an "In Service" status) as cards.
+const AttendingPatients: React.FC = () => {
+  const { t } = useTranslation();
+  const {
+    concepts: { defaultTransitionStatus },
```
NethmiRodrigo, line 28: Not related to your change but I’m not confident if this is the config we should be using. But, mostly because of the naming of the config itself. Although its description does say "The UUID of the default status for attending a service in the queues eg In Service.”, I feel like it would read better if the config itself was named to `defaultAttendingStatusConceptUuid`. 
Also, not something needed in this PR, maybe a follow up?

packages/esm-service-queues-app/src/hooks/useExpectedAppointments.ts
```diff
@@ -0,0 +1,33 @@
...
+  startDateTime: number | null;
+  status: string;
+}
+
+// Terminal statuses are dropped so the count reflects what is still expected rather than everything that
+// was ever booked for the day.
+const excludedStatuses = new Set(['cancelled', 'completed', 'missed']);
+
+export function useExpectedAppointments(locationUuid?: string) {
+  const startOfDay = dayjs().startOf('day').format('YYYY-MM-DDTHH:mm:ss.SSSZZ');
```
NethmiRodrigo: The [backend doesn’t actually parse](https://github.com/Bahmni/openmrs-module-appointments/blob/master/api/src/main/java/org/openmrs/module/appointments/util/DateUtil.java#L37-L44) the `ZZ` part, so we should send an ISO string here
```suggestion
  const startOfDay = dayjs().startOf('day').toISOString();
```

packages/esm-service-queues-app/src/hooks/useExpectedAppointments.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      data: { data: appointments },
+      error: undefined,
+      isLoading: false,
+    } as ReturnType<typeof useSWR>);
+  });
+
+  it("requests today's appointments via the appointment-scheduling endpoint", () => {
+    renderHook(() => useExpectedAppointments('loc-1'));
+    const url = mockUseSWR.mock.calls[0][0] as string;
+    expect(url).toContain('/appointments?forDate=');
```
NethmiRodrigo: We should assert that the value is correct here, and not just the path
```suggestion
     const url = decodeURIComponent(mockUseSWR.mock.calls[0][0] as string);
     expect(url).toContain(`forDate=${dayjs().startOf('day').toISOString()}`);
```

packages/esm-service-queues-app/src/home.component.tsx
```diff
@@ -1,14 +1,54 @@
...
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
+  const {
+    concepts: { defaultStatusConceptUuid },
+  } = useConfig<ConfigObject>();
+  const { totalCount } = useQueueEntries({
+    service: selectedServiceUuid,
+    location: selectedQueueLocationUuid,
+    status: defaultStatusConceptUuid,
+    isEnded: false,
+  });
```
NethmiRodrigo: Since we only want the count, why aren’t we using `useQueueEntriesMetrics` here?

```suggestion
  const { count: totalCount } = useQueueEntriesMetrics({
     service: selectedServiceUuid,
     location: selectedQueueLocationUuid,
     status: defaultStatusConceptUuid,
     isEnded: false,
   });
```

packages/esm-service-queues-app/src/home.component.tsx
```diff
@@ -1,14 +1,54 @@
...
+import { useQueueEntries } from './hooks/useQueueEntries';
+import { useServiceQueuesStore } from './store/store';
+import { type ConfigObject } from './config-schema';
+import styles from './home.scss';
 
 const Home: React.FC = () => {
+  const { t } = useTranslation();
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
+  const {
+    concepts: { defaultStatusConceptUuid },
```
NethmiRodrigo: This concept, on the other hand, we probably shouldn’t use here. This is the concept UUID for the default status, and the default status can be anything, we don’t state anywhere that this should be the waiting status. But in this component we explicitly show for the waiting status. Since what status we render here isn’t configurable, we should introduce a new config to configure the waiting status concept UUID and use that here instead.

packages/esm-service-queues-app/src/patient-queue-header/queue-location-header-title.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+  }, []);
+
+  if (error) {
+    return (
+      <InlineNotification kind="error" title={t('failedToLoadLocations', 'Failed to load locations')} hideCloseButton />
+    );
+  }
+
+  const locationLabel = selectedQueueLocationName ?? t(dashboardTitle.key, dashboardTitle.value);
+  const showService = Boolean(selectedQueueLocationName && selectedServiceUuid);
```
NethmiRodrigo: The modal lets you submit location "All" with a service picked.
```suggestion
  const showService = Boolean(selectedServiceUuid);
```

packages/esm-service-queues-app/src/patient-queue-header/queue-location-header-title.component.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    // Resolve the default only once per login location.
+    if (getValueFromSessionStorage(LOGIN_LOCATION_STORAGE_KEY) === sessionLocationUuid) {
+      return;
+    }
+
+    updateValueInSessionStorage(LOGIN_LOCATION_STORAGE_KEY, sessionLocationUuid);
+    updateSelectedQueueLocationUuid(resolvedDefault.uuid);
+    updateSelectedQueueLocationName(resolvedDefault.name);
```
NethmiRodrigo, line 57: We should reset the selected service here as well, right?

```suggestion
    updateSelectedQueueLocationName(resolvedDefault.name);
    updateSelectedService(null, t('all', 'All'));
```

packages/esm-service-queues-app/src/metrics/metrics-cards/waiting-patients.extension.tsx
```diff
@@ -1,79 +1,34 @@
...
-type ServiceListItem = Service | Concept;
+import { useServiceQueuesStore } from '../../store/store';
 
 export default function WaitingPatientsExtension() {
   const { t } = useTranslation();
-  const layout = useLayoutType();
-  const { selectedServiceUuid, selectedServiceDisplay, selectedQueueLocationUuid } = useServiceQueuesStore();
-  const { services } = useQueueServices();
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
   const { serviceCount } = useServiceMetricsCount(selectedServiceUuid, selectedQueueLocationUuid);
```
NethmiRodrigo: This function `useServiceMetricsCount` hardcodes `status=Waiting` as a concept name

packages/esm-service-queues-app/src/metrics/metrics-cards/waiting-patients.extension.tsx
```diff
@@ -1,79 +1,34 @@
...
-
-  const handleServiceChange = ({ selectedItem }) => {
-    updateSelectedService(selectedItem.uuid, selectedItem.display);
-    if (selectedItem.uuid === undefined) {
-      setInitialSelectItem(true);
-    } else {
-      setInitialSelectItem(false);
-    }
-  };
+  const waitingCount = selectedServiceUuid ? serviceCount : isNaN(totalCount) ? '--' : totalCount;
```
NethmiRodrigo: Since `totalCount` is already filtered by status and because of the comment ^ about the hook hardcoding the waiting status concept name, 
```suggestion
  const waitingCount = isNaN(totalCount) ? '--' : totalCount;
```

packages/esm-service-queues-app/src/patient-queue-header/patient-queue-header.scss
NethmiRodrigo: This is no longer used

packages/esm-service-queues-app/src/expected-appointments/expected-appointments.component.tsx
```diff
@@ -0,0 +1,114 @@
...
+                    <TableHeader key={key} {...headerProps}>
+                      {header.header}
+                    </TableHeader>
+                  );
+                })}
+              </TableRow>
+            </TableHead>
+            <TableBody>
+              {dataRows.map((row) => {
+                const appointmentRow = rows.find((r) => r.id === row.id);
```
NethmiRodrigo: It’s always better to do things like this outside of the render function because otherwise this happens everytime you render.

packages/esm-service-queues-app/src/modals/change-queue-location/change-queue-location.modal.tsx
```diff
@@ -0,0 +1,163 @@
...
+      console.error('Failed to load queues for the change-queue-location modal: ', queuesError);
+    }
+  }, [queuesError]);
+
+  const services = useMemo(
+    () =>
+      uniqBy(
+        queues.flatMap((queue) => queue.service),
+        (service) => service?.uuid,
+      ).sort((a, b) => a.display.localeCompare(b.display, getLocale())),
```
NethmiRodrigo: We have a nullish operator on the line above `service?.uuid` shouldn’t we do it for `a?.display` and `b?.display`?

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -228,6 +228,14 @@ export const configSchema = {
       _type: Type.ConceptUuid,
       _default: '159947AAAAAAAAAAAAAAAAAAAAAAAAAAAAAA',
     },
+    waitingStatusConceptUuid: {
+      _type: Type.ConceptUuid,
+      _default: '51ae5e4d-b72b-4912-bf31-a17efb690aeb',
+      _description:
+        'The UUID of the status for patients waiting to be attended to, eg Waiting. The dashboard waiting list, ' +
+        'its tab count and the "Waiting" metric are all scoped to this status, so it is configured separately from ' +
+        'defaultStatusConceptUuid (which is only the status a newly created queue entry starts in).',
```
NethmiRodrigo: We don’t need to explain here its difference from `defaultStatusConceptUuid`.

```suggestion
        'The concept UUID of the status for patients waiting to be attended to.',
```

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat! @ibacher is it okay if we merge this now, or should we resolve the whole location tagging issue prior

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
```diff
@@ -8,17 +8,37 @@ import { type ConfigObject } from '../../config-schema';
...
+  const { selectedServiceUuid, selectedQueueLocationUuid } = useServiceQueuesStore();
   const {
     concepts: { defaultStatusConceptUuid },
   } = useConfig<ConfigObject>();
-  const { waitTime } = useAverageWaitTime(selectedServiceUuid, defaultStatusConceptUuid);
+  const { waitTime, isLoading, error } = useAverageWaitTime(
+    selectedServiceUuid,
+    selectedQueueLocationUuid,
+    defaultStatusConceptUuid,
+  );
```
denniskigen: Could this card read `waitingStatusConceptUuid` too? The backend filters the entries it averages by the supplied status (and ending an entry only sets `endedAt`, it doesn't change the status), so with the new config the waiting list, tab count, and Waiting metric can target one status while this card averages another. Switching it keeps all the waiting-related views aligned.

```suggestion
    concepts: { waitingStatusConceptUuid },
  } = useConfig<ConfigObject>();
  const { waitTime, isLoading, error } = useAverageWaitTime(
    selectedServiceUuid,
    selectedQueueLocationUuid,
    waitingStatusConceptUuid,
  );
```

NethmiRodrigo:
@VeronicaMuthee pinging you to get feedback on the design

NethmiRodrigo:
So let’s not implement the appointments related features for now, because - 
1. To show the appointments - the appointment location has to exactly match the selected queue location (even if the appointment location is a child of the queue location)
2. As an alternative, if we show appointments of all locations, that is potentially a security(? for the lack of a better word) issue

TLDR - for this PR, lets remove the expected appointments tag and the appointments metric card

NethmiRodrigo:
Also @UjjawalPrabhat, for the metric card “Average wait time”, lets process the time the same way we do for the column “Wait time” in the table because `3 hours 59 min` is more readable than `239.25 mins` 😁

UjjawalPrabhat:
@NethmiRodrigo Applied your suggestions. Let me know your thoughts!!

image

NethmiRodrigo:
@UjjawalPrabhat based on your screenshot, the Avg. wait time shows up as — instead of the actual value

NethmiRodrigo:
> @UjjawalPrabhat based on your screenshot, the Avg. wait time shows up as — instead of the actual value

Aaah, nvm I just remembered that you fixed it in the backend queue module, and we need to update the version of the queue module in the [distro’s pom.xml](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/a09ff7a0136dee61f3d2281e8a7bd26b175858c9/distro/pom.xml#L38) to have the latest version of it (that has your fix) in dev3. Mind opening a PR for that @UjjawalPrabhat? You’ll have to update it to the [current version the queue module is on](https://github.com/openmrs/openmrs-module-queue/blob/04895c372518b82e972f988373950ad65baf6f19/pom.xml#L12)

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+const mockUseConfig = vi.mocked(useConfig<ConfigObject>);
+const mockUseAverageWaitTime = vi.mocked(useAverageWaitTime);
+const mockFormatDuration = vi.mocked(formatDuration);
+
+vi.mock('../metrics.resource', () => ({
+  useAverageWaitTime: vi.fn(),
+}));
+
+// Spies on the real formatDuration so we can assert it is never handed a non-finite value. Node's ICU
+// tolerates NaN (it returns "0 minutes"), but Intl.DurationFormat in the browser throws a RangeError.
```
NethmiRodrigo: Node 22 has no Intl.DurationFormat, the framework mock loads the formatjs polyfill here, and that throws on NaN too.
```suggestion
// Spies on the real formatDuration so we can assert it is never handed a non-finite value.
// Intl.DurationFormat rejects NaN, natively and in the formatjs polyfill the test mocks load.
```

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  it.each([
+    ['while loading', { isLoading: true }],
+    ['on error', { error: new Error('failed to fetch the average wait time') }],
+    // The queue module divides by zero when no entry has an `endedAt` and returns {"averageWaitTime": "NaN"}.
+    ['when the backend returns NaN', { waitTime: { averageWaitTime: 'NaN' as unknown as number } }],
+    ['when the backend returns no value', { waitTime: { averageWaitTime: null as unknown as number } }],
+    // Newer queue module versions return 0 instead of NaN when nothing matches the query.
+    ['when nothing matches the query', { waitTime: { averageWaitTime: 0 } }],
```
NethmiRodrigo:
```suggestion
 // Queue <= 3.0.0 divides by zero when entries match the query but none has both timestamps.
    ['when the backend returns NaN', { waitTime: { averageWaitTime: 'NaN' as unknown as number } }],
    ['when the backend returns no value', { waitTime: { averageWaitTime: null as unknown as number } }],
    // Queue <= 3.0.0 returns 0 when no entries match at all; newer versions return null in both cases.
    ['when nothing matches the query', { waitTime: { averageWaitTime: 0 } }],
```

packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx
```diff
@@ -1,25 +1,50 @@
...
+  useEffect(() => {
+    if (error) {
+      console.error('Failed to load the average wait time metric: ', error);
+    }
+  }, [error]);
+
+  // With nothing to average the queue module returns 0, or "NaN" from older versions that divide by
+  // zero. Neither is a real measurement, and a genuine average is never exactly 0, so both show "--".
+  const averageWaitTime = Number(waitTime?.averageWaitTime);
+  const hasWaitTime = waitTime?.averageWaitTime != null && Number.isFinite(averageWaitTime) && averageWaitTime > 0;
```
NethmiRodrigo: Two things here. The version story is backwards: the released queue module returns both values, 0 when no entries match and "NaN" when the matched entries have no end times, and the null from your backend fix is the newer shape. Separately, a small average like 0.4 passes the > 0 check but still renders "0 minutes" after rounding, so lets round before the check instead.


```suggestion
  // With nothing to average the queue module returns null, or 0 / "NaN" from released versions
  // (an empty match set / no matched entry with both timestamps). None of these is a real
  // measurement, so all show "--". Rounding first also keeps a sub-minute average on "--".
   const averageWaitTime = Number(waitTime?.averageWaitTime);
   const hasWaitTime =
     waitTime?.averageWaitTime != null && Number.isFinite(averageWaitTime) && Math.round(averageWaitTime) > 0;
```

UjjawalPrabhat:
> > @UjjawalPrabhat based on your screenshot, the Avg. wait time shows up as — instead of the actual value
> 
> Aaah, nvm I just remembered that you fixed it in the backend queue module, and we need to update the version of the queue module in the [distro’s pom.xml](https://github.com/openmrs/openmrs-distro-referenceapplication/blob/a09ff7a0136dee61f3d2281e8a7bd26b175858c9/distro/pom.xml#L38) to have the latest version of it (that has your fix) in dev3. Mind opening a PR for that @UjjawalPrabhat? You’ll have to update it to the [current version the queue module is on](https://github.com/openmrs/openmrs-module-queue/blob/04895c372518b82e972f988373950ad65baf6f19/pom.xml#L12)

@NethmiRodrigo I don't think this is related to the backend PR. The reason why it was showing '--' was coz nobody had *finished* waiting yet. The card averages completed waits only, the backend skips entries missing `endedAt` ([QueueUtils](https://github.com/openmrs/openmrs-module-queue/blob/04895c372518b82e972f988373950ad65baf6f19/api/src/main/java/org/openmrs/module/queue/utils/QueueUtils.java#L60-L71)) and we scope it to today's entries in the selected location/service/status ([metrics.resource.ts#L51-L62](https://github.com/openmrs/openmrs-esm-patient-management/blob/2aab632cccc6ddfb9d0aa56cf56c7cf4a9dec98d/packages/esm-service-queues-app/src/metrics/metrics.resource.ts#L51-L62)). I clicked Call/serve on one of them and it now reads `19 minutes`. 

Also, even without the version bump we're [guarding it in the frontend](https://github.com/openmrs/openmrs-esm-patient-management/blob/2aab632cccc6ddfb9d0aa56cf56c7cf4a9dec98d/packages/esm-service-queues-app/src/metrics/metrics-cards/average-wait-time.extension.tsx#L27-L31) anyway, so it should work fine. Should we still consider the bump?

image

UjjawalPrabhat:
@NethmiRodrigo I did find another bug where we were sending the window as browser-local midnight but the module parses it with a zone-less [`SimpleDateFormat`](https://github.com/openmrs/openmrs-module-queue/blob/04895c372518b82e972f988373950ad65baf6f19/api/src/main/java/org/openmrs/module/queue/utils/QueueUtils.java#L35-L42), so a UTC server read it as 08:00 my time and hid the whole morning:

```
startedOnOrAfter=2026-08-19 00:00:00  (local, +0800)  → 0.0   → "--"
startedOnOrAfter=2026-08-18 16:00:00  (UTC)           → 18.6  → "19 minutes"
```

Fixed by sending it as UTC (`dayjs().startOf('day').utc()`), since the format can't carry an offset — ISO throws a `ParseException`. Only bites when the browser is ahead of the server, so I only caught it coz I tested past midnight. :)

VeronicaMuthee:
@UjjawalPrabhat, thanks for your great work. The page hierarchy is clean and scannable, and at a glance a user can easily see "who's being seen now" as a card band and "who's next" as the table. 

Here are my 2 cents .....
- I'd argue that the most operationally useful thing currently missing from the card (Attending) is the service that the patient is attending. I would suggest replacing the birthdate line with the service. As it is currently, one cannot tell where the patient is being attended.
- Compare the third card ("Comfort Amina Zweh"): the two-line name pushes the gender label and priority tag into a different arrangement than the other two cards, so the row of cards doesn't read as a consistent set. Is this something that can be fixed so that the layout tolerates long names?

UjjawalPrabhat:
@VeronicaMuthee What do you think about this one?
image

CC: @NethmiRodrigo @ibacher @denniskigen

VeronicaMuthee:
> @VeronicaMuthee What do you think about this one? image
> 
> CC: @NethmiRodrigo @ibacher @denniskigen

@UjjawalPrabhat, how about this order in the card details:

- Row 1 — who: patient name
- Row 2 — demographics: sex · age ("Female · 32 yrs")
- Row 3 — operational state: service + urgency pill

UjjawalPrabhat:
> @UjjawalPrabhat, how about this order in the card details:
> 
> * Row 1 — who: patient name
> * Row 2 — demographics: sex · age ("Female · 32 yrs")
> * Row 3 — operational state: service + urgency pill

@VeronicaMuthee Done!!
image

VeronicaMuthee:
> > @UjjawalPrabhat, how about this order in the card details:
> > 
> > * Row 1 — who: patient name
> > * Row 2 — demographics: sex · age ("Female · 32 yrs")
> > * Row 3 — operational state: service + urgency pill
> 
> @VeronicaMuthee Done!! image

Awesome @UjjawalPrabhat, this looks great!

NethmiRodrigo:
I guess I'm so used to seeing a pill in a specific corner when in a card that the pill being in the sort of middle on the bottom looks slightly off to me.

@VeronicaMuthee @UjjawalPrabhat what if we try something like this (ignore the profile card being a circle) I mean the seperation of the service and pill  -
Screenshot 2026-08-19 at 6 30 12 PM

UjjawalPrabhat:
> I guess I'm so used to seeing a pill in a specific corner when in a card that the pill being in the sort of middle on the bottom looks slightly off to me.
> 
> @VeronicaMuthee @UjjawalPrabhat what if we try something like this (ignore the profile card being a circle) I mean the seperation of the service and pill -

@NethmiRodrigo I think it looks better as well :D
image

ibacher:
Is the triage priority something that makes sense when a patient is attending a visit? Priority is about placement in a queue. Once they're being seen, I'm not sure it makes much sense.

NethmiRodrigo:
> Is the triage priority something that makes sense when a patient is attending a visit? Priority is about placement in a queue. Once they're being seen, I'm not sure it makes much sense.

Hmm, just throwing out assumptions, is it helpful for the person managing the queue to see that an urgent patient is currently being attended?

VeronicaMuthee:
> Is the triage priority something that makes sense when a patient is attending a visit? Priority is about placement in a queue. Once they're being seen, I'm not sure it makes much sense.

Great point, 100% agree. Priority is for deciding who gets called next. Once a patient transitions to In Service, that decision has been made, and the pill, in this case, is describing a question that has already been answered. The context about the patient, even if it no longer drives queue order, is, however, still important. So, there is still value in displaying that.

packages/esm-service-queues-app/src/metrics/metrics.resource.ts
```diff
@@ -49,13 +51,21 @@ export function useActiveVisits(locationUuid?: string) {
...
 }
 
-export function useAverageWaitTime(serviceUuid: string, statusUuid: string) {
-  const apiUrl = `${restBaseUrl}/queue-metrics?queue=${serviceUuid}&status=${statusUuid}`;
+export function useAverageWaitTime(serviceUuid: string, locationUuid: string, statusUuid: string) {
+  // Service queues are an outpatient concern, so the average is scoped to entries started today rather
+  // than to the whole history of the queue. The queue module parses this with a zone-less
+  // `yyyy-MM-dd HH:mm:ss` SimpleDateFormat and throws on anything else, so the value is sent in UTC,
+  // the timezone servers run in — browser-local time would shift the window by the client's offset.
+  const startOfDay = useMemo(() => dayjs().startOf('day').utc().format('YYYY-MM-DD HH:mm:ss'), []);
```
NethmiRodrigo: @ibacher is this okay to do, as in is it okay that we format this to UTC?

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -36,19 +25,35 @@ function DefaultQueueTable() {
...
-  const { queueEntries, isLoading, error, isValidating } = useQueueEntries(searchCriteria);
+  const { queueEntries, isLoading, error, isValidating, totalCount } = useQueueEntries(searchCriteria);
+
+  const heading = (
+    <div className={styles.headerContainer}>
+      <div className={isDesktop(layout) ? styles.desktopHeading : styles.tabletHeading}>
+        <h2>
+          {isLoading || error
+            ? t('waitingList', 'Waiting list')
+            : t('waitingListWithCount', 'Waiting list ({{count}})', { count: totalCount })}
```
NethmiRodrigo: So this count would show the total count, but this wouldn’t get updated when a user does a search, maybe we want to instead use `filteredQueueEntries.length` during a search instead?

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat! Just one minor nit

packages/esm-service-queues-app/src/metrics/metrics.resource.ts
ibacher: No. We can't guess what timezone this needs to be sent in.

ibacher: So basically this needs a backend fix.

NethmiRodrigo:
>  Should we still consider the bump?

Yes, we should because we need to test the changes made to the backend module as well

packages/esm-service-queues-app/src/metrics/metrics.resource.ts
UjjawalPrabhat: @NethmiRodrigo @ibacher Raised the backend PR [here](https://github.com/openmrs/openmrs-module-queue/pull/121).

packages/esm-service-queues-app/src/routes.json
NethmiRodrigo: No need to mention this in your comment in `metrics.resource.ts`, but we do need to update this dependency here. The `-0` makes it so that the snapshot version is also valid.
```suggestion
    "queue": >=3.1.0-0 <4.0.0
```

NethmiRodrigo (commented):
Phew, that was a long PR. Thanks for seeing it through @UjjawalPrabhat!

PR #2619, opened 2026-07-07
(feat) O3-5802: Ward app add extension slot for displaying visit attributes

cioan:
## Requirements

## Summary
This PR adds a new ExtensionSlot to the maternal-ward-patient-card.component that would allow us to slot in an extension that displays the visit attribute value

## Screenshots
Screenshot 2026-07-06 at 3 50 38 PM

## Related Issue
https://openmrs.atlassian.net/browse/O3-5802


packages/esm-ward-app/src/ward-patient-card/ward-patient-card.scss
```diff
@@ -92,6 +92,22 @@
 .wardPatientCardHeader {
   @extend .dotSeparatedChildren;
   border-top: 0;
+
+  // Reserve space for the absolutely-positioned Inborn/Outborn badge (top-right) when present,
+  // so header content never slides under it. The badge is contributed by an extension slot wrapped
+  // in .wardPatientCardHeaderBadge, which immediately follows the header.
+  &:has(+ .wardPatientCardHeaderBadge div:not(:empty)) {
```
chibongho, line 99: This looks a bit complex. Maybe we can use flexbox for this? Something like. Not a blocker.
```
.headerContainer {
  display: flex;
}

<div class="headerContainer">
  <MaternalWardPatientCardHeader>  <!-- add `flex:1` style to this component --> 
  <div className={styles.admissionRequestCardHeaderBadge} />
</div>
```

chibongho (approved):
One minor comment. Looks good otherwise.

packages/esm-ward-app/src/ward-patient-card/ward-patient-card.scss
cioan: Thanks @chibongho ! I have tried to use flex but it does not look right:
Screenshot 2026-07-09 at 10 32 30 AM

Here are things I have tried:

`maternal-ward-patient-card.component.tsx — both the mother and child headers now wrap the header + badge in a .headerContainer:

  <div className={styles.headerContainer}>
    <MaternalWardPatientCardHeader {...{ wardPatient }} />
    <div className={styles.wardPatientCardHeaderBadge}>
      <ExtensionSlot name="maternal-ward-patient-card-header-slot" state={{ visit }} />
    </div>
  </div>
  
  ward-patient-card.scss:
  - Removed the padding-right reservation from .wardPatientCardHeader and the position: absolute block from .wardPatientCardHeaderBadge.
  - Added .headerContainer { display: flex; ... } with the header child getting flex: 1 so it fills the row and pushes the badge to the top-right.

  Two deliberate choices worth flagging:
  - flex: 1 is scoped to .headerContainer > .wardPatientCardHeader, not put on the shared .wardPatientCardHeader class — that class is also used by the default ward card, where the header is a direct
  flex-item of .wardPatientCard and relies on width: 100% to fill its row; a global flex: 1 would have broken that layout.
  - .headerContainer has background-color: colors.$white so the header row stays a seamless white bar behind the badge. The card itself uses $ui-02 (not white), so without this the badge area to the right
  of the flex: 1 header would have shown the card's grey background.

  The badge only takes up space when the extension slot actually renders content (:has(div:not(:empty))), so an empty slot leaves no gap.`

Unless you see an easy flex based solution that I missed, I am going to merge the PR before attempting to position the div using flex? Thanks!

PR #2622, opened 2026-07-10
(fix) Request a minimal custom representation in useConceptAnswers

pirupius:
## Requirements

## Summary
`useConceptAnswers` (used by coded person-attribute fields, obs fields, and the cause-of-death field) fetched `/concept/{uuid}` with no `v=` representation, falling back to OpenMRS's default concept representation - names, descriptions, mappings, conceptClass, datatype, version, attributes, links, etc. Every consumer of this hook only reads `answer.uuid` and `answer.display` per answer to populate a dropdown.

This adds a minimal custom representation, `v=custom:(uuid,display,answers:(uuid,display))`, requesting only the fields actually used.

Measured against a live OpenMRS instance (a concept with 4 answers, 3 names each): default representation returned 3,396 bytes; the custom representation returns 493 bytes - an 85% payload reduction. 

## Screenshots
  N/A — no UI change. Dropdown behavior and rendered options are unchanged, only the wire payload shrinks.

## Other

Verified:
- `ConceptAnswers`/`ConceptResponse` types already only require `uuid`/`display` on each answer, so no type changes were needed.
- Checked all consumers of `useConceptAnswers` (`cause-of-death.component.tsx`, `obs-field.component.tsx`, `coded-person-attribute-field.component.tsx`) none rely on fields beyond `uuid`/`display`.
- Added `field.resource.test.ts` asserting the fetch key includes the `v=` param (didn't previously exist).
- Full `field/` test suite, `tsc --noEmit`, and `eslint` all pass clean.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/field/field.resource.test.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/field/field.resource.test.ts
@@ -1,0 +1,34 @@
+import useSWRImmutable from 'swr/immutable';
+import { vi, describe, it, expect, type Mock } from 'vitest';
+import { renderHook } from '@testing-library/react';
+import { restBaseUrl } from '@openmrs/esm-framework';
+import { useConceptAnswers } from './field.resource';
+
+vi.mock('swr/immutable', () => ({
+  default: vi.fn().mockReturnValue({
+    data: undefined,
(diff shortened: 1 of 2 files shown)


chibongho (approved):
LGTM.

PR #2623, opened 2026-07-10
(test) O3-5812: Make appointments and ward e2e tests work with other distros

chibongho:
## Requirements

## Summary

This PR modifies our existing E2E tests for the appointments and ward apps so that they can work on other distros (specifically, PIH EMR). 
- Remove hard-coded UUIDs for concepts / encounter types, etc.. in favor of either querying those values via EMRAPI, or make them configurable in the .env file.
- Simplify / remove some tests steps that assumes what appointment services / inpatient wards the serve has

## Related Issue
https://issues.openmrs.org/browse/O3-5812


e2e/core/global-setup.ts
```diff
@@ -15,7 +16,19 @@ async function globalSetup() {
...
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Basic ${token}`,
+      },
+    },
+  );
+  const emrConfigurationData = await emrConfigurationResponse.json();
+  await writeFile('e2e/.emr-configuration.json', JSON.stringify(emrConfigurationData ?? {}, null, 2));
+
+  const session = await requestContext.post(`${process.env.E2E_BASE_URL}/ws/rest/v1/session`, {
```
ibacher: We should test out whether this work (it should) be we shouldn't need to send the authorization header twice. It would be good to do the login first, grab the configuration settings and then continue.

e2e/core/global-setup.ts
```diff
@@ -15,7 +16,19 @@ async function globalSetup() {
   const token = Buffer.from(`${process.env.E2E_USER_ADMIN_USERNAME}:${process.env.E2E_USER_ADMIN_PASSWORD}`).toString(
     'base64',
   );
-  await requestContext.post(`${process.env.E2E_BASE_URL}/ws/rest/v1/session`, {
+  const emrConfigurationResponse = await requestContext.get(
```
ibacher: I feel like we should have an upfront check on the necessary values and bail if the values aren't set. As-is it will result in some unclear downstream failures that will require digging through reports when we could probably fail things faster and with a cleaner message here.

ibacher (commented):
I like the general thrust of this and supporting E2E tests on alternative distributions was always a goal, so this is a great addition. A couple of tiny notes.

e2e/core/global-setup.ts
```diff
@@ -25,6 +38,14 @@ async function globalSetup() {
       Authorization: `Basic ${token}`,
     },
   });
+  const sessionData = await session.json();
+  const currentProviderUuid = sessionData?.currentProvider?.uuid;
+
+  if (!currentProviderUuid) {
+    throw new Error('Unable to determine the current provider UUID from the session response.');
+  }
+  await writeFile('e2e/.current-provider-uuid.json', JSON.stringify({ currentProviderUuid }, null, 2));
```
denniskigen: As far as I can tell, nothing reads `e2e/.current-provider-uuid.json`, and the ward specs no longer need a provider now that `encounterProviders` has been removed from these helpers. The REST session endpoint also returns no current provider when the user has either zero or multiple provider records, so this check can abort the entire suite for otherwise valid admin accounts. Could we remove this block, along with the unused `mkdir` import?

e2e/specs/cancel-admission-request.spec.ts
```diff
@@ -21,14 +21,14 @@ let provider: Provider;
 let visit: Visit;
 let wardPatient: Patient;
 
-test.beforeEach(async ({ api, page }) => {
+test.beforeEach(async ({ api, page, emrConfiguration }) => {
   await changeToWardLocation(api);
   bedtype = await generateBedType(api);
   bed = await generateRandomBed(api, bedtype);
   provider = await getProvider(api);
```
denniskigen: Looks like this spec retained its `getProvider` call after the other ward specs dropped theirs. `provider` is no longer used, so could we remove this call, the variable, and the `getProvider` and `Provider` imports?

e2e/commands/encounter-operations.ts
```diff
@@ -91,21 +87,14 @@ export const generateWardAdmissionRequest = async (
...
       patient: patientId,
       location: process.env.E2E_WARD_LOCATION_UUID,
-      encounterType: 'b2c4d5e6-7f8a-4e9b-8c1d-2e3f8e4a3b8f',
-      encounterProviders: [
-        {
-          provider: providerId,
-          encounterRole: '240b26f9-dd88-4172-823d-4a8bfeb7841f',
-        },
-      ],
+      encounterType: process.env.E2E_ADMISSION_ENCOUNTER_TYPE_UUID,
```
denniskigen: I couldn't find any callers of `createBedAssignmentEncounter` on `main` or this branch. Could we delete it instead of changing it? As written, the helper now creates an admission encounter despite its name. If it needs to stay, `emrConfiguration.bedAssignmentEncounterType` appears to be the appropriate metadata rather than `E2E_ADMISSION_ENCOUNTER_TYPE_UUID`.

e2e/core/global-setup.ts
denniskigen: Agreed. Playwright doesn't throw on HTTP error statuses, so if this request fails, the error body gets written to `.emr-configuration.json` and only surfaces later as `undefined` concept UUIDs in the obs payloads. Something like:

```ts
if (!emrConfigurationResponse.ok()) {
  throw new Error(`Failed to fetch the EMR configuration: ${emrConfigurationResponse.status()}`);
}
```

chibongho:
- Remove unused functions in encounter-operations.ts (`createBedAssignmentEncounter()` and `deleteEncounter()`)
- in global-setup.ts, authenticate first before requesting emrConfiguration, and make the test fail if the request has error
- remove writing to unused file `e2e/.current-provider-uuid.json`
- remove unused getProvider() function
  - We no longer need it, and it also assumed that the test account is "admin", which might not work in other instances / distros

e2e/core/global-setup.ts
denniskigen: Thanks, the HTTP status check is helpful. Could we also add the upfront value validation Ian requested? A successful response can still omit the ADMIT disposition or required descriptor concepts, and the three new environment variables can be unset. In those cases the tests submit `undefined` UUIDs and fail later with a less actionable error. Could global setup report the specific missing configuration keys before continuing?

chibongho: Sure. Just added the EMRAPI configuration checks.

denniskigen (approved):
LGTM. Thanks, @chibongho!

PR #2645, opened 2026-07-29
(fix) O3-5846: Make 'Admit elsewhere' create an admission request instead of a transfer request

chibongho:
## Requirements

## Summary
See: https://openmrs.atlassian.net/browse/O3-5846

The user has 3 options to handle a pending transfer request: accept the transfer, cancel the transfer, or transfer elsewhere (by making another transfer request). The user can similarly do one of those 3 things for a pending admit request. Currently "Admit elsewhere" results in a new transfer request, which is incorrect. This PR changes it so that "Admit elsewhere" creates another admit request.

Secondary change:
Clicking on "Admit elsewhere" / "Transfer elsewhere" opens the admit / transfer request workspace. That workspace has a field that asks for the admission / transfer type (configured as dispositions in EMRAPI), but it's only shown if there is more than one disposition of that type. (In the case where there is only one disposition, that one is automatically selected, and the field is hidden). This PR changes it so that an "Admit elsewhere" / "Transfer elsewhere" request automatically inherit the original request's disposition, and thus also making the field hidden. 

### Before secondary change
image

### After secondary change (same as current behavior)
image

Testing done:
Ran internal PIH E2E tests for the following scenarios:
- Create admission request for patient in HTML form, then in ward app, have them admit elsewhere to location X, then login to location X and accept admit request
- Create admission request for patient in HTML form, then in ward app, cancel the admission request
- Create admission request for patient in HTML form, then in ward app,have them admitted. Then create a transfer request for patient to location X, then login to location X and accept transfer request

## Related Issue

https://openmrs.atlassian.net/browse/O3-5846

### DIFF
--- a/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
@@ -301,7 +301,7 @@ export default function PatientAdmitOrTransferForm({
             )}
           />
         </div>
-        {!formDefaultValues.transferType && (
+        {!formDefaultValues.transferType && dispositionsOfRequestedType?.length > 1 && (
           <div className={styles.field}>
             <h2 className={styles.productiveHeading02}>
               {isTransfer ? t('transferType', 'Transfer type') : t('admitType', 'Admit type')}
@@ -317,7 +317,7 @@ export default function PatientAdmitOrTransferForm({
                     invalid={!!error?.message}
                     invalidText={error?.message}>
                     {dispositionsOfRequestedType.map((disposition) => (
-                      <RadioButton id={disposition.uuid} labelText={disposition.name} value={disposition.uuid} />
+                      <RadioButton id={disposition.uuid} labelText={disposition.name} value={disposition.conceptCode} />
                     ))}
                   </RadioButtonGroup>
                 </ResponsiveWrapper>
--- a/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
+++ b/packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
@@ -57,9 +57,13 @@ export default function PatientAdmitOrTransferForm({
   const [isSubmitting, setIsSubmitting] = useState(false);
   const { createEncounter, emrConfiguration, isLoadingEmrConfiguration, errorFetchingEmrConfiguration } =
     useCreateEncounter();
-  const dispositionsWithTypeTransfer = useMemo(
-    () => emrConfiguration?.dispositions.filter(({ type }) => type === 'TRANSFER'),
-    [emrConfiguration],
+
+  const dispositionType = wardPatient.inpatientRequest?.dispositionType ?? 'TRANSFER';
+  const isTransfer = dispositionType === 'TRANSFER';
+  // dispositions of either type TRANSFER or ADMIT, depending on the disposition type of the wardPatient
+  const dispositionsOfRequestedType = useMemo(
+    () => emrConfiguration?.dispositions.filter(({ type }) => type === dispositionType),
+    [emrConfiguration, dispositionType],
   );
   const { wardPatientGroupDetails } = useAppContext<WardViewContext>('ward-view-context') ?? {};
   const currentAdmission = wardPatientGroupDetails?.inpatientAdmissionsByPatientUuid?.get(patient?.uuid);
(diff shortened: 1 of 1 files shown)


packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
```diff
@@ -71,28 +75,34 @@ export default function PatientAdmitOrTransferForm({
...
   );
 
   type FormValues = z.infer<typeof zodSchema>;
 
   const formDefaultValues: Partial<FormValues> = useMemo(() => {
     const defaultValues: FormValues = {};
-    if (dispositionsWithTypeTransfer?.length === 1) {
-      defaultValues.transferType = dispositionsWithTypeTransfer[0].uuid;
+    if (wardPatient.inpatientRequest?.dispositionType) {
+      defaultValues.transferType = wardPatient.inpatientRequest.disposition.uuid;
```
chibongho, line 100: Previous, `defaultValues.transferType` stores the uuid of the disposition (not the concept). I changed it to store the concept uuid or the conceptCode of the disposition's concept instead. (Note that `wardPatient.inpatientRequest.disposition` is a concept, and `wardPatient.inpatientRequest.disposition.uuid` is a concept uuid)

ibacher:
> Ran internal PIH E2E tests

Any chance we can add a version of these here? 😁

ibacher (approved):

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
```diff
@@ -283,9 +301,11 @@ export default function PatientAdmitOrTransferForm({
             )}
           />
         </div>
-        {dispositionsWithTypeTransfer?.length > 1 && (
+        {!formDefaultValues.transferType && (
```
denniskigen: Could we guard this branch on `dispositionsOfRequestedType`? The `emrConfiguration` value is `undefined` while SWR is loading (and remains undefined if the fetch fails), and for a patient with no pending request `formDefaultValues.transferType` is empty, so this branch renders and the `.map` below throws:

```
TypeError: Cannot read properties of undefined (reading 'map')
```

When the EMR configuration fetch fails, that crashes the workspace instead of showing the inline error notification with a disabled Save. The old `dispositionsWithTypeTransfer?.length > 1` check was also guarding against this.

```suggestion
        {!formDefaultValues.transferType && dispositionsOfRequestedType?.length > 1 && (
```

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
```diff
@@ -296,7 +316,7 @@ export default function PatientAdmitOrTransferForm({
                     {...field}
                     invalid={!!error?.message}
                     invalidText={error?.message}>
-                    {dispositionsWithTypeTransfer.map((disposition) => (
+                    {dispositionsOfRequestedType.map((disposition) => (
                       <RadioButton id={disposition.uuid} labelText={disposition.name} value={disposition.uuid} />
```
denniskigen: Greptile's finding is half right. When the type field is shown and the user picks a radio, we now submit the raw radio value as the disposition obs value, where main maps it to the conceptCode first. That value is the emrapi disposition config uuid from dispositionConfig.json, not a concept reference, so request creation fails for any implementation with more than one disposition of a given type. I repro'd this with two TRANSFER dispositions configured: the payload contained `value: 'disposition-config-uuid-2'`. You won't see it in the RefApp or on dev3 because they only have one disposition per type, so the radio group never renders.

The inherited path is fine. `wardPatient.inpatientRequest.disposition` is a concept, and I verified it works against both a local backend and dev3.

```suggestion
                      <RadioButton id={disposition.uuid} labelText={disposition.name} value={disposition.conceptCode} />
```

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
```diff
@@ -71,28 +75,34 @@ export default function PatientAdmitOrTransferForm({
     () =>
       z.object({
         location: z.string({
-          required_error: t('pleaseSelectTransferLocation', 'Please select transfer location'),
+          required_error: isTransfer
+            ? t('pleaseSelectTransferLocation', 'Please select transfer location')
+            : t('pleaseSelectAdmitLocation', 'Please select admit location'),
```
denniskigen, line 80: The new keys (`admitType`, `pleaseSelectAdmitLocation`, `pleaseSelectAdmitType`, `patientAdmitRequestCreatedForPatient`, `errorCreatingAdmitRequest`) need to land in `en.json`, otherwise they never reach Transifex. Running `yarn turbo run extract-translations` should do it.

denniskigen (commented):
Thanks for this, @chibongho. I've left a few comments below.

packages/esm-ward-app/src/ward-workspace/patient-transfer-bed-swap/patient-admit-or-transfer-request-form.component.tsx
chibongho: Good catch. Thanks!

chibongho:
> > Ran internal PIH E2E tests
> 
> Any chance we can add a version of these here? 😁

Yeah. Just need to add a test function to create admission requests, and take out the maternal ward stuff, and the tests should be applicable to the refapp too.

PR #2646, opened 2026-07-29
(fix) O3-5846: Admit elsewhere should create an admission request instead of a transfer request

sudeshkar:
## Requirements

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (No visual redesign — the copy simply follows the request type that is already shown on the card.)

## Summary

In the ward app a pending request can be handled by accepting it, cancelling it, or sending the patient elsewhere. For a pending *transfer* request that last action is "Transfer elsewhere"; for a pending *admission* request it is "Admit elsewhere". Both were creating a **transfer** request, because the shared form always recorded a disposition of type `TRANSFER` under `dispositionDescriptor.internalTransferLocationConcept`.

The backend derives the request type from the disposition obs group — `inpatient_requests.hql` matches on the disposition concept and reads the location from either the admission or the internal-transfer location concept — so "Admit elsewhere" now records a disposition of type `ADMIT` under `admissionLocationConcept`, producing a new **admission** request instead.

`PatientAdmitOrTransferForm` takes a `dispositionType` prop that selects the disposition and the location concept, and the labels and snackbar messages follow it. It defaults to `TRANSFER`, so the "Transfers" workspace (transferring an already-admitted patient) is unchanged.

Note that emrapi exposes no admission-request encounter type, and the HQL above does not filter requests by encounter type, so the encounter is still created with `transferRequestEncounterType` as before.

## Screenshots

Same "Admit elsewhere" action, on a patient with a pending admission request, sent to Ward 2:

| Before | After |
| --- | --- |
| 02-BEFORE-transfer-request-created | 03-AFTER-admission-request-created |

Verified end to end against dev3 on both builds:

- before — `GET emrapi/inpatient/request?dispositionType=ADMIT,TRANSFER&dispositionLocation=<Ward 2>` → `dispositionType: "TRANSFER"`
- after — same call → `dispositionType: "ADMIT"`, disposition `Admit to hospital`

## Related Issue

https://openmrs.atlassian.net/browse/O3-5846

## Other

New test `patient-transfer-request.test.tsx` asserts the created encounter payload for both request types — it fails on `main` and passes with this change.


sudeshkar:
Good catch — confirmed and fixed in 154c101d.

The finding is real: `inpatientRequestsOfWardByPatientUuid` is built from `useInpatientRequest()`, which defaults to `["ADMIT", "TRANSFER"]`, so a mother awaiting admission can have a child awaiting transfer. `MaternalAdmissionRequestCard` already labels each related patient from their own `dispositionType` (`Also admit / cancel:` vs `Also transfer / cancel:`), so mixed selections are expected by design.

Rather than restricting the selection — which would silently drop the child from a joint mother/child action, defeating the point of the maternal card — each patient's disposition is now built from their own request type.

One case the report did not cover: related patients in the **Transfers** workspace are bed-mates rather than pending requests, and their `currentInpatientRequest` can be a pending **DISCHARGE** (see `ward-view.resource.ts`, which handles `TRANSFER | DISCHARGE`). Deriving each payload straight from the patient's own `dispositionType` would have turned "also transfer this person" into a discharge request. So the resolution only honours a patient's own type when it is `ADMIT` or `TRANSFER`, and otherwise falls back to the type the form was opened for:

```ts
const requestTypeOf = ({ inpatientRequest }: WardPatient) =>
  inpatientRequest?.dispositionType === "ADMIT" || inpatientRequest?.dispositionType === "TRANSFER"
    ? inpatientRequest.dispositionType
    : dispositionType;
```

This mirrors the `wp.inpatientRequest?.dispositionType ?? dispositionType` pattern already used in `admit-patient-button.component.tsx`.

Covered by a new test, `keeps each related patient on their own request type`, which submits an ADMIT primary with a TRANSFER related patient and asserts both payloads separately. I verified it fails when related patients inherit the primary type and passes with the fix.

sudeshkar:
@greptileai

sudeshkar:
Fixed in 78fee845 — good catch again.

Picking the first configured disposition of the related patient's type was indeed wrong where an implementation configures more than one disposition per type. The pending request already carries its own disposition concept (`InpatientRequest.disposition`, fetched by `useInpatientRequest`'s rep), so there is no need to guess: a related patient keeping their own request type now keeps that exact disposition, and only its location changes.

```ts
const dispositionValueOf = ({ inpatientRequest }: WardPatient, requestType) =>
  requestType === dispositionType
    ? dispositionsOfRequestedType.find(({ uuid }) => uuid === values.disposition)?.conceptCode
    : (inpatientRequest?.disposition?.uuid ??
       emrConfiguration.dispositions.find(({ type }) => type === requestType)?.conceptCode);
```

Patients on the type the form was opened for still get the disposition selected on the form.

Added `preserves the specific disposition a related pending request was created with`, which gives a related transfer patient a second, distinct transfer disposition and asserts it survives submission. Verified it fails against the previous "first configured disposition" behaviour and passes now.

@greptileai

NethmiRodrigo:
Thanks @sudeshkar! But closing this since it is a duplicate of #2645

PR #2636, opened 2026-07-23
(feat) O3-5830: Wire calendar header, refactor calendar view

Pratikdeb07:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/en-US/docs/frontend-modules/contributing/#contributing-guidelines) label. See existing PR titles for inspiration.
- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui))
- [x] My work includes tests or is validated by existing tests.

## Summary
The appointments calendar currently has a simple monthly workload grid with a single back-button header. Clicking any day cell navigates away from the calendar to a filtered list page. The individual view components (weekly time-slot grid, daily schedule, day-detail modal) have been built in prior PRs  but are not yet wired into the application.

## Screenshots
image

## Related Issue
https://openmrs.atlassian.net/browse/O3-5830

## Other
N/A

PR #2665, opened 2026-08-10
(chore) Verify generated translation catalogs

denniskigen:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) label. See existing PR titles for inspiration.

## Summary

The pre-commit hook already ran translation extraction, but it did not fail when extraction rewrote a catalog. Git could still create the commit without those generated changes, which allowed translation drift to be left behind.

This adds one `check-translations` command. It forces extraction, fails when a tracked `packages/*/translations/en.json` changes, and also fails when extraction creates a catalog that has never been committed. We're running the same command in Husky for quick local feedback and in CI through `pre_verify_command`, so bypassing the hook cannot let translation drift merge.

Running the command on current `main` removes `nextMonth`, `prev`, and `previousMonth` from the appointments catalog. Those keys were left behind by the calendar refactor in #2636. This commits that generated cleanup so the new check starts from a clean baseline.

## Other

Extraction reads the full working tree, so an unrelated unstaged translation change can block a partial commit. The old hook had the same extraction scope; the difference is that it now fails instead of silently leaving the generated changes unstaged.


chibongho (approved):
Looks fine to me in principle. A couple things:
- I think there were rare times when I had to modify the translation files manually, but I don't remember why.
- How would pluralized translation strings work? Something like this:
```
packages/esm-ward-app/translations/en.json:  "days_one": "{{count}} day",
packages/esm-ward-app/translations/en.json:  "days_other": "{{count}} days"
```

denniskigen:
Thanks, @chibongho.

This check only covers `en.json`; other locale files are unaffected. Within `en.json`, values are safe and keys aren't. `i18next-parser` [reads and merges the existing catalog](https://github.com/i18next/i18next-parser/blob/6c10a2b66ebadb8250039d078ad2a53e52753a6e/src/transform.js#L238-L255), so a value you hand-edit survives extraction. `minutes_one` in ward-app is live proof: the catalog says `{{count}} minutes` while the source default says `{{count}} minute`, and that value has survived extraction since it was introduced in #1592.

This means the check validates the generated key set and catalog shape, not whether existing values match their source defaults.

Hand-adding a key is different. If extraction runs before the key is staged, it drops the key and the check passes because the catalog has returned to its committed state. If the key is staged, the hook fails. If the hook is bypassed and the key is committed, CI fails until the key is represented in scanned source.

The existing escape hatch is a comment-form `t()` marker. We have 43 of those in scanned package source, including 18 in `esm-appointments-app/src/constants.ts`. The change that introduced those constants also added `src/constants.ts` to the appointments extraction glob for exactly that reason; no other package explicitly scans a `constants.ts` file.

Plurals are fine. There are 26 plural keys across four English catalogs, 14 of them in ward-app, and they round-trip unchanged. The parser reads `count` and the plural defaults from the options object, which is what `t('days', { count, defaultValue_one, defaultValue_other })` relies on.

One separate bug fell out of checking this: `minutes_one` currently renders “1 minutes.” I'll fix that separately.

PR #2667, opened 2026-08-11
(fix) O3-5877: Show a clear message when patient search query is below the minimum character count

solomonfortune:
## Requirements

## Summary

Patient search previously showed "Sorry, no patient charts were found" whenever a query was shorter than `minSearchCharacters`, because the query was below the configured minimum. This was misleading since it looked like a real empty result.

This PR adds a `useMinSearchCharacters` hook that fetches the configured `minSearchCharacters` global property and uses it to show a distinct message ("Please enter at least N characters to search") when the query is too short, while leaving the original "no results" message intact for genuine empty results. Applied to both the compact search and the full search page.

Both views wait for the setting to load, and the shared search hook blocks patient search requests until the trimmed query meets the configured minimum. Shortening the query below the minimum clears the previous results.

### Before
Screenshot (545)

### After
Screenshot (546)

## Related Issue
https://openmrs.atlassian.net/browse/O3-5877

### DIFF
--- a/packages/esm-patient-search-app/src/compact-patient-search/patient-search.component.tsx
+++ b/packages/esm-patient-search-app/src/compact-patient-search/patient-search.component.tsx
@@ -1,9 +1,10 @@
-import React, { forwardRef, useCallback } from 'react';
+﻿import React, { forwardRef, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Layer, Tile } from '@carbon/react';
 import { EmptyCardIllustration } from '@openmrs/esm-framework';
 import { type PatientSearchResponse } from '../types';
 import CompactPatientBanner, { type CompactPatientBannerHandle } from './compact-patient-banner.component';
+import { useMinSearchCharacters } from '../patient-search.resource';
 import Loader from './loader.component';
 import styles from './patient-search.scss';
 import { SWRConfig } from 'swr';
@@ -13,8 +14,9 @@ interface PatientSearchProps extends PatientSearchResponse {
 }
 
 const PatientSearch = forwardRef<CompactPatientBannerHandle, PatientSearchProps>(
-  ({ data: searchResults, fetchError, hasMore, isLoading, isValidating, setPage, totalResults }, ref) => {
+  ({ data: searchResults, fetchError, hasMore, isLoading, isValidating, query, setPage, totalResults }, ref) => {
     const { t } = useTranslation();
+    const { minSearchCharacters } = useMinSearchCharacters();
 
     const fetchMore = useCallback(() => setPage((page) => page + 1), [setPage]);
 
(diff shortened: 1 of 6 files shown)


solomonfortune:
@denniskigen and @ibacher. please review this at your time of convenience.

denniskigen:
Thanks, @solomonfortune, for adding a clearer message to both search views. I’ve added a requirement to [O3-5877](https://openmrs.atlassian.net/browse/O3-5877): can we also prevent patient search requests until the trimmed query meets the configured `minSearchCharacters` value?

It’s probably best to put the guard in the shared search hook so both views use the same minimum for requests and messaging. The ticket includes acceptance criteria for waiting for the setting to load, reaching the minimum, and shortening the query below it, along with test coverage.


[O3-5877]: https://openmrs.atlassian.net/browse/O3-5877?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -260,3 +260,12 @@ export function useRestPatients(
     [mappedData, isLoading, error, patientUuids, size, isValidating, setSize],
   );
 }
+
+export function useMinSearchCharacters() {
+  const url = `${restBaseUrl}/systemsetting?q=minSearchCharacters&v=custom:(property,value)`;
+  const { data, isLoading, error } = useSWR<FetchResponse<{ results: Array<{ value: string }> }>, Error>(url, fetcher);
```
denniskigen:
```suggestion
  const { data, isLoading, error } = useSWRImmutable<FetchResponse<{ results: Array<{ value: string }> }>, Error>(url, fetcher);
```

Prefer using [useSWRImmutable](https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations) here as this resource is not likely to change on the backend.

packages/esm-patient-search-app/src/compact-patient-search/patient-search.component.tsx
```diff
@@ -50,6 +64,25 @@ const PatientSearch = forwardRef<CompactPatientBannerHandle, PatientSearchProps>
       );
     }
 
+    if (!isLoadingMinSearchCharacters && query.trim().length < minSearchCharacters) {
```
denniskigen: Can we show the loading state while `isLoadingMinSearchCharacters` is true? We're still showing “Sorry, no patient charts were found” while the setting loads, before we've run a search. Can we add a test for this case too?

solomonfortune (commented):
Thanks for catching that @denniskigen. i have fixed it. We were falling through to the no-results message while the setting was still loading; now it shows the loading state instead, and I have added a test for it.

denniskigen (approved):
LGTM! Thanks, @solomonfortune.

PR #2674, opened 2026-08-12
(fix) O3-5700: Bed administration background revalidation error handling to preserve rows and clear stale errors

omeriinnocent:
## Requirements

## Summary

Fixes Bed Administration background revalidation so that transient refresh failures no longer replace loaded bed rows with a blocking error state, and stale errors clear on the next successful revalidation.

Changes:

- `summary.resource.ts` - migrated `useBedsGroupedByLocation` to SWR:
  - `keepPreviousData: true` retains previous data during background revalidation
  - per-request `AbortController`s cancel in-flight fetches when locations change or the hook unmounts
  - `revalidateOnFocus: true` re-fetches beds when the window regains focus (overrides the app-wide default from esm-core)
- `bed-administration-table.component.tsx` - updated error handling:
  - show the blocking `ErrorState` only when there are no loaded rows
  - show a non-blocking "Background fetch failed" indicator when background revalidation fails with data already loaded
  - clear the warning once a revalidation succeeds
- `bed-administration-table.scss` - added `.backgroundErrorNotification` styling
- `bed-administration-table.component.test.tsx` - added tests:
  - initial load failure shows the blocking error
  - background revalidation failure preserves rows and shows the warning
  - successful revalidation clears the warning and updates rows

## Screenshots

Test run
image

Test run for the fix
image
## Related Issue

https://openmrs.atlassian.net/browse/O3-5700


omeriinnocent:
follow up of this pr https://github.com/openmrs/openmrs-esm-patient-management/pull/2538?

omeriinnocent:
@denniskigen @chibongho @NethmiRodrigo  some of your time please.

packages/esm-bed-management-app/src/bed-administration/bed-administration-table.component.test.tsx
```diff
@@ -0,0 +1,150 @@
+import React from 'react';
+import { render, screen, waitFor } from '@testing-library/react';
+import '@testing-library/jest-dom';
+// BedAdministrationTable and summaryResource are imported after mocks to ensure mocks are applied
+
+jest.mock('../summary/summary.resource', () => ({
```
denniskigen: This repo uses vitest, so this file never runs - `jest` isn't defined and CI fails at collection ([logs](https://github.com/openmrs/openmrs-esm-patient-management/actions/runs/31595654148/job/94110393208)). Can we fold these scenarios into `bed-administration-table.test.tsx` using `vi.*`? It already covers this component.

These also mock the hook itself, so they'd pass against the old code too. Here's a hook-level test that fails on main and passes on your branch - feel free to lift it as `src/summary/summary.resource.test.tsx`:

```tsx
it('clears the error and repopulates beds after a successful revalidation', async () => {
  vi.mocked(useConfig).mockReturnValue({ admissionLocationTagName: 'Admission Location' } as any);
  let failBeds = true;
  vi.mocked(openmrsFetch).mockImplementation(async (url: string) => {
    if (url.includes('/location?tag=')) {
      return { data: { results: [{ uuid: 'loc-1', display: 'Ward A' }] } } as any;
    }
    if (failBeds) throw new Error('boom 500');
    return { data: { results: [{ id: 1, uuid: 'bed-1', bedNumber: 'B-1', status: 'AVAILABLE' }] } } as any;
  });

  const wrapper = ({ children }: { children: React.ReactNode }) =>
    React.createElement(SWRConfig, { value: { provider: () => new Map(), dedupingInterval: 0 } }, children);
  const { result } = renderHook(() => useBedsGroupedByLocation(), { wrapper });

  await waitFor(() => expect(result.current.errorFetchingBedsGroupedByLocation).toBeTruthy());

  failBeds = false;
  await act(async () => {
    await result.current.mutateBedsGroupedByLocation();
  });

  await waitFor(() => expect(result.current.errorFetchingBedsGroupedByLocation).toBeNull());
});
```

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -84,66 +84,66 @@ export const useLocationName = (locationUuid: string) => {
...
+    }
+  };
 
-  const mutate = useCallback(() => {
-    setIsValidating(true);
-  }, []);
+  const { data, error, isLoading, isValidating, mutate } = useSWR(swrKey, fetcher, {
+    keepPreviousData: true,
+    revalidateOnFocus: true,
+  });
```
denniskigen, line 112:
```suggestion
  const { data, error, isLoading, isValidating, mutate } = useSWR(swrKey, fetcher, {
    keepPreviousData: true,
  });
```
esm-core deliberately disables `revalidateOnFocus` for every app, and the same config sets a 30-minute `focusThrottleInterval` that this hook doesn't override ([openmrsComponentDecorator.tsx#L28-L36](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/src/openmrsComponentDecorator.tsx#L28-L36)) - so opting back in here buys focus revalidation at most once per half hour. That's too unreliable to be the thing that clears the warning, and it silently deviates from the app-wide default. It's probably better to drop it and let the next manual refresh (e.g. after a bed save) clear the warning instead.

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -84,66 +84,66 @@ export const useLocationName = (locationUuid: string) => {
...
-        });
-    }
     return () => {
-      isSubscribed = false;
+      // Abort any in-flight requests when unmounting or when locations change.
+      activeControllers.forEach((c) => c.abort());
+      activeControllers.clear();
     };
-  }, [admissionLocations, isLoadingAdmissionLocations, isValidating]);
+  }, [admissionLocations]);
```
denniskigen: Can we drop the AbortController machinery altogether - the ref, this effect, the `signal`, and the `catch`? SWR already ignores stale responses, so the aborts buy us nothing, and they misfire: running this branch locally, the very first load of the Bed allocation page intermittently aborts its own bed request and renders the full-page error state on a healthy backend ("Error undefined:" - the AbortError has no HTTP status). Network trace from a single page load:

```
GET /openmrs/ws/rest/v1/bed?locationUuid=... [FAILED: net::ERR_ABORTED]
GET /openmrs/ws/rest/v1/bed?locationUuid=... → 200 OK
```

Same misfire reproducible in a hook test (rename a location while a fetch is in flight - same uuids, so the SWR key doesn't change - and the hook reports our own `AbortError` as its error). I checked removal locally - the sticky-error fix still holds, the existing suite passes, and the page loads cleanly. The `catch` also rethrows in both branches, so it goes away for free.

packages/esm-bed-management-app/src/bed-administration/bed-administration-table.component.tsx
```diff
@@ -179,6 +182,11 @@ const BedAdministrationTable: React.FC = () => {
                 type="inline"
               />
             </div>
+            {errorFetchingBedsGroupedByLocation && hasLoadedRows ? (
+              <div className={styles.backgroundErrorNotification}>
+                <InlineLoading description={t('backgroundFetchFailed', 'Background fetch failed')} />
```
denniskigen: `InlineLoading` here renders an active spinner, so a failed refresh reads as something loading forever. But rather than swapping in a different component: a persistent notification in a widget card header isn't a pattern we use anywhere in O3 - `InlineNotification` lives in forms and workspaces, and non-blocking fetch/update failures are surfaced with `showSnackbar` (e.g. [error fetching patients](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-patient-search-app/src/compact-patient-search/compact-patient-search.component.tsx), [error updating upcoming appointments](https://github.com/openmrs/openmrs-esm-patient-management/blob/main/packages/esm-appointments-app/src/patient-appointments/patient-upcoming-appointments-card.component.tsx)). Since the ticket asks for a non-blocking warning, it's probably better to drop this block and fire a warning snackbar when a background revalidation fails while rows are on screen, roughly:

```tsx
useEffect(() => {
  if (errorFetchingBedsGroupedByLocation && hasLoadedRows) {
    showSnackbar({
      kind: 'warning',
      isLowContrast: true,
      title: t('backgroundFetchFailed', 'Background fetch failed'),
    });
  }
}, [errorFetchingBedsGroupedByLocation]);
```

That also keeps the header row intact on narrow screens.

packages/esm-bed-management-app/src/summary/summary.resource.ts
```diff
@@ -84,66 +84,66 @@ export const useLocationName = (locationUuid: string) => {
...
+          method: 'GET',
+          signal: controller.signal,
+        });
+        if (bedsFetchResult?.data?.results?.length) {
+          return bedsFetchResult.data.results.map((bed) => ({ ...bed, location }));
+        }
+        return null;
+      });
+
+      const updatedWards = (await Promise.all(promises)).filter(Boolean) as Array<Array<Bed & { location: any }>>;
```
denniskigen: nit: can we keep these typed? `Array<Array<Bed & { location: Location }>>` - `Location` just needs adding to the type import from `../types`. Ditto the cast on line 140.

denniskigen (changes requested):
Thanks for taking up the retarget from #2538, @omeriinnocent. I've left some feedback - the main blocker is that the new test file doesn't run under vitest.

omeriinnocent:
Hey @denniskigen thanks for the detailed review — I’ve applied the requested fix. Background refresh failures now keep the existing rows visible, and the error clears again once a later revalidation succeeds, i have confirmed this with tests which all are passing.

omeriinnocent:
@chibongho @denniskigen @VeronicaMuthee may you take a look at this

omeriinnocent:
hey @denniskigen @jayasanka-sack @VeronicaMuthee ,can this go in?

PR #2681, opened 2026-08-16
(feat) O3-5770: Add a Clinic Administrator screen for Service Queues

UjjawalPrabhat:
## Requirements

## Summary

Clinic-wide monitoring for Service Queues, as a tab inside `<Home>` rather than a screen of its own, so an administrator keeps the standard dashboard alongside it. Users without the privilege see no tabs and their dashboard is unchanged. Off by default, since the privilege is satisfied by any System Developer.

Also reworks the by-queue page to match: in-service patients become Attending cards, each remaining status gets its own table, and the header's filter box and **Add patient** button move into those tables' toolbars — **Add patient** only on the waiting one.

Worth noting:

- "Average wait" here and "Avg. wait time" on the dashboard measure different things — waits that have finished, versus how long the people still waiting have been waiting, so they will not agree. With nobody seen yet today one reads `--` while the other reads hours.
- Waiting and Attending count only the two configured statuses, so in a deployment using some other unfinished status they will sum to less than the queue's total.

## Screenshots

https://github.com/user-attachments/assets/e286b124-129b-45a5-baa8-4322751a8581

## Related Issue

[O3-5664](https://openmrs.atlassian.net/browse/O3-5664)

## Other

[O3-5664]: https://openmrs.atlassian.net/browse/O3-5664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/esm-service-queues-app/src/clinic-administrator/clinic-administrator-home.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { type Queue } from '../types';
+import ClinicAdministratorHome from './clinic-administrator-home.component';
+
+vi.mock('../hooks/useClinicQueueMetrics');
+vi.mock('../create-queue-entry/hooks/useQueueLocations');
+vi.mock('../store/store', async (importOriginal) => ({
+  ...((await importOriginal()) as object),
+  updateSelectedQueueLocationUuid: vi.fn(),
+  updateSelectedQueueLocationName: vi.fn(),
+  useServiceQueuesStore: vi.fn(() => ({ selectedQueueLocationUuid: null, selectedQueueLocationName: null })),
```
jwnasambu: Kindly the mock at this line hardcodes `useServiceQueuesStore` to always return { selectedQueueLocationUuid: null }. The test at line 109 checks that updateSelectedQueueLocationUuid was called with 'loc-2', but never verifies that useClinicQueueMetrics actually received that UUID. If the wiring between the store update and the hook call were broken e.g., stale closure, missing subscription, the test would still pass. Adding an assertion that mockUseClinicQueueMetrics was called with 'loc-2' as the location argument closes this gap.

packages/esm-service-queues-app/src/clinic-administrator/clinic-administrator-home.test.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  it('puts the queue with the longest wait first, so the worst one needs no sorting', () => {
+    render(<ClinicAdministratorHome />);
+
+    const rows = screen.getAllByRole('row').slice(1);
+    expect(rows[0]).toHaveTextContent('Clinician review');
+    expect(rows[rows.length - 1]).toHaveTextContent('Antenatal');
+  });
+
+  it('narrows to a single location when one is chosen', async () => {
```
jwnasambu: This line tests the manual dropdown selection path. There is no test that verifies what happens when queueLocations has exactly one entry  specifically, whether PatientQueueHeader's useEffect fires and writes to the store before the dashboard renders. This is the exact path the Greptile finding describes, and the missing test is why the bug went unnoticed. Adding a test that renders ClinicAdministratorHome with a single queue location and asserts that updateSelectedQueueLocationUuid is not called on mount would guard against regression once the blocking fix is applied.

packages/esm-service-queues-app/src/root.component.tsx
```diff
@@ -25,6 +29,16 @@ const Root: React.FC = () => {
...
 };
 
+function DefaultView() {
+  const session = useSession();
+
+  if (!session?.user) {
+    return null;
+  }
+
+  return userHasAccess(clinicAdministratorPrivilege, session.user) ? <ClinicAdministratorHome /> : <Home />;
```
chibongho: This means admins will see a different UI from non-admins (which sees the queue table in `<Home>`), with no way for admins to see the non-admin UI.

I wonder if we should just show the the admin screen within `<Home>`, if you have the right permission, as a tab inside it. (So Admins see 2 tabs, and non-admins just see what they see now.) That way they can switch between this admin screen and the current `<Home>` screen.

Thoughts? @VeronicaMuthee @denniskigen

packages/esm-service-queues-app/src/hooks/useClinicQueueMetrics.ts
```diff
@@ -0,0 +1,116 @@
...
+  totalWaitMinutes: 0,
+  averageWaitMinutes: null,
+  longestWait: null,
+};
+
+function average(totalWaitMinutes: number, waiting: number) {
+  return waiting === 0 ? null : Math.round(totalWaitMinutes / waiting);
+}
+
+export function useClinicQueueMetrics(locationUuid?: string) {
```
chibongho: It doesn't feel great that we're calculating the wait time statistics client-side, as:
- we need to fetch queue entries across all queues in the selected location, which could measure in the hundreds. With `useQueueEntries()` using `useOpenmrsFetchAll()`, this could mean multiple requests.
- there is already similar statistics calculation on the server-side. 
  - I understand that it doesn't given us everything we want, as we need the stats calculation for each queue and also for all queues in the location, but I think adding another endpoint where it calls `QueueUtils.computeAverageWaitTimeInMinutes()` for each queue and also for all queues in the location would do.

packages/esm-service-queues-app/src/clinic-administrator/clinic-administrator-home.component.tsx
```diff
@@ -0,0 +1,240 @@
...
+              className={styles.locationDropdown}
+              id="clinicAdministratorLocationDropdown"
+              items={[{ id: allLocations, name: allLocationsLabel }, ...queueLocations]}
+              itemToString={(item: { name?: string } | null) => item?.name ?? ''}
+              label={selectedQueueLocationName ?? allLocationsLabel}
+              onChange={handleLocationChange}
+              titleText={t('location', 'Location')}
+              type="inline"
+            />
+            <ConfigurableLink to={`${window.getOpenmrsSpaBase()}home/service-queues/waiting-list`}>
```
chibongho: These 2 links look out of place. I think if we inline the admin screen into `<Home>` then we can do away with these links. (See comment below)

packages/esm-service-queues-app/src/hooks/useClinicQueueMetrics.ts
UjjawalPrabhat: @chibongho I do agree that it might not be a great idea to cal wait time statistics client-side. But do you think a seperate end point would be worth it? 
I've tried a bit of a diff approach by having a `groupBy=queue` parameter on the existing `/queue-entry-metric` instead. Raised the PR [here](https://github.com/openmrs/openmrs-module-queue/pull/123). Let me know your thoughts!!

packages/esm-service-queues-app/src/root.component.tsx
UjjawalPrabhat: @chibongho Applied your suggestions and rebased the PR with the updated design changes. Have another look!!

CC: @ibacher @NethmiRodrigo

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.component.tsx
```diff
@@ -0,0 +1,217 @@
...
+          <QueueDuration startedAt={totals.longestWait.startedAt} thresholds={waitTimeThresholds} />
+        ),
+    },
+  ];
+
+  return (
+    <>
+      <div className={metricsStyles.metricsBorder} data-testid="clinic-administrator-metrics">
+        {totalsCards.map(({ key, title, value }) => (
+          <QueueMetricTile key={key} headerLabel={title} value={isLoading || value == null ? '--' : value} />
```
denniskigen: **[suggestion]** The table shows ErrorState when the roll-up fails, but the tiles above still render. Waiting and attending come through as 0, so it looks like an empty clinic rather than a failed load. I checked `useClinicQueueMetrics`: on error `data` is missing, so totals are the empty summarise (zeros / nulls). `AverageWaitTimeExtension` already treats error as `--`. The current error test still feeds success totals, so it would not catch this.

**Suggestion:** Include `error` in the dash condition, same as the avg-wait card.

```tsx
<QueueMetricTile key={key} headerLabel={title} value={isLoading || error || value == null ? '--' : value} />
```

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -19,19 +12,33 @@ import styles from './attending-patients.scss';
...
-  const { queueEntries, isLoading, error } = useQueueEntries({
-    service: selectedServiceUuid,
-    location: selectedQueueLocationUuid,
-    status: defaultTransitionStatus,
-    isEnded: false,
-  });
+
+  const searchCriteria = useMemo(
+    () =>
+      queueUuid
```
NethmiRodrigo, line 30: There is no test where we verify that passing the queue id fetches the entries of that queue instead of using the selected location and service. `attending-patients.test.tsx` only renders `<AttendingPatients />` without props, and `queue-tables-for-all-statuses.test.tsx` mocks `AttendingPatients` out. Lets add a test in `attending-patients.test.tsx` that renders `<AttendingPatients queueUuid="q1" />` and asserts `useQueueEntries` was called with the queue and not the store's location.

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.component.tsx
NethmiRodrigo:
```suggestion
          <QueueMetricTile key={key} headerLabel={title} value={isLoading || error || value == null ? '--' : value} />
```

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+    render(<ClinicOverview />);
+
+    expect(screen.getByRole('link', { name: 'Clinician review' })).toHaveAttribute(
+      'href',
+      expect.stringContaining('queue-table-by-status/q2'),
+    );
+    expect(screen.getAllByRole('columnheader')).toHaveLength(7);
+  });
+
+  it('puts the queue with the longest wait first', () => {
```
NethmiRodrigo, line 156: The initial order comes from the rows useMemo in ClinicOverview, and Carbon only calls `sortRow` when a user clicks a column header, so rendering and asserting the order never runs `sortRow` at all. Lets add a test that clicks the `Longest wait` header twice and asserts the row order after each click, so the numeric compare and the nulls-at-the-bottom rule are tested.

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+function queue(uuid: string, display: string, location: string, service: string): Queue {
+  return { uuid, display, name: display, location: { display: location }, service: { display: service } } as Queue;
+}
+
+function rollup(queue: Queue, waiting: number, attending: number, longestWaitMinutes: number | null, patient = '') {
+  return {
+    queue,
+    waiting,
+    attending,
+    totalWaitMinutes: longestWaitMinutes ?? 0,
```
NethmiRodrigo: `totalWaitMinutes` isn't a field of `QueueRollup` or of the totals that `useClinicQueueMetrics` returns, the as` ReturnType<typeof useClinicQueueMetrics>` cast in `givenMetrics` is what keeps `tsc` from flagging it. 

```suggestion
```

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+  rollup(queue('q3', 'Antenatal', 'MCH clinic', 'Antenatal care'), 0, 0, null),
+];
+
+function givenMetrics(overrides: Partial<ReturnType<typeof useClinicQueueMetrics>> = {}) {
+  mockUseClinicQueueMetrics.mockReturnValue({
+    rollups,
+    totals: {
+      waiting: 20,
+      attending: 5,
+      totalWaitMinutes: 520,
```
NethmiRodrigo: Ditto
```suggestion
```

packages/esm-service-queues-app/src/home.component.tsx
```diff
@@ -1,18 +1,72 @@
...
+          <Tab className={styles.tab}>{t('clinicOverview', 'Clinic overview')}</Tab>
+          <Tab className={styles.tab}>{t('waitingList', 'Waiting list')}</Tab>
+        </TabList>
+        <TabPanels>
+          <TabPanel className={styles.tabPanel}>
+            <ClinicOverview />
+          </TabPanel>
+          <TabPanel className={styles.tabPanel}>
+            <WaitingListView />
+          </TabPanel>
```
NethmiRodrigo, line 59: The Tabs are mounted even when they aren't selected so components like waiting would still do its polling, etc. 
```suggestion
          <TabPanel className={styles.tabPanel}>
             {selectedTabIndex === clinicOverviewTabIndex && <ClinicOverview />}
          </TabPanel>
          <TabPanel className={styles.tabPanel}>
             {selectedTabIndex !== clinicOverviewTabIndex && <WaitingListView />}
          </TabPanel>
```

packages/esm-service-queues-app/src/queue-table/default-queue-table.component.tsx
```diff
@@ -104,11 +122,13 @@ function QueueTableSection() {
...
-        queueUuid={null}
-        statusUuid={null}
+        queueUuid={queueUuid ?? null}
+        statusUuid={status ? statusUuid : null}
         tableFilters={
           <>
-            <AddPatientToQueueButton />
+            {/* Adding a patient puts them in the waiting status, so the control only belongs on
+                that table — not on Finished service, or any other status a deployment adds. */}
+            {statusUuid === waitingStatusConceptUuid && <AddPatientToQueueButton />}
```
NethmiRodrigo: Why are we showing the queue button if the status matches `waitingStatusConceptUuid` but when the user clicks the button, create the queue entry with `defaultStatusConceptUuid`?

packages/esm-service-queues-app/src/queue-table/default-queue-table.test.tsx
```diff
@@ -137,6 +138,41 @@ describe('DefaultQueueTable', () => {
...
+
+    it('heads the table with the status and offers to add a patient to a waiting list', () => {
+      const waitingStatus = {
+        uuid: getDefaultsFromConfigSchema<ConfigObject>(configSchema).concepts.waitingStatusConceptUuid,
+        display: 'Waiting',
+      } as Concept;
+
+      renderWithSwr(<DefaultQueueTable queueUuid="queue-uuid" status={waitingStatus} />);
+
+      expect(screen.getByRole('heading', { name: /waiting \(0\)/i })).toBeInTheDocument();
```
NethmiRodrigo, line 168: We should assert that the search criteria was correct
```suggestion
      expect(mockUseQueueEntries).toHaveBeenCalledWith({ queue: 'queue-uuid', isEnded: false, status: waitingStatus.uuid });
       expect(screen.getByRole('heading', { name: /waiting \(0\)/i })).toBeInTheDocument();
```

packages/esm-service-queues-app/src/views/queue-tables-for-all-statuses.component.tsx
```diff
@@ -40,130 +40,36 @@ const QueueTablesForAllStatuses: React.FC<QueueTablesForAllStatusesProps> = ({
...
-        <QueueTablesByStatus selectedQueue={selectedQueue} searchTerm={searchTerm} />
-      ) : (
+      <PatientQueueHeader title={!isLoadingQueue ? selectedQueue?.display : <SkeletonText />} />
+      {isLoadingQueue ? (
         <QueueTableByStatusSkeleton />
+      ) : (
+        <>
+          <QueueTableMetrics selectedQueue={selectedQueue} />
+          <AttendingPatients queueUuid={selectedQueue.uuid} />
+          {tabledStatuses.length === 0 ? (
```
NethmiRodrigo: This check must count the in-service status too, otherwise a queue that only allows In Service shows "No status configured"
```suggestion
          {!selectedQueue?.allowedStatuses?.length ? (
```

packages/esm-service-queues-app/src/views/queue-tables-for-all-statuses.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <div data-testid="status-table">
+      {status.display}:{queueUuid}
+    </div>
+  )),
+}));
+
+const config = getDefaultsFromConfigSchema<ConfigObject>(configSchema);
+const inServiceUuid = config.concepts.defaultTransitionStatus;
+
+// The backend returns allowed statuses in reverse workflow order, which the view flips back.
```
NethmiRodrigo: The order comes from the concept set
```suggestion
// The concept set in the dictionary returns the statuses as [Finished service, In Service, Waiting].
 // The view reverses the list, so the tables show Waiting first.
```

packages/esm-service-queues-app/src/hooks/useClinicQueueMetrics.ts
```diff
@@ -0,0 +1,89 @@
...
+export interface QueueRollup extends Summary {
+  queue: Queue;
+}
+
+export function useClinicQueueMetrics(locationUuid?: string, serviceUuid?: string) {
+  const {
+    concepts: { waitingStatusConceptUuid, defaultTransitionStatus },
+  } = useConfig<ConfigObject>();
+
+  const searchParams = new URLSearchParams({ groupBy: 'queue', isEnded: 'false' });
```
denniskigen, line 41: Can we keep the wait metrics scoped to waiting entries? `isEnded=false` includes both Waiting and In Service entries. [The queue module ends the old entry and starts a new one at transition](https://github.com/openmrs/openmrs-module-queue/blob/c3a8acddc76d2a8c3f69a3a8164a9dac3b223c0c/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java#L123-L147), so an In Service entry's `startedAt` is the service start. `averageOpenWaitTime` and `longestOpenWait` can therefore report time in service as waiting time.

Adding `status=waitingStatusConceptUuid` here would also remove the attending counts because [#123 calculates all three metrics from the same result set](https://github.com/openmrs/openmrs-module-queue/blob/c3a8acddc76d2a8c3f69a3a8164a9dac3b223c0c/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java#L105-L135). This probably needs a second filtered request for the wait metrics or a metric-specific status filter in #123.

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -158,6 +158,25 @@ export const configSchema = {
       'what your digital signage expects (e.g. "Now serving"). ' +
       'Avoid "completed", which the queue backend treats as a signal to remove the ticket.',
   },
+  clinicAdministratorScreen: {
+    enabled: {
+      _type: Type.Boolean,
+      _default: false,
+      _description:
```
denniskigen: Can we mention the dependency on [queue module #123](https://github.com/openmrs/openmrs-module-queue/pull/123) here? Without it, `/queue-entry-metric` does not understand `groupBy` or the new metric names, so enabling this screen quietly shows an empty clinic instead of an error. Once #123 is released, it's probably best to name the minimum queue module version.

packages/esm-service-queues-app/src/home.test.tsx
```diff
@@ -55,4 +80,77 @@ describe('Home Component', () => {
       expect(screen.getByRole('columnheader', { name: new RegExp(header, 'i') })).toBeInTheDocument();
     });
   });
+
+  it('shows no tabs at all for a user without access, so their dashboard is unchanged', () => {
+    render(<Home />);
+
+    expect(screen.queryByRole('tab')).not.toBeInTheDocument();
+    expect(screen.queryByTestId('clinic-overview')).not.toBeInTheDocument();
+  });
```
denniskigen:
```suggestion
  it('shows no tabs at all for a user without access, so their dashboard is unchanged', () => {
    givenConfig({ enabled: true });
    mockUserHasAccess.mockReturnValue(false);

    render(<Home />);

    expect(screen.queryByRole('tab')).not.toBeInTheDocument();
    expect(screen.queryByTestId('clinic-overview')).not.toBeInTheDocument();
  });
```

As written, this only exercises `enabled: false`, so it would still pass if `userHasAccess` stopped gating the tabs.

denniskigen (dismissed):
Thanks for the follow-up, @UjjawalPrabhat. The latest changes address the earlier feedback well; I found one issue with the wait metric scope and left two smaller suggestions.

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.component.tsx
```diff
@@ -0,0 +1,217 @@
...
+          <div className={queueTableStyles.headerContainer}>
+            <div className={isDesktop(layout) ? queueTableStyles.desktopHeading : queueTableStyles.tabletHeading}>
+              <h2>{t('queues', 'Queues')}</h2>
+            </div>
+          </div>
+
+          {isLoading ? (
+            <DataTableSkeleton role="progressbar" compact={isDesktop(layout)} zebra columnCount={7} rowCount={4} />
+          ) : error ? (
+            <ErrorState error={error} headerTitle={t('queues', 'Queues')} />
```
denniskigen: This renders “Queues” twice in the error state, once in the section header and again inside `ErrorState`. Can we keep one heading and give the error card the same inline spacing as the empty state?

The loading skeleton probably wants the same spacing. It also renders Carbon’s header and toolbar placeholders by default even though this table has neither, so can we set `showHeader={false}` and `showToolbar={false}` here?

packages/esm-service-queues-app/src/hooks/useClinicQueueMetrics.ts
```diff
@@ -0,0 +1,112 @@
...
+    return `${restBaseUrl}/queue-entry-metric?${searchParams.toString()}`;
+  }
+
+  const counts = useSWR<{ data: GroupedByQueue<CountMetrics> }, Error>(metricsUrl(['countsByStatus']), openmrsFetch);
+
+  // A separate request, scoped to waiting entries: the queue module ends an entry and starts a new
+  // one at transition, so an In Service entry's `startedAt` is when service began rather than when
+  // the patient joined the queue. Open waits rather than `averageWaitTime`, which the queue module
+  // measures over waits that have already finished.
+  const waits = useSWR<{ data: GroupedByQueue<WaitMetrics> }, Error>(
```
denniskigen: Since this went in, [queue module #123](https://github.com/openmrs/openmrs-module-queue/pull/123) grew a `waitStatus` parameter that does this scoping server-side: the counts still cover every unfinished entry while the two wait metrics measure only entries in the named status ([controller](https://github.com/openmrs/openmrs-module-queue/blob/6368e350cae41fd259d323a4103415aecf5500cc/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java#L110-L139), [test covering exactly this shape](https://github.com/openmrs/openmrs-module-queue/blob/6368e350cae41fd259d323a4103415aecf5500cc/omod/src/test/java/org/openmrs/module/queue/web/QueueEntryMetricRestControllerTest.java#L324-L347)). Can we go back to a single request with `waitStatus=waitingStatusConceptUuid` and all three metrics? That halves the requests - each of these loads every unfinished entry at the location on the server - measures counts and waits at one instant, and deletes the join below.

It'd also let us drop the "leaves out queues with nobody waiting" premise: the backend seeds a row for every queue at the location, so a queue with nobody waiting comes back with `averageOpenWaitTime: null` rather than being absent.

denniskigen:
Should the Clinic overview initially cover all queue locations, or should it default to the user’s login location?

UjjawalPrabhat:
> Should the Clinic overview initially cover all queue locations, or should it default to the user’s login location?

Good observation @denniskigen!! So, the [current behaviour](https://github.com/openmrs/openmrs-esm-patient-management/blob/35c29ab8102ab1990c56f69bd8e74de95f7d5bd5/packages/esm-service-queues-app/src/patient-queue-header/queue-location-header-title.component.tsx#L35-L41) is the latter: it defaults to the login location rather than all locations. The header finds the nearest `queue location`-tagged ancestor of the login location, falling back to all locations if none exists in the chain.

Ideally, I think we'd want the Clinic overview to be clinic-scoped, while the Waiting list staying as-is, since the whole point of introducing the clinic screen is meant to separate administrator vs. clerk/clinician workflows.

A couple of things complicate this though:

* The location filter is shared, so changing the default affects both tabs.
* We could make the default configurable based on privilege, but the [backend](https://github.com/openmrs/openmrs-module-queue/blob/706ca3b1c9c0503ff99c2e8f0f53a5c38eab82ae/api/src/main/java/org/openmrs/module/queue/api/dao/impl/QueueEntryDaoImpl.java#L152) matches locations literally - `location=<hospital>` returns nothing because queues are attached to the clinics below it. Repeated `location` params do work, but then the frontend needs to resolve and enumerate the queue locations under the hospital.

Given that, how do you think we should go about it?

CC: @ibacher @NethmiRodrigo

packages/esm-service-queues-app/src/config-schema.ts
```diff
@@ -158,6 +158,28 @@ export const configSchema = {
       'what your digital signage expects (e.g. "Now serving"). ' +
       'Avoid "completed", which the queue backend treats as a signal to remove the ticket.',
   },
+  clinicAdministratorScreen: {
```
NethmiRodrigo, line 161: Do we usually block off entire screens by configuration @denniskigen? I feel like since this is gated by the privilege, isn't that sufficient? 

> Off by default because the privilege is ' +
        'satisfied by any user with the System Developer role, so turning this on would give the tabs to ' +
        'every administrator whether or not the privilege has been rolled out. 

I see this comment, but I don't think this is a valid concern. What I mean is, we don't do this in another places just because the System Developer satisfies all privileges

packages/esm-service-queues-app/src/admin/admin-page/admin-page.component.tsx
```diff
@@ -147,10 +147,7 @@ const AdminPage = () => {
               )}
             </DataTable>
             {queueTableRows.length === 0 && (
-              <Tile className={styles.emptyState}>
-                <EmptyCardIllustration />
-                <p>{t('noQueuesToDisplay', 'No queues to display')}</p>
-              </Tile>
+              <EmptyState className={styles.emptyState} displayText={t('noQueuesToDisplay', 'No queues to display')} />
```
NethmiRodrigo: Can't we move the `emptyState` style into the parent element of the `EmptyState` component so that we don't have to apply this styling at every occurrence of it.

packages/esm-service-queues-app/src/attending-patients/attending-patients.test.tsx
```diff
@@ -103,6 +103,19 @@ describe('AttendingPatients', () => {
     expect(screen.queryByText('Patient 3')).not.toBeInTheDocument();
   });
 
+  it('asks for the given queue rather than the selected location and service', () => {
```
NethmiRodrigo:
```suggestion
  it('uses the queue when given rather than the selected location and service', () => {
```

packages/esm-service-queues-app/src/clinic-administrator/clinic-overview.component.tsx
```diff
@@ -0,0 +1,231 @@
...
+import { useTranslation } from 'react-i18next';
+import { ConfigurableLink, ErrorState, isDesktop, useConfig, useLayoutType } from '@openmrs/esm-framework';
+import { useClinicQueueMetrics, type QueueRollup } from '../hooks/useClinicQueueMetrics';
+import EmptyState from '../empty-state/empty-state.component';
+import QueueDuration from '../queue-table/components/queue-duration.component';
+import { QueueMetricTile } from '../queue-table/queue-table-metrics-card.component';
+import { useServiceQueuesStore } from '../store/store';
+import { formatWaitTimeInMinutes } from '../wait-time';
+import { spaBasePath } from '../constants';
+import { type ConfigObject } from '../config-schema';
```
NethmiRodrigo: Nit - hooks shouldn't be before components
```suggestion
import EmptyState from '../empty-state/empty-state.component';
import QueueDuration from '../queue-table/components/queue-duration.component';
import { QueueMetricTile } from '../queue-table/queue-table-metrics-card.component';
import { useServiceQueuesStore } from '../store/store';
import { useClinicQueueMetrics, type QueueRollup } from '../hooks/useClinicQueueMetrics';
import { formatWaitTimeInMinutes } from '../wait-time';
import { spaBasePath } from '../constants';
import { type ConfigObject } from '../config-schema';
```

packages/esm-service-queues-app/src/config-schema.ts
NethmiRodrigo: Like this mentions ^, you need to update routes.json file with the updated version dependency

NethmiRodrigo:
> Given that, how do you think we should go about it?

I think given the UI design, the behaviour you've implemented is actually somewhat confusing. There is a location and queue service picker header that visually would seem to be applied to the two tabs, but it is not? I understand that applying the queue filter itself is not applicable to the clinic overview, but I think you should stick to the same behaviour for both tabs. The table should display whatever is selected in the queue header, and not the login location.

NethmiRodrigo:
Also, @UjjawalPrabhat can you add a screen-recording instead to the PR description. Especially to show what clicking on the individual queues in the clinic overview component does.

packages/esm-service-queues-app/src/config-schema.ts
NethmiRodrigo: @UjjawalPrabhat, let's remove this configuration

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat! I did some very minute changes to mark retired queues as retired and not have them be linkable since redirecting to a page that would just show an error isn't very helpful. I also added a loading indicator to the metric tiles so that they wouldn't show a `--` when loading.

PR #2692, opened 2026-08-23
(fix) O3-5918: Allow birth and death on the same calendar day

kanchana404:
## Requirements

## Summary

The registration form could not record a patient who died on the same calendar day they were born.

The `deathDate-after-birthdate` test in the registration validation schema required the death date to be strictly after the birth date. Both values are dates normalised to midnight, and the separately entered death time takes no part in that comparison, so any death on the birth date itself was rejected — even with a valid death time entered.

This compares the two values at calendar-day precision and accepts the death date when it is the same as, or later than, the birth date. The death time is deliberately not folded into this check: the neighbouring `deathDate-before-today` test already combines date and time to guard against deaths in the future, and duplicating that here would change behaviour the ticket doesn't ask for.

Tests cover the four boundaries from the acceptance criteria: a death date equal to the birth date passes, one day earlier fails, a later date passes, and a future death date still fails. The last of these is the pre-existing test, kept as a regression guard on the neighbouring `.max(new Date())` rule.

## Screenshots

N/A — validation logic only, no UI changes.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5918

## Other

No translation strings were changed, so the generated catalogs are untouched. Verified with `yarn verify` (27/27 tasks), the full registration suite (37 files, 273 passing), `tsc`, `eslint`, `check-translations` and a production build.

### DIFF
--- a/e2e/support/github/Dockerfile
+++ b/e2e/support/github/Dockerfile
@@ -24,10 +24,15 @@ RUN apk update && \
 RUN rm -rf /usr/share/nginx/html/*
 
 COPY --from=frontend /etc/nginx/nginx.conf /etc/nginx/nginx.conf
-# this assumes that NOTHING in the framework is in a subdirectory
-COPY --from=frontend /usr/share/nginx/html/* /usr/share/nginx/html/
+COPY --from=frontend /usr/share/nginx/html/ /usr/share/nginx/html/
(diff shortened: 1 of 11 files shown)


denniskigen (approved):
LGTM! Thanks, @kanchana404. Nice, focused fix, and the boundary tests make the intended calendar-day behavior clear.

kanchana404:
Thanks @denniskigen

PR #2694, opened 2026-08-24
(fix) O3-5919: Clear stale death metadata when marking a patient alive

kanchana404:
## Requirements

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. (See also: [Styleguide](http://om.rs/o3ui)).

## Summary

Marking a deceased patient as alive sent `dead: false` but omitted the death metadata fields. Because the existing non-null death date remained stored, OpenMRS continued to return `dead: true`, along with the stale death date and non-coded cause of death.

The cause is that omitting a property is not the same as clearing it. When `isDead` is false, `getPatientDeathInfo` returned only `dead: false` and left the death fields out of the payload entirely. OpenMRS clears the coded `causeOfDeath` by itself, but it retains `deathDate` and `causeOfDeathNonCoded` when those properties are absent from the update. Sending explicit nulls is what actually clears them.

This sends `null` for `deathDate`, `causeOfDeath` and `causeOfDeathNonCoded` when the patient is not dead, mirroring the style the `isDead: true` branch already uses when it nulls whichever cause-of-death field does not apply. The person payload types are widened to permit null, and `causeOfDeathNonCoded` is added — it was missing from the type altogether even though the `isDead: true` branch has always sent it.

**Scope.** This is confined to `getPatientDeathInfo` and the death-related types. It does not touch `getDeletedNames` (O3-5910, #2693) or the unawaited `forEach` around `deletePersonName` (O3-5913), both of which live in the same file and are being handled separately.

One test outside those files changed: the payload assertion in `patient-registration.test.tsx` matches `person` exactly rather than partially, so it had to record the three new properties. Without that the existing test fails. It has the side benefit of pinning the new payload at the render-to-request level.

Tests cover both levels the ticket asks for: a unit test that the generated payload contains `dead: false` plus the three explicit nulls, and a request-level test asserting on the body actually handed to `savePatient`. Both fail without the fix. The coded and non-coded cause-of-death paths are also covered so the `isDead: true` behaviour stays pinned.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5919

## Other

Verified with `yarn verify` (27/27 tasks), the full registration suite (38 files, 283 passing), `tsc`, `eslint` and a production build. FHIR still represents the patient as not deceased, since `mapPatientToFhirPatient` derives `deceasedBoolean` from `dead`.


denniskigen (approved):
LGTM. Thanks, @kanchana404!

PR #2697, opened 2026-08-24
(fix) O3-5913: Wait for patient data deletions before reporting save success

denniskigen:
## Requirements

## Summary

Patient edits could report success while name, identifier, or person-attribute deletions were still running. The deletion paths discarded their promises, and attribute deletion also caught failures and only logged them. A failed cleanup could therefore leave stale patient data behind after the UI reported a successful save.

This change collects and awaits the name and identifier deletion requests. It also moves attribute deletion out of the synchronous patient-payload builder and into the asynchronous save flow. Deletion failures now reject the save through the existing error handling, and the patient update does not begin until the requested deletions complete.

Deferred-promise tests cover all three deletion types. They verify that the patient save remains pending while deletion is pending and does not run when deletion fails.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5913


packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -114,6 +114,13 @@ export class FormManager {
       currentLocation,
     );
 
+    await Promise.all([
+      ...FormManager.getDeletedNames(values, patientUuidMap).map((name) =>
+        deletePersonName(name.nameUuid, name.personUuid),
+      ),
+      FormManager.deletePatientAttributes(isNewPatient, values, patientUuidMap),
+    ]);
```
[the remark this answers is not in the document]
denniskigen: Thanks for flagging this. The 404 premise does not match the OpenMRS REST behavior. These requests do not use `purge=true`, so the resources are voided, not removed. The REST resource looks them up by UUID, and those lookups include voided rows, so `PersonNameResource1_8.delete` and the person-attribute equivalent simply void and save the same row again. I verified this against a live backend: a second DELETE on an already-voided name returns 204, and the same is true for a person attribute. The T-Rex reproduction mocks `deletePersonName` to reject with a 404 on the second submit, so it demonstrates the consequence of an assumed response rather than actual backend behavior. We shouldn't need client-side retry state or special 404 handling here.

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -114,6 +114,13 @@ export class FormManager {
       currentLocation,
     );
 
+    await Promise.all([
```
ibacher: Should this be `allSettled()` rather than `all()`? This is semi-related to the greptile comment, but `all()` will effectively short-circuit if either of the two requests rejects which could mean we continue execution while one of the requests is still being processed by the server.

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -279,15 +282,19 @@ export class FormManager {
...
-    if (patientUuid) {
-      Object.keys(initialIdentifierValues)
-        .filter((identifierFieldName) => !patientIdentifiers[identifierFieldName])
-        .forEach(async (identifierFieldName) => {
-          await deletePatientIdentifier(patientUuid, initialIdentifierValues[identifierFieldName].identifierUuid);
-        });
-    }
-
-    return Promise.all(identifierTypeRequests);
+    const identifierDeletionRequests = patientUuid
```
ibacher, line 303: I _think_ the `patientUuid` guard was here intended to only fire these requests for an existing patient. Are we sure it's safe to drop?

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
```diff
@@ -279,15 +282,19 @@ export class FormManager {
...
+      ? Object.keys(initialIdentifierValues)
+          .filter((identifierFieldName) => !patientIdentifiers[identifierFieldName])
+          .map((identifierFieldName) =>
+            deletePatientIdentifier(patientUuid, initialIdentifierValues[identifierFieldName].identifierUuid),
+          )
+      : [];
+
+    const [identifiers] = await Promise.all([
```
ibacher: Similar remark here about `all()` vs `allSettled()` albeit I think it only applies to the "inner" `all()` calls (i.e., wrapping two `allSettled()` calls in `all()` seems fine.

ibacher (approved):
Weak recommendation to consider `allSettled()` over `all()` for these use cases, but LGTM.

ibacher (approved):

packages/esm-patient-registration-app/src/patient-registration/form-manager.ts
denniskigen: The guard is still here, just expressed by the ternary. When `patientUuid` is empty, `identifierDeletionRequests` is `[]`, so no deletion requests are sent for a new patient.

PR #2702, opened 2026-08-27
(fix) Add memoization of search results and reduce secondary queries

ibacher:
## Requirements

## Summary

This fixes some minor issues found while troubleshooting search slowness in O3. None of these changes affect the major causes of the performance issues, but they are all no-brainers that improve the general performance characteristics of the search. Specifically:

1. SWR configuration tweaks to reduce request churn are moved higher in the tree. The goal here is to make it so we only try to fetch the patient photo, etc. once.
2. Memoized the `mappedData` which we tried to use as a memoization key lower down, but which never worked because `mappedData` changed on every execution
3. Stopped revalidating the first page of search results since the search endpoint is heavy.

The trade-off here is that someone could have slightly stale data in the patient search if the data changes while they keep the search open, but it should improve the overall performance and the loss of up-to-the-minute data has only very minor (IMO) implications.


pirupius (approved):
No reservations on my end.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -95,20 +95,27 @@ export function useInfinitePatientSearch(
 
   const shouldFetch = isSearching && Boolean(searchQuery);
 
+  // Re-fetching page 1 on every page load would cost a round trip and replace the rendered rows'
+  // objects, breaking the identity they memoize on.
   const { data, isLoading, isValidating, setSize, error, size } = useSWRInfinite<InfinitePatientSearchResponse, Error>(
     shouldFetch ? getUrl : null,
     fetcher,
-    { keepPreviousData: true },
+    { keepPreviousData: true, revalidateFirstPage: false },
```
denniskigen: I traced this through SWR 2.2.5 and reproduced it locally: with `revalidateFirstPage: false`, a cached first page is not fetched again when the same search key remounts or on focus. It remains cached for the SPA session unless the cache is cleared or the hook is explicitly mutated.

One workflow I’m concerned about is: a clerk searches for “Mary” and gets no match, registers Mary, then searches for “Mary” again to check her in. Registration uses SPA navigation and does not invalidate the patient-search key, so the cached empty result is shown and Mary remains missing.

Can we bound this staleness somehow, perhaps by explicitly revalidating only page 1 when the search UI reopens while retaining `revalidateFirstPage: false` for `setSize` appends? That would preserve the scrolling optimization without caching the initial result for the entire session.

denniskigen:
Good stuff, @ibacher. The memoization changes check out. I left one question about the lifetime of cached first-page results.

packages/esm-patient-search-app/src/patient-search.resource.tsx
ibacher: Good point. I've added a hook to invalidate the cache when the search component mounts. This _should_ mean that we revalidate basically every time the search is re-opened while not just revalidating on every page fetch. That feels to me like the right UX, but I'm happy to hear disagreements.

ibacher:
NB I don't want to just sneak this in, but I noticed that the patient search extension we use elsewhere _doesn't_ have debouncing and have added that here. Debouncing user requests is helpful so that we don't trigger too many queries in the backend, but also so that we're not populating the SWR cache with too much noise.

packages/esm-patient-search-app/src/patient-search.resource.tsx
denniskigen: Thanks, this addresses my concern. I verified that cached first-page results are refreshed when search is reopened without revalidating page 1 during pagination. The added debouncing also looks good.

denniskigen (approved):
LGTM. Thanks, @ibacher!

PR #2708, opened 2026-08-31
(fix) Keep patient search on the selected page and load result pages in parallel

pirupius:
## Requirements

## Summary

Fixes two problems in the advanced patient search. Follow up to #2702 

### Pagination reverted to page 1

Selecting page 2 would snap back to page 1 a moment later.

`usePagination` rebuilds `goTo` whenever `totalPages` changes, and the search page reset the page with `useEffect(() => goTo(1), [query, goTo])`. Because the search appends result pages in the background, `totalPages` kept growing, `goTo` kept getting a new identity and the effect kept firing. The reset now runs only when the query itself changes.

### Search was slow

The search loaded the entire result set 50 rows at a time, each request waiting on the previous one, before the first 20 rows settled. Measured with a probe against the old code: a 500-result query took **10 serialised round trips**; a 1700-result query took 34.

`useInfinitePatientSearch` now fetches pages with SWR's `parallel: true`, and the advanced search requests them in waves of 4 rather than walking the set one page at a time. A 1700-result query now costs **10 round trips instead of 34**, with at most 4 requests in flight.

Two smaller fixes:

- `totalCount=true` was sent on *every* page request, but only page 1's count is ever read, so now only page 1 asks for it. On this path the count is the size of the list the search handler already built in memory, so this doesn't save a backend query; it just asks for what we use.
- `totalResults` read the count from `data[0]`, which under `keepPreviousData` is the *outgoing* query's page so it reported the previous query's count during a query change. It now reads the count for the current query.

## Screenshots

Before

https://github.com/user-attachments/assets/2acbd5f4-db0c-4f49-9136-1fd9125e28c3

After

https://github.com/user-attachments/assets/654b49bc-8d40-4d11-8ba0-97614a127d82

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
@@ -12,10 +12,16 @@ const resultsPerPage = 50;
 
 /**
  * How many pages to request at once. Every page of a result set still gets loaded — this only
- * bounds how many are in flight together, so a query matching thousands of patients arrives in a
- * handful of waves instead of one burst of requests at the backend.
+ * bounds how many are in flight together, so a broad query arrives in a handful of waves instead of
+ * one burst of requests at the backend.
+ *
+ * Deliberately small. The REST patient search re-runs the whole search and hydration on every page
+ * request and applies `startIndex` in memory at the end, so each page in a wave costs a full search
+ * at the backend, not a slice of one. Paging serially held that to one at a time; this trades a
+ * bounded amount of concurrency for far fewer round trips, and four is low enough for a modest
+ * install to absorb when several people are searching at once.
  */
-const pagesPerWave = 8;
+const pagesPerWave = 4;
 
 interface AdvancedPatientSearchProps {
   query: string;
@@ -56,11 +62,12 @@ const AdvancedPatientSearchComponent: React.FC<AdvancedPatientSearchProps> = ({
 
   // The refine-search filters and the result count below both run over the rows held on the client,
   // so the whole result set has to be loaded. `useInfinitePatientSearch` fetches pages in parallel,
-  // so request them a wave at a time: a 1700-result query then costs five round trips rather than
-  // the thirty-four it took to walk the set one page at a time, without leaving any patient
-  // unreachable. Waiting on `isValidating` is what keeps a wave from being requested on top of the
-  // one still in flight — `currentPage` advances as soon as the pages are asked for, not when they
-  // land, so without it every wave would collapse back into a single burst.
+  // so request them a wave at a time: `person.searchMaxResults` caps the search at 1000 patients by
+  // default, so even a broad query is around twenty pages, and those arrive in five waves rather
+  // than twenty sequential round trips — without leaving any patient unreachable. Waiting on
+  // `isValidating` is what keeps a wave from being requested on top of the one still in flight —
+  // `currentPage` advances as soon as the pages are asked for, not when they land, so without it
+  // every wave would collapse back into a single burst.
   const pagesNeeded = Math.ceil(totalResultsForQuery / resultsPerPage);
 
   // A failed request also ends the wave, so stop on `fetchError`: the count that `pagesNeeded` is
--- a/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
@@ -8,6 +8,15 @@ import PatientSearchComponent from './patient-search-lg.component';
 import RefineSearch, { initialFilters } from './refine-search/refine-search.component';
 import styles from './advanced-patient-search.scss';
 
+const resultsPerPage = 50;
+
+/**
+ * How many pages to request at once. Every page of a result set still gets loaded — this only
+ * bounds how many are in flight together, so a query matching thousands of patients arrives in a
+ * handful of waves instead of one burst of requests at the backend.
+ */
+const pagesPerWave = 8;
+
 interface AdvancedPatientSearchProps {
   query: string;
   inTabletOrOverlay?: boolean;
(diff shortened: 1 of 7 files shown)


denniskigen:
Can you untangle your changes from https://github.com/openmrs/openmrs-esm-core/pull/1856? It's not likely that the linked PR will get merged today.

pirupius:
@denniskigen i've tested the branch independent of the core PR and updated the PR description with the sample videos with before (main) and the branch in this PR.  
The test with core was just to address the pending issue with pagination that was discussed in the core PR

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -145,16 +142,40 @@ export function useInfinitePatientSearch(
...
+  );
 
   // Re-fetching page 1 on every page load would cost a round trip and replace the rendered rows'
   // objects, breaking the identity they memoize on. See `useRevalidateFirstPageOnce` for how the
   // resulting staleness is bounded.
   const { data, isLoading, isValidating, setSize, error, size, mutate } = useSWRInfinite<
     InfinitePatientSearchResponse,
     Error
-  >(shouldFetch ? getUrl : null, fetcher, { keepPreviousData: true, revalidateFirstPage: false });
+  >(shouldFetch ? getUrl : null, fetcher, { keepPreviousData: true, revalidateFirstPage: false, parallel: true });
```
denniskigen, line 178: nit: Do we have any load numbers for `pagesPerWave = 8`? These requests are served by [`LivingPatientSearchHandler1_11`](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_11/LivingPatientSearchHandler1_11.java), which calls the unpaged `getPatients(searchPhrase)` from offset 0 for up to `person.searchMaxResults` (default 1000) on every page request, filters dead patients in Java, and only then applies the REST `startIndex` and `limit` via `NeedsPaging`. So each page request re-runs the full search and hydration; that was true before this PR too, but now up to eight of them run concurrently per user. It's probably worth confirming that's safe for smaller implementations.

Related: since `totalCount=true` only serializes the size of the already loaded list on this path, dropping it from later pages doesn't actually save a count query. The optimization is harmless, but can we reword the comment in `buildUrl` so it doesn't assert one?

denniskigen (commented):
This is close, @pirupius. The paging tests are useful. I've left one pagination issue and a couple of follow-ups.

packages/esm-patient-search-app/src/patient-search-page/patient-search-lg.component.tsx
```diff
@@ -33,8 +33,15 @@ const PatientSearchComponent: React.FC<PatientSearchComponentProps> = ({
...
+  // as the infinite search appends server pages. Depending on `goTo` alone would therefore snap the
+  // user back to page 1 every time another page of results lands. Reset only on an actual new query.
+  const previousQueryRef = useRef(query);
   useEffect(() => {
-    goTo(1);
+    if (previousQueryRef.current !== query) {
+      previousQueryRef.current = query;
+      goTo(1);
+    }
   }, [query, goTo]);
```
denniskigen: `usePagination` doesn't clamp its page when the data shrinks. If I'm on page 3 and apply a refine filter that leaves 10 matches, this keeps page 3 and shows the empty state under a "10 search results" header instead of those patients. I reproduced this both in a focused test and in the running app. This is the [same failure class as #684](https://github.com/openmrs/openmrs-esm-patient-management/pull/684#issuecomment-1531374228), but with a different trigger - that repro changed the query, while this one changes the filtered result set under the same query. Tracking decreases keeps the selected page while results are appended, but resets it when filtering invalidates the page:

```suggestion
  // `goTo` from `usePagination` is rebuilt whenever `totalPages` changes, and `totalPages` grows
  // as the infinite search appends server pages. Depending on `goTo` alone would therefore snap the
  // user back to page 1 every time another page of results lands. Reset on a new query, or when the
  // result set shrinks under the same query (e.g. a refine filter) and the page may be out of range.
  const previousQueryRef = useRef(query);
  const previousTotalPagesRef = useRef(totalPages);
  useEffect(() => {
    const queryChanged = previousQueryRef.current !== query;
    const resultSetShrank = previousTotalPagesRef.current > totalPages;
    previousQueryRef.current = query;
    previousTotalPagesRef.current = totalPages;
    if (queryChanged || resultSetShrank) {
      goTo(1);
    }
  }, [query, totalPages, goTo]);
```

Repro: search a term with 41+ results, go to page 3, apply a sex filter that leaves under 21 matches.

packages/esm-patient-search-app/src/patient-search.resource.tsx
```diff
@@ -179,9 +200,9 @@ export function useInfinitePatientSearch(
       isValidating,
       setPage: setSize,
       currentPage: size,
-      totalResults: shouldFetch ? (data?.[0]?.data?.totalCount ?? 0) : 0,
+      totalResults: totalCount ?? 0,
```
denniskigen: nit: `keepPreviousData` keeps the previous query's rows until the new first page resolves, but this makes `totalResults` zero during that interval, so the compact search shows those retained patients under "0 search results". It's probably better to keep a separate current-query count for sizing the page requests, while `totalResults` continues to describe the returned `data`. This is limited to the compact banner for one round trip, so it can be ignored if we'd rather handle it separately.

packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
```diff
@@ -39,16 +48,26 @@ const AdvancedPatientSearchComponent: React.FC<AdvancedPatientSearchProps> = ({
...
+  const pagesNeeded = Math.ceil(totalResults / resultsPerPage);
 
   useEffect(() => {
-    if (searchResults?.length === currentPage * 50 && hasMore) {
-      setPage((page) => page + 1);
+    if (!isValidating && pagesNeeded > currentPage) {
+      setPage(Math.min(currentPage + pagesPerWave, pagesNeeded));
     }
-  }, [searchResults, currentPage, hasMore, setPage]);
+  }, [isValidating, pagesNeeded, currentPage, setPage]);
```
denniskigen:
```suggestion
    if (!fetchError && !isValidating && pagesNeeded > currentPage) {
      setPage(Math.min(currentPage + pagesPerWave, pagesNeeded));
    }
  }, [fetchError, isValidating, pagesNeeded, currentPage, setPage]);
```

If one request in a wave fails, `isValidating` becomes false while `pagesNeeded` is still ahead, so we keep opening later waves even though the UI has switched to `ErrorState`. I reproduced this with SWR retries disabled: a failure at `startIndex=50` still requested `startIndex=450` and later pages, and overlapping waves exceeded the eight-request bound. Can we add this error path to the paging test?

denniskigen:
Heads up: https://github.com/openmrs/openmrs-esm-core/pull/1856 stands merged. Sorry for the mixup.

packages/esm-patient-search-app/src/patient-search.resource.tsx
pirupius: This is valid point especially for smaller implementations, i've reduced this to 4 to reduce the effect

denniskigen (approved):
LGTM. Thanks, @pirupius!

PR #2710, opened 2026-09-01
(fix) O3-5940: Don't offer queues a patient is already in

UjjawalPrabhat:
## Requirements

## Summary
Adding a patient to a queue they were already in was possible all the way to submit, where the backend rejected it with a `400`. The queue entry form now leaves those queues out of the Service options, and leaves a location out of Queue Location once the patient is in every queue there.

### Before
image

### After 
image

## Related Issue
https://openmrs.atlassian.net/browse/O3-5940


packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -46,14 +47,32 @@ const createQueueServiceSchema = (t: TFunction) =>
...
+  // along with any location where they are in every queue. A location with no queues at all is kept,
+  // so that its misconfiguration still surfaces.
+  const { queueEntries, isLoading: isLoadingQueueEntries } = useQueueEntries(
+    { patient: patientUuid, isEnded: false },
+    'custom:(uuid,queue:(uuid))',
+    Boolean(patientUuid),
+  );
+  const activeQueueUuids = useMemo(() => new Set(queueEntries.map((entry) => entry.queue?.uuid)), [queueEntries]);
+
   const memoizedQueueLocations = useMemo(
```
NethmiRodrigo: Let's keep the location list as it was and let the `Service` field carry the filtering on its own. When the patient is in every queue at the location, the select would be empty and the `helperText` would say so underneath. That removes the second `useQueues()` call, this filter, the location reset effect below, and the "already in every queue at the available locations" notice. The service skeleton should still wait on `isLoadingQueueEntries` so the options do not disappear after they render.
```
const isPatientInEveryQueue = queues.length > 0 && memoizedQueues.length === 0;
```
In the `Service`dropdown field, the misconfiguration error notice should be rendered based on the unfiltered queue list. Also, if the dropdown has no options to display (because the patient is in every queue), the `helperText` would be displayed but would get replaced by the `required` error as soon as the user clicks out of the dropdown so the invalid boolean needs to account for that condition as well:
```
 ) : !queues.length ? (
   <InlineNotification ... />
 ) : (
   <Select
     {...field}
     labelText=""
     id="queueService"
     helperText={
       isPatientInEveryQueue
         ? t('patientInEveryQueueAtLocation', 'This patient is already in every queue at this location')
         : undefined
     }
     invalid={!!errors.queueService && !isPatientInEveryQueue}
     invalidText={errors.queueService?.message}
```

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
NethmiRodrigo: To avoid the priority options rendering before the queue services
```suggestion
              isLoadingServices ? (
```

NethmiRodrigo: If the backend detects a duplicate queue entry, the response looks like this -
```
{"error":{"message":"Invalid Submission","code":"webservices.rest.error.invalid.submission","globalErrors":[{"code":"queue.entry.error.duplicate","message":"This patient is already in this queue"}],"fieldErrors":{}}}
```
The code `queue.entry.error.duplicate` is in a key called `globalErrors`. 
The frontend looks for this error code like - 
```
const errorMessage = error?.responseBody?.error?.message || error?.message || '';
const isDuplicatePatientError = errorMessage.includes(DUPLICATE_QUEUE_ENTRY_ERROR_CODE);
```
But, `error.responseBody.error.message` is "Invalid Submission", so the correct snackbar isn't what renders.

`queue-entry-error.utils.ts` already has `isDuplicateQueueEntryError`, so let's use that here instead:

```suggestion
          if (isDuplicateQueueEntryError(error)) {
```

You would have to change `DUPLICATE_QUEUE_ENTRY_ERROR_CODE` to `queue.entry.error.duplicate` in constants.ts, and in `getErrorMessage` in `queue-entry-error.utils.ts`:
```
const globalErrorCodes = (err?.responseBody?.error?.globalErrors ?? []).map((globalError) => globalError.code);
 return [message, ...globalErrorCodes].filter(Boolean).join(' ');
```

The test in `queue-entry-actions.test.tsx` at line 208 would also need to be updated.

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.test.tsx
```diff
@@ -13,42 +13,63 @@ import {
...
-    }),
-  };
-});
+vi.mock('../../hooks/useQueues', () => ({
+  useQueues: vi.fn(),
+}));
+
+vi.mock('../../hooks/useQueueEntries', () => ({
+  useQueueEntries: vi.fn(),
+  useMutateQueueEntries: vi.fn(() => ({ mutateQueueEntries: vi.fn() })),
```
NethmiRodrigo: Let's define a `mockMutateQueueEntries` and share it like [edit-queue-entry.test.tsx does](https://github.com/openmrs/openmrs-esm-patient-management/blob/ccad2560ce5783f5b4613404f56fb06b6260ea3a/packages/esm-service-queues-app/src/modals/edit-queue-entry.test.tsx#L10-L19), and add a test where `postQueueEntry` rejects with `{ responseBody: { error: { message: 'Invalid Submission', globalErrors: [{ code: 'queue.entry.error.duplicate' }] } } }` and we assert the warning snackbar and one `mutateQueueEntries` call.

```suggestion
  useMutateQueueEntries: () => ({ mutateQueueEntries: mockMutateQueueEntries }),
```

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat, but getting the list of all queues just to filter out the locations seems a bit of an unnecessary complexity so I added an alternative suggestion

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -18,13 +18,14 @@ import { zodResolver } from '@hookform/resolvers/zod';
...
-import { useMutateQueueEntries } from '../../hooks/useQueueEntries';
+import { useMutateQueueEntries, useQueueEntries } from '../../hooks/useQueueEntries';
 import { useQueueLocations } from '../hooks/useQueueLocations';
 import { useQueues } from '../../hooks/useQueues';
-import { DUPLICATE_QUEUE_ENTRY_ERROR_CODE } from '../../constants';
+import { isDuplicateQueueEntryError } from '../../modals/queue-entry-error.utils';
 import { useServiceQueuesStore } from '../../store/store';
 
 export interface QueueFieldsProps {
+  patientUuid: string;
```
NethmiRodrigo: Shouldn't this be marked as optional?
```suggestion
  patientUuid?: string;
```

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -46,15 +47,21 @@ const createQueueServiceSchema = (t: TFunction) =>
  * This component contains form fields for starting a patient's queue entry.
  */
 
-const QueueFields = React.memo(({ setOnSubmit, defaultInitialServiceQueue }: QueueFieldsProps) => {
+const QueueFields = React.memo(({ patientUuid, setOnSubmit, defaultInitialServiceQueue }: QueueFieldsProps) => {
   const { t } = useTranslation();
   const schema = useMemo(() => createQueueServiceSchema(t), [t]);
   const { sessionLocation } = useSession();
   const { queueLocations, isLoading: isLoadingQueueLocations } = useQueueLocations();
-  const memoizedQueueLocations = useMemo(
```
NethmiRodrigo, line 54: Why did we get rid of the memoized queue locations?

packages/esm-service-queues-app/src/modals/queue-entry-error.utils.ts
```diff
@@ -7,18 +7,24 @@ export function getErrorMessage(error: unknown): string {
...
     err?.message ||
-    ''
-  );
+    '';
+
+  // Validation failures carry a generic "Invalid Submission" message, with the code identifying the
+  // actual problem in `globalErrors`, so those codes are appended for the callers below to match on.
+  const globalErrorCodes = (err?.responseBody?.error?.globalErrors ?? []).map((globalError) => globalError.code);
+
+  return [message, ...globalErrorCodes].filter(Boolean).join(' ');
```
NethmiRodrigo: This string is also the snackbar subtitle in queue-entry-confirm-action.modal.tsx, so the user would read the raw code. Let's return only the message here and have `isDuplicateQueueEntryError` check the globalErrors codes directly.
```suggestion
return message;
   }
  
   function getGlobalErrorCodes(error: unknown): string[] {
     const err = error as { responseBody?: { error?: { globalErrors?: Array<{ code?: string }> } } };
     return (err?.responseBody?.error?.globalErrors ?? []).map((globalError) => globalError.code).filter(Boolean);
   }
```
and in `isDuplicateQueueEntryError`:
```
   return getGlobalErrorCodes(error).includes(DUPLICATE_QUEUE_ENTRY_ERROR_CODE);
```

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -248,7 +262,12 @@ const QueueFields = React.memo(({ setOnSubmit, defaultInitialServiceQueue }: Que
...
                   labelText=""
                   id="queueService"
-                  invalid={!!errors.queueService}
+                  helperText={
+                    isPatientInEveryQueue
+                      ? t('patientInEveryQueueAtLocation', 'This patient is already in every queue at this location')
+                      : undefined
+                  }
+                  invalid={!!errors.queueService && !isPatientInEveryQueue}
                   invalidText={errors.queueService?.message}
```
NethmiRodrigo: This way, we can avoid errors being hidden if there are no patients in queue
```suggestion
                  invalid={!!errors.queueService}
                  invalidText={
                       isPatientInEveryQueue
                         ? t('patientInEveryQueueAtLocation', 'This patient is already in every queue at this location')
                         : errors.queueService?.message
                     }
```

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
UjjawalPrabhat: So please correct me if I'm wrong but from what I understood was that it's dependency is `queueLocations`, which comes from [useQueueLocations](https://github.com/openmrs/openmrs-esm-patient-management/blob/18fbddf63da0df9b17e6f9c7daca6aa0a728e0e1/packages/esm-service-queues-app/src/create-queue-entry/hooks/useQueueLocations.ts#L8-L11) and that hook's own `useMemo` keys off the `data` it gets from `useFhirFetchAll`, which [returns data: data?.flatMap((d) => getData(d.data))](https://github.com/openmrs/openmrs-esm-core/blob/main/packages/framework/esm-react-utils/src/useOpenmrsInfinite.ts#L107) without memoising it. So data is a new array every render, queueLocations is a new array every render and the memo's dependency check never matched — it recomputed every time and never once returned a cached value. Nothing reads its identity either; it's only mapped into `SelectItems` in the dropdown, so a stable reference wouldn't help even if it were stable.

That being said what are your thoughts?

NethmiRodrigo: Aaah, yeah I think you're right!

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -46,15 +47,20 @@ const createQueueServiceSchema = (t: TFunction) =>
...
-  const memoizedQueueLocations = useMemo(
-    () => queueLocations.map((l) => ({ id: l.id, name: l.name })),
-    [queueLocations],
+
+  // A patient cannot be added to a queue they are already in, so those queues are left out below
+  const { queueEntries, isLoading: isLoadingQueueEntries } = useQueueEntries(
+    { patient: patientUuid, isEnded: false },
+    'custom:(uuid,queue:(uuid))',
+    Boolean(patientUuid),
   );
```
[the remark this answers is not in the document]
NethmiRodrigo: The backend already has a guard to prevent duplicate queue entry creations, so this is the for-convenience frontend guard. The backend `QueueEntryValidator` rejects a duplicate entry with the `queue.entry.error.duplicate code`. The submit handler in this PR already catches that code, shows the "Patient already in queue" warning, and refreshes the entries. The service filter dropdown is also rendered only after the queue lookup finishes loading, so the user can't select an active queue entry before.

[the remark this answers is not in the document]
NethmiRodrigo: Can't believe I'm arguing with a bot, but, fine... Added an inline notification if there is an error. Hope you're happy now. Don't make me turn you off

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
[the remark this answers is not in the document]
UjjawalPrabhat: @NethmiRodrigo You should have fought back....this is gonna be memorable. Hahahaha!!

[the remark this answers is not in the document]
NethmiRodrigo: I saw the reply and said "Ugh" in real life, but didn't want to waste tokens on it, lol

PR #2716, opened 2026-09-03
(test) Assert the search paging tests through what the user sees

denniskigen:
## Requirements

## Summary

Follow up to #2708. The paging tests it added stub the result views and then assert through the stubs: which uuid slice got passed to `PatientSearchResults`, and test ids for the empty and error states. This reworks them so the assertions read what a user would see.

- The stub for `PatientSearchResults` now renders each patient's name, and the empty, error and loading states are the real components. Assertions are `getByText('Patient 20')` and the real "Sorry, no patient charts were found" and error copy. The stub stays because the framework mock hides the banner's contents, so a real banner can't show a name in tests.
- The 1700-result test now proves the last page is reachable the way a user would: filter by a postcode only the last patient has and expect that patient. Before, it only counted requests.
- The wave and error tests no longer copy `pagesPerWave` from the component, so tuning it doesn't break them. The wave test bounds concurrency between one page and the whole set, and the error test asserts no request follows the error state.
- The paging file runs under jsdom, same as `advanced-patient-search.test.tsx` and for the same reason: the refine-search form submit doesn't fire under happy-dom.

No behaviour changes. The paging file takes roughly 200ms longer, mostly the filter step and jsdom.

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search-paging.test.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search-paging.test.tsx
@@ -1,19 +1,31 @@
+/**
+ * @vitest-environment jsdom
+ *
+ * The refine-search form submit does not fire its callback under happy-dom (see
+ * advanced-patient-search.test.tsx). Run this file under jsdom.
+ */
 import React from 'react';
 import { vi, describe, it, expect, beforeEach } from 'vitest';
 import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { SWRConfig } from 'swr';
 import { type FetchResponse, getDefaultsFromConfigSchema, openmrsFetch, useConfig } from '@openmrs/esm-framework';
+import { type SearchedPatient } from '../types';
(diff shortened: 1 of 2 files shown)


packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search-paging.test.tsx
```diff
@@ -30,29 +42,29 @@ const noRetryWrapper = ({ children }: { children: React.ReactNode }) => (
...
 
-const startIndexesOf = (urls: Array<string>) =>
-  urls.map((url) => Number(new URL(url, 'http://localhost').searchParams.get('startIndex') ?? 0));
-
 // A big clinic's result set: far more pages than one wave, and not a whole number of pages.
 const TOTAL = 1700;
 const PAGE_SIZE = 50;
-// Mirrors `pagesPerWave` in the component: how many requests may be in flight together.
-const PAGES_PER_WAVE = 4;
+const TOTAL_PAGES = Math.ceil(TOTAL / PAGE_SIZE);
```
pirupius, line 48: 👍🏾

pirupius (approved):

PR #2717, opened 2026-09-03
(fix) Match the birthdate filters against the stored date and trim the postcode filter

denniskigen:
## Requirements

## Summary

This PR applies two fixes to how the refine search filters match patients.

The date of birth filters parsed each patient's birthdate with `new Date` and read the local day, month and year. The backend serialises birthdates as midnight in its own timezone (`1940-01-01T00:00:00.000+0000`), so in a browser west of that timezone every patient shifts to the previous day and misses the filters, and a patient with no recorded birthdate parses as 1 January 1970 and matches filters for that date. The filters now read the calendar date off the string, and a patient without a birthdate matches no date of birth filter.

The postcode filter compared the typed value exactly, so a stray leading or trailing space matched nothing. It's trimmed when the filters are applied, which also keeps the applied-filter count honest for a whitespace-only value.

Tests cover a null birthdate against a 1 January 1970 filter, a birthdate serialised with a non-UTC offset (the suite runs in UTC, where `new Date` reads it as the previous day), and a padded postcode.

#### Before

dob1970filterbefore

#### After
dob1970filterafter

#### Before
postcodetrailingspacebefore

#### After
postcodetrailingspaceafter

### DIFF
--- a/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
+++ b/packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
@@ -23,6 +23,20 @@ const resultsPerPage = 50;
  */
 const pagesPerWave = 4;
 
+/**
+ * Reads the calendar date off a REST birthdate. The backend serialises birthdates as midnight in
+ * its own timezone (for example `1940-01-01T00:00:00.000+0000`), so parsing them with `Date` would
+ * shift the day, month and year for a browser in a different timezone, and would turn a missing
+ * birthdate into 1 January 1970.
+ */
+function parseBirthdate(birthdate: string | null | undefined) {
+  const match = /^(\d{4})-(\d{2})-(\d{2})/.exec(birthdate ?? '');
+  if (!match) {
+    return null;
+  }
+  return { year: Number(match[1]), month: Number(match[2]), day: Number(match[3]) };
+}
+
 interface AdvancedPatientSearchProps {
   query: string;
   inTabletOrOverlay?: boolean;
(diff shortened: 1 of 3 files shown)


packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
```diff
@@ -23,6 +23,20 @@ const resultsPerPage = 50;
...
+ * shift the day, month and year for a browser in a different timezone, and would turn a missing
+ * birthdate into 1 January 1970.
+ */
+function parseBirthdate(birthdate: string | null | undefined) {
+  const match = /^(\d{4})-(\d{2})-(\d{2})/.exec(birthdate ?? '');
+  if (!match) {
+    return null;
+  }
+  return { year: Number(match[1]), month: Number(match[2]), day: Number(match[3]) };
+}
```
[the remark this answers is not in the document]
denniskigen: This can't happen with the OpenMRS backend. `Person.birthdate` is a `java.util.Date` on a `DATE` column, so the REST payload only ever carries real calendar dates.

packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
```diff
@@ -23,6 +23,20 @@ const resultsPerPage = 50;
  */
 const pagesPerWave = 4;
 
+/**
+ * Reads the calendar date off a REST birthdate. The backend serialises birthdates as midnight in
+ * its own timezone (for example `1940-01-01T00:00:00.000+0000`), so parsing them with `Date` would
+ * shift the day, month and year for a browser in a different timezone, and would turn a missing
+ * birthdate into 1 January 1970.
```
pirupius, line 30: 👍🏾

packages/esm-patient-search-app/src/patient-search-page/advanced-patient-search.component.tsx
```diff
@@ -23,6 +23,20 @@ const resultsPerPage = 50;
...
+ */
+function parseBirthdate(birthdate: string | null | undefined) {
```
pirupius, line 32: Have u tested this behaviour with approximated dob?

ibacher: This will work perfectly with approximated dates. Approximated dates are actually stored as exact date/times with a boolean flag. The FHIR2 module makes some assumptions about how to truncate longer dates, but that's simply because that's the standard way to communicate "we don't know what date exactly".

ibacher (approved):

PR #2718, opened 2026-09-03
(fix) Keep the header search input in step with the query in the URL

denniskigen:
## Requirements

## Summary

On the search page the header search input is seeded from the query in the URL, but only once, when it mounts. A browser back or forward, or a link to another query, changes the results and the heading while the input keeps showing the previous term. The compact search and the search bar now follow the term they're given whenever it changes.

Repro on main: search for one term, then another, then press the browser back button. The results go back to the first term while the input still shows the second.

#### Before
headerinputafterbackbefore

#### After
headerinputafterbackafter


pirupius (approved):
Nice catch @denniskigen

PR #1362, opened 2024-11-07
(feat) O3-4180: Configure Advanced search filter to support person attributes

usamaidrsk:
## Requirements

- [x] This PR has a title that briefly describes the work done including the ticket number. If there is a ticket, make sure your PR title includes a [conventional commit](https://o3-docs.openmrs.org/docs/frontend-modules/contributing.en-US#contributing-guidelines) label. See existing PR titles for inspiration.
- [ ] My work conforms to the [OpenMRS 3.0 Styleguide](https://om.rs/styleguide) and [design documentation](https://om.rs/o3ui).
- [x] My work includes tests or is validated by existing tests.

## Summary

This PR enhances the patient search functionality by adding support for advanced filtering based on person attributes. It now enables filtering by multiple attribute types, including strings, concepts, and locations, for more refined and comprehensive search results. Additionally, this update introduces configuration to enable or disable specific fields such as date of birth, postal code, gender, and age, allowing for greater flexibility in search customization.

#### Testing Config
```typescript
{
  "@openmrs/esm-patient-search-app": {
    "search": {
      "patientChartUrl": "${openmrsSpaBase}/patient/${patientUuid}/chart/",
      "showRecentlySearchedPatients": true,
      "disableTabletSearchOnKeyUp": false,
      "searchFields": {
        "fields": {
          "gender": {
            "enabled": true
          },
          "dateOfBirth": {
            "enabled": true
          },
          "age": {
            "enabled": true
          },
          "postcode": {
            "enabled": true
          }
        },
        "personAttributes": [
          {
            "attributeTypeUuid": "14d4f066-15f5-102d-96e4-000c29c2a5d7"
          },
          {
            "attributeTypeUuid": "8d87236c-c2cc-11de-8d13-0010c6dffd0f"
          }
        ]
      }
    },
    "includeDead": true,
    "contactAttributeType": [
      "14d4f066-15f5-102d-96e4-000c29c2a5d7",
      "e3d177ee-04ad-11ed-828d-0242ac1e0002"
    ],
    "defaultIdentifierTypes": [
      "05a29f94-c0ed-11e2-94be-8c13b969e334"
    ]
  }
}
``` 

## Screenshots

image

https://github.com/user-attachments/assets/b13f9f3d-984e-438a-b11c-c0b9d7f5e5ec

## Related Issue

https://openmrs.atlassian.net/browse/O3-4180

PR #2720, opened 2026-09-03
(fix) Apply the configured age bounds to the refine search age field

denniskigen:
## Requirements

## Summary

The config schema has offered `searchFilterFields.age.min` and `max` since the person attribute filters landed in #1362, and the age field reads them, but neither field builder ever passed them through. The built-in fields were created with only a name and a type, so setting the values did nothing. Both the desktop panel and the tablet dialog now hand the configured bounds to the field.

A maximum of 0, the schema default, means no maximum. Carbon marks any value above `max` invalid, so passing 0 through would have flagged every age. The schema description says so now.

Carbon only draws the invalid border and icon for a value outside the bounds, so the field now also says which range is allowed: "Enter an age between 18 and 65", or "Enter an age of at least 0" when no maximum is set. Two new strings for Transifex: `ageOutOfRange` and `ageBelowMinimum`.

Tests cover configured bounds on desktop and in the tablet dialog, the unbounded default, and the message for an out-of-range value.

For context, the implementer configs I could find on GitHub (MSF's LIME-EMR base and site configs, CRUDEM's Ozone HSC) all set `age.min: 0` and none set `max`, so today's practical effect is that a negative age gets flagged for them.

#### Before
age70withmax65before

#### After
age70withmax65after

### DIFF
--- a/packages/esm-patient-search-app/src/config-schema.ts
+++ b/packages/esm-patient-search-app/src/config-schema.ts
@@ -51,7 +51,7 @@ export const configSchema = {
         max: {
           _type: Type.Number,
           _default: 0,
-          _description: 'The maximum value for the age field',
+          _description: 'The maximum value for the age field. 0 means no maximum.',
         },
       },
       postcode: {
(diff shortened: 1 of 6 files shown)


NethmiRodrigo (approved):
LGTM!

PR #2721, opened 2026-09-03
(fix) O3-5917: Keep newly added relationship rows correctly associated

denniskigen:
## Requirements

## Summary

New relationship rows don't have a server uuid, so after #2383 switched the row keys from the index to the uuid, every new row shared the same undefined key and React reconciled them by position. The relationship type select was also uncontrolled, so removing one of two new rows left the surviving row showing the removed row's person and type while Formik held the other row's values.

This gives each new relationship a client-side id when it's pushed, keys the rows on the server uuid or that id, and makes the select controlled by its Formik value. The id never reaches the server since `saveRelationships` only picks the fields it needs.

The section test's `fetchPerson` mock returned a response object where the real function returns an array, so nothing could exercise the person search. Fixed that, added an interaction test that fills two rows, removes the first and checks the second, and made the file fail if any test logs a React key warning.

## Screenshots

Both runs enter Joshua Johnson / Doctor in row 1 and Richard Jones / Patient (Doctor) in row 2, then delete row 1.

### Before: the surviving row still shows Joshua Johnson / Doctor.

before1tworows

before2afterremovingfirstrow

### After: it shows Richard Jones / Patient (Doctor).

after1tworows

after2afterremovingfirstrow

## Related Issue

https://openmrs.atlassian.net/browse/O3-5917

## Other

Registrations queued offline by a build before this change would still rehydrate their new rows without a client id. That's a narrow upgrade edge, so I've left it out here.

Not touched here either: the related person's name for a new row lives only in the Autosuggest DOM, since `relatedPersonName` is never written to Formik. That's pre-existing.

### DIFF
--- a/packages/esm-patient-registration-app/src/patient-registration/patient-registration.types.ts
+++ b/packages/esm-patient-registration-app/src/patient-registration/patient-registration.types.ts
@@ -153,6 +153,11 @@ export interface RelationshipValue {
    */
   initialrelationshipTypeValue?: string;
   uuid?: string;
+  /**
+   * Client-side identifier for a relationship that has not been saved yet.
+   * Used as a stable React key until the server assigns a uuid. Never sent to the server.
+   */
+  clientId?: string;
 }
 
 export interface FormValues {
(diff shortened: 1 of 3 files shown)


NethmiRodrigo (approved):
Yikes, how was this never caught? Thanks @denniskigen!

PR #2725, opened 2026-09-04
(fix) Tell the user when the refine filters, not the query, emptied the search results

denniskigen:
## Requirements

## Summary

When a refine filter matches nothing, the results area shows the same message as a query with no hits: "Sorry, no patient charts were found", with a hint to try again with the patient's ID number. That points the user at the wrong fix. The query found patients; a filter hid them.

When the query found patients and the filters removed all of them, the empty state now reads "No patients match these filters", with the hint "Remove or change a filter to see more patients". A query that itself found nothing keeps the original message. This follows the pattern the queues, active visits, patient lists and ward apps already use for a filter that hides everything, minus the "above", since on the search page the filters sit beside the results on desktop and below them on tablet.

Two new strings for Transifex: `noPatientsMatchFilters` and `adjustFiltersHint`.

#### Before

filtersmatchnothingbefore

#### After

filtersmatchnothingafter


chibongho (commented):
Code is fine, but do we want to make the distinction between the free-from-string query vs other filters?

There was talk, a long time ago, about having the filtering done on server-side instead of client-side. If we still want to do that at some point, It'd be harder to tell whether an empty result was because of the query or the filters.

Can we just tell the user to refine both the filter and the query when they see empty search results?

packages/esm-patient-search-app/src/patient-search-page/patient-search-views.component.tsx
```diff
@@ -10,17 +10,24 @@ interface PatientSearchResultsProps {
   searchResults: SearchedPatient[];
 }
 
-export const EmptyState: React.FC = () => {
+interface EmptyStateProps {
+  title?: string;
+  hint?: string;
+}
```
ibacher, line 18: Probably a good idea to call out that we expect these to be translated strings.

ibacher:
> about having the filtering done on server-side instead of client-side

Still ultimately a goal. The impasse is: we can't filter by the various secondary attributes (gender, etc.) via the REST API and the FHIR API doesn't support the filters by patient attributes. If we can make one of those trade-offs we could move this server-side.

denniskigen:
Fair question. I'd lean towards keeping the distinction while we already have the unfiltered results on the client, since it tells the user what to change. If filtering moves server-side, we can revisit the message based on what the endpoint returns.

ibacher (approved):

ibacher:
Thanks Dennis! I think this looks like a real improvement!

PR #2731, opened 2026-09-08
(fix) O3-5958: Default the queue location to the one selected in Service Queues

Muta-Jonathan:
## Requirements

## Summary
Fixes an issue where the "Add to Queue" form's location field remains blank if the user's immediate login location is not queue-tagged.

## Description
Previously, the `queueLocation` field defaulted strictly to `sessionLocation?.uuid`. If a user logged into a location that was not explicitly queue-tagged, the `Select` component would fail to find a matching option and render as blank. 
However, the Service Queues dashboard header already correctly resolves the nearest queue-tagged ancestor and stores it in `useServiceQueuesStore`. 
This PR updates the form's `defaultValues` to leverage the store's resolved location first (`selectedQueueLocationUuid`), falling back to the session location. This ensures the form pre-fills correctly with the resolved ancestor location.

## How to test
1. Log into the EMR with a session location that is **not** a queue location, but has a parent/ancestor that is queue-tagged.
2. Navigate to the Service Queues app and click to add a patient to a queue.
3. Verify that the "Queue Location" field is automatically pre-filled with the queue-tagged ancestor location (the same one displayed in the dashboard header), rather than being blank.

## Before

https://github.com/user-attachments/assets/e79852cd-631f-48c6-afc5-9497509bd677

## After

https://github.com/user-attachments/assets/3d22fc7c-2ebd-4aa9-b7be-06ddda05389c

## Related Issue
https://openmrs.atlassian.net/browse/O3-5958


Muta-Jonathan:
cc @denniskigen @NethmiRodrigo @VeronicaMuthee this is a minor bug we found in service queues hope it makes sense for the fix

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -165,6 +165,20 @@ const QueueFields = React.memo(({ patientUuid, setOnSubmit, defaultInitialServic
     setOnSubmit?.(onSubmit);
   }, [onSubmit, setOnSubmit]);
 
+  useEffect(() => {
```
dkayiwa: I think it's worth settling whether this is the right place for the fix, because the service queues app already resolves a default queue location and this effect re-derives a weaker version of it.

`queue-location-header-title.component.tsx` works out "the login location if queue-tagged, else its nearest queue-tagged ancestor, else All" and persists it through `updateSelectedQueueLocationUuid`. `checked-in-patients.component.tsx:100`, in this same folder, reads it back as `selectedQueueLocationUuid ?? session?.sessionLocation?.uuid`, and the `useForm` call above already defaults `queueService` from that same store. So the whole thing reduces to one line in `defaultValues`:

```js
queueLocation: selectedQueueLocationUuid ?? sessionLocation?.uuid ?? '',
```

The store is backed by sessionStorage, so that value is readable at mount and no effect is needed at all. (In the patient chart visit form path the header isn't mounted, so a stored value can be left over from an earlier dashboard visit, and the session fallback covers there being nothing stored.)

The header's version also does three things this effect doesn't, and the first is why I doubt the ticket is actually fixed here: it walks up to the nearest queue-tagged ancestor. A login location that isn't itself tagged `queue location` is a common setup, and in that case the header shows the resolved ancestor while this form shows nothing. I checked what this patch does in that situation and the field is still blank; the only change is that the Service group stops rendering. If that is what the reporter is hitting, O3-5958 gets closed with their field still empty. The other two things it does (resolve once per login location, and skip while loading or on error) are the two comments below.

I also couldn't reproduce the mechanism in the description. `useSession()` doesn't go through SWR and never returns an unloaded session: it either returns a `Session` or throws for Suspense (`esm-react-utils/src/useSession.ts`), and `openmrsComponentDecorator` wraps every component in a `<Suspense>` boundary, so `sessionLocation` is already present by the time `useForm` evaluates `defaultValues`. That has been true since esm-core #797 in October 2023. Mounting the component against main with `useQueueLocations` still loading and then letting it resolve, which is the real sequence, gives a field already prefilled with the session location.

Could you add a test that fails on main and passes with this change? That would pin down the repro and cover the change at the same time. `queue-fields.test.tsx` already has the mocks for it, and the full service-queues suite (191 tests) passes unchanged with this patch applied, so nothing exercises it right now.

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -165,6 +165,20 @@ const QueueFields = React.memo(({ patientUuid, setOnSubmit, defaultInitialServic
...
+    if (!isLoadingQueueLocations) {
+      if (queueLocation) {
+        if (!queueLocations.some((l) => l.id === queueLocation)) {
+          setValue('queueLocation', '');
+        }
```
dkayiwa: This branch has no way of knowing when the locations list can't be trusted. `useQueueLocations` goes through `useServerFetchAll`, which returns `data: undefined` whenever `error` is set, so `queueLocations` becomes `[]` with `isLoading` still false on any failed refresh. When that happens this clears the location the user picked, and since `queueLocation` is then falsy the effect on line 209 clears their service and priority too.

Then the branch just below refills the field with the *session* location instead of the one they chose. I stepped it through: pick "Other Ward" and a service, let one locations refresh fail, let the next one succeed, and you land on the session location with an empty service. Nothing is surfaced to the user, so if they simply reselect a service they file the queue entry at the wrong location. Main keeps both selections through the same sequence.

The refresh isn't hypothetical. `queue-location-header-title` calls `useQueueLocations` too, so the key is already in the SWR cache when the Add patient to queue modal opens over the queues dashboard, and `revalidateIfStale` (true in the framework's default SWR config) revalidates it on mount. Only that request has to fail.

The hook already returns `error`, so destructuring it on line 54 and requiring it here is enough:

```js
const { queueLocations, isLoading: isLoadingQueueLocations, error: queueLocationsError } = useQueueLocations();
...
if (!isLoadingQueueLocations && !queueLocationsError) {
```

which is the guard `queue-location-header-title` uses, for the same reason it gives there: a degraded resolution must not be persisted.

packages/esm-service-queues-app/src/create-queue-entry/queue-fields/queue-fields.component.tsx
```diff
@@ -165,6 +165,20 @@ const QueueFields = React.memo(({ patientUuid, setOnSubmit, defaultInitialServic
...
+      } else if (sessionLocation?.uuid) {
+        if (queueLocations.some((l) => l.id === sessionLocation.uuid)) {
+          setValue('queueLocation', sessionLocation.uuid);
+        }
```
dkayiwa: This reapplies the default after the user has changed the field, which makes the dropdown's own "Select a queue location" option impossible to pick.

Choosing it fires `field.onChange('')`, which clears `queueService` and `priority` (the Select's `onChange` does it, and so does the effect on line 209), and then this branch immediately puts the session location back. I ran it: after choosing the blank option the location is back to the session location's uuid and the service is empty, so the click quietly resets the service and priority instead of clearing the location. On main the same click clears the location and hides the Service group, which is what the markup is written for.

Worth knowing that `sessionLocation` is in the dependency array and `handleSessionResponse` stores the freshly parsed session object on every refetch with no equality check, so this effect re-runs on its own too, not only when the user touches the field.

`queue-location-header-title` guards the same thing with `LOGIN_LOCATION_STORAGE_KEY`, commented "so a manual selection sticks". A ref does it here:

```suggestion
      } else if (!hasDefaultedLocation.current && sessionLocation?.uuid) {
        if (queueLocations.some((l) => l.id === sessionLocation.uuid)) {
          hasDefaultedLocation.current = true;
          setValue('queueLocation', sessionLocation.uuid);
        }
```

with `const hasDefaultedLocation = useRef(false);` alongside the other hooks.

Muta-Jonathan:
Thanks @dkayiwa for the thorough review and for catching those issues with the useEffect. You were spot on.

I looked into it again and realized my initial assumption about sessionLocation loading asynchronously was wrong, since the `<Suspense>` boundary handles that before the form even renders.

You were also right about the real root cause of O3-5958: it was happening because the form would stay blank if a user logged into a location that wasn't queue-tagged itself, instead of resolving to the nearest queue-tagged ancestor.

I've just pushed an update that drops the `useEffect` entirely and goes with your suggestion to use `selectedQueueLocationUuid` directly in the defaultValues. I also added the tests you requested in `queue-fields.test.tsx` to verify the pre-fill behavior.

Let me know if this looks good to go!

dkayiwa (commented):
The description and title still describe the `useEffect` version. The description walks through an effect that watches `sessionLocation` and `queueLocations` and clears the field when a stored location is no longer valid, and it opens with the SWR diagnosis that turned out not to be the cause. None of that is in the diff now, and "asynchronously" in the title says the opposite of what the change does, since reading the store works precisely because the value is already there at mount. This repo squash-merges with the PR title as the commit subject, and O3-5958 points at this page rather than at the comment thread, so those two are the version that sticks.

"How to test" needs the same correction. With a session location that is itself a queue location, the field prefills on main as well, so those steps pass either way and never show the bug. The case that fails on main is a login location that isn't queue-tagged but has a queue-tagged ancestor: blank on main, and with this change it picks up the ancestor the dashboard header resolved and stored.

For the title, maybe "(fix) O3-5958: Default the queue location to the one selected in Service Queues".

Nothing else from me.

Muta-Jonathan:
Thanks @dkayiwa have updated them and the After Video

dkayiwa (approved):

chibongho (approved):
LGTM

NethmiRodrigo (approved):
Thanks @Muta-Jonathan!

PR #2738, opened 2026-09-10
(feat) O3-5967: Show the queue number instead of the status in the default queue table

UjjawalPrabhat:
## Requirements

## Summary

The Status column tells you nothing: every table the default definition applies to is already scoped to one status, so it repeats the same value on every row under a heading that already says it. The queue number is what staff actually call out to the patient, and it's currently only visible on the public queue screen.

### Before
image

### After
image

## Related Issue

[O3-5967](https://openmrs.atlassian.net/browse/O3-5967)

## Other

[O3-5967]: https://openmrs.atlassian.net/browse/O3-5967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #2750, opened 2026-09-16
(fix) O3-5984: Label in-service patients with the configured status name

NethmiRodrigo:
## Requirements

## Summary

On the queue page, the metric tile for the in-service status reads "In Service" because it renders the status concept's display name, but the card list above the tables had a hardcoded "Attending" heading. The two labels came from different sources.

This PR derives the card list heading (and its error state title) from the configured `defaultTransitionStatus` concept via the existing `useConcept` hook, so it always matches the metric tile regardless of what the status concept is named or which locale is active. The clinic overview's "Attending" column header and totals card use the same label. While the concept is loading, the label falls back to "In Service".

The `attending` translation key is replaced by `inService`.

## Screenshots

Before (from the ticket):

before

After:
Screenshot 2026-09-16 at 9 25 55 PM
Screenshot 2026-09-16 at 9 26 05 PM

## Related Issue
https://openmrs.atlassian.net/browse/O3-5984


chibongho (approved):
LGTM.

packages/esm-service-queues-app/src/attending-patients/attending-patients.test.tsx
```diff
@@ -56,6 +62,22 @@ describe('AttendingPatients', () => {
...
+    mockUseConcept.mockReturnValue({
+      concept: { uuid: concepts.defaultTransitionStatus, display: 'In Service' } as Concept,
+      error: undefined,
+      isLoading: false,
+    });
+
+    render(<AttendingPatients />);
+
+    expect(mockUseConcept).toHaveBeenCalledWith(concepts.defaultTransitionStatus);
+    expect(screen.getByRole('heading', { name: 'In Service' })).toBeInTheDocument();
```
denniskigen:
```suggestion
      concept: { uuid: concepts.defaultTransitionStatus, display: 'Being seen' } as Concept,
      error: undefined,
      isLoading: false,
    });

    render(<AttendingPatients />);

    expect(mockUseConcept).toHaveBeenCalledWith(concepts.defaultTransitionStatus);
    expect(screen.getByRole('heading', { name: 'Being seen' })).toBeInTheDocument();
```

nit: Can we use a distinct display name here and update the assertion to match? The mocked display matches the loading fallback, so this test would still pass if the heading ignored the concept.

denniskigen (approved):
LGTM. Thanks, @NethmiRodrigo! One small persnickety suggestion on the test below.

PR #2752, opened 2026-09-16
(fix) O3-5983: Add the actions menu back to in-service patient cards

NethmiRodrigo:
## Requirements

## Summary

The new in-service patient cards on the queue page replaced the table rows that carried the **Move**, **Edit**, **Remove patient** and **Undo transition** actions, so those actions disappeared for in-service patients.

This PR adds a three-dot overflow menu to the queue status bar at the bottom of each in-service card with those four actions, in the same order as the old table menu. The items come from the queue table's own `useActionOverflowMenuItems` hook (exported from the actions cell), so the handlers, labels, delete styling and visibility rules are shared with the table rather than duplicated. **Undo transition** is only offered when the entry has a previous queue entry, as in the table.

The card used to be a single link. Now only the patient block (photo, name, sex and age) is the link to the patient chart, and hovering highlights just that block. The bottom bar with the queue name, priority tag and the menu is a sibling of the link, so it is not part of the clickable area and opening the menu never navigates.

## Screenshots

Before (from the ticket):

before

After (earlier iteration with the menu in the top-right corner; the menu now sits in the bottom bar, see the latest commit):

https://github.com/user-attachments/assets/8b9cb9ff-6204-4d4e-a64f-576c36a3a6f4

## Related Issue
https://openmrs.atlassian.net/browse/O3-5983


chibongho (commented):
Thanks for the quick fix!

Right now both the patient card and the 3-dot menu are clickable, with one inside the other. I wonder if we should consider not making the whole card clickable, but leave some right margin so the 3-dot menu's hitbox is separate from the card's. (For the record, the ward app's patient card also has the same problem, when showing the "Pending Transfer" indicator)

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -92,27 +94,38 @@ function AttendingPatientCard({ queueEntry }: { queueEntry: QueueEntry }) {
...
-        <span className={styles.service}>{queueEntry.queue?.display}</span>
-        <QueuePriority
-          priority={queueEntry.priority}
-          priorityComment={queueEntry.priorityComment ?? undefined}
-          priorityConfigs={priorityConfigs}
-        />
+    <div className={styles.card}>
+      <div className={styles.actionsMenu}>
+        <OverflowMenu iconDescription={t('actionsMenu', 'Actions menu')} flipped size="sm">
+          <ActionOverflowMenuItem actionKey="move" queueEntry={queueEntry} />
```
denniskigen: Can we render `OverflowMenuItem`s directly here? The shared wrapper drops the props and ref Carbon supplies for keyboard navigation, so ↓ stays on Move. I tested a shared hook that returns the visible items directly, and navigation works, including wrapping when Undo is hidden.

https://github.com/user-attachments/assets/0e99352b-ba45-4bb1-bf1b-3ddd7d50aeb1

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
```diff
@@ -92,27 +94,38 @@ function AttendingPatientCard({ queueEntry }: { queueEntry: QueueEntry }) {
...
+          <ActionOverflowMenuItem actionKey="edit" queueEntry={queueEntry} />
+          <ActionOverflowMenuItem actionKey="remove" queueEntry={queueEntry} />
```
denniskigen: nit: Can we add `isDelete: true` to the shared `remove` action? The confirmation already uses a danger button, and this would be consistent with Undo transition.

denniskigen (commented):
Thanks, @NethmiRodrigo, for restoring these actions and sharing the handlers with the table. I tested the flows locally and left a couple of suggestions.

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
NethmiRodrigo: Done. The wrapper is now a hook that returns the visible items as plain `OverflowMenuItem`s, and the queue table uses it too. The card test checks that ↓ moves from Move to Edit and wraps from Undo back to Move.

packages/esm-service-queues-app/src/attending-patients/attending-patients.component.tsx
NethmiRodrigo: Done. `isDelete: true` is on the shared remove action, so the item is red in the card menu and in the table menu.

NethmiRodrigo:
> Right now both the patient card and the 3-dot menu are clickable, with one inside the other. I wonder if we should consider not making the whole card clickable, but leave some right margin so the 3-dot menu's hitbox is separate from the card's. (For the record, the ward app's patient card also has the same problem, when showing the "Pending Transfer" indicator)

@chibongho Changed it so that the menu is now next to the link and not overlaid on top of it. A side effect though, that I'm not a big fan of, to make the divider line go all the way to the edge of the card I had to extend the row by 40 px wider. Unfortunately this means, that there is a 40px offset (the column where the menu is) where clicking won't do any action.

chibongho:
> > Right now both the patient card and the 3-dot menu are clickable, with one inside the other. I wonder if we should consider not making the whole card clickable, but leave some right margin so the 3-dot menu's hitbox is separate from the card's. (For the record, the ward app's patient card also has the same problem, when showing the "Pending Transfer" indicator)
> 
> @chibongho Changed it so that the menu is now next to the link and not overlaid on top of it. A side effect though, that I'm not a big fan of, to make the divider line go all the way to the edge of the card I had to extend the row by 40 px wider. Unfortunately this means, that there is a 40px offset (the column where the menu is) where clicking won't do any action.

@NethmiRodrigo How about like this?
image

Commit: 84b9cb1f

NethmiRodrigo:
> @NethmiRodrigo How about like this? image
> 

@chibongho doesn't this look a bit off with just the right side left out 😅?

chibongho:
@NethmiRodrigo How about moving the 3-dot menu to the queue status bottom bar, and have the bar not be part of the clickable area of the patient card?

NethmiRodrigo:
> @NethmiRodrigo How about moving the 3-dot menu to the queue status bottom bar, and have the bar not be part of the clickable area of the patient card?

@chibongho yeah, that's a pretty good idea, I like that! Thanks!

chibongho (approved):
Looks great!

PR #2758, opened 2026-09-18
(fix) O3-5991: Allow post-registration redirects to non-SPA URLs

cioan:
## Requirements

## Summary

`BeforeSavePrompt` unconditionally stripped the SPA base from its target URL before navigating:

    navigate({ to: `\${openmrsSpaBase}/${getUrlWithoutPrefix(target)}` });

`getUrlWithoutPrefix` splits on `getOpenmrsSpaBase()` and returns index [1], so a URL that does not live under the SPA base yields `undefined`, which the template literal stringifies. Any such target navigated to `/openmrs/spa/undefined`.

This affects implementations that bridge O3 to legacy OpenMRS pages: a `links.submitButton` such as
`${openmrsBase}/patientDashboard.form?patientId=${patientUuid}` was unusable, as was the `afterUrl` query parameter, which shares the path. Navigating away from a dirty registration form to a non-SPA page failed the same way.

The stripping is only needed for the `cancelNavigation` caller, where `target` comes from single-spa's `evt.detail.newUrl` and includes the origin. When there is no SPA-relative portion to extract, pass the URL through unchanged; `navigate` already routes non-SPA paths to `window.location.assign`, so no change is needed in esm-core.

## Screenshots
Screenshot 2026-09-18 at 4.24.12 PM

## Related Issue
none


packages/esm-patient-registration-app/src/patient-registration/before-save-prompt.component.tsx
```diff
@@ -62,7 +62,15 @@ const BeforeSavePrompt: FC<BeforeSavePromptProps> = ({ when, redirect }) => {
...
     if (typeof target === 'string') {
-      navigate({ to: `\${openmrsSpaBase}/${getUrlWithoutPrefix(target)}` });
+      const urlWithoutPrefix = getUrlWithoutPrefix(target);
+
+      // `target` does not necessarily point inside the SPA. It may be a legacy OpenMRS
+      // page, either configured via `links.submitButton` or navigated to from within the
+      // form. Such a URL has no SPA-relative portion to extract, so pass it through
+      // unchanged and let `navigate` choose between single-spa routing and a page load.
+      navigate({
+        to: typeof urlWithoutPrefix === 'string' ? `\${openmrsSpaBase}/${urlWithoutPrefix}` : target,
```
chibongho, line 72: I don't think we really need `getUrlWithoutPrefix()` to strip the `openmrsSpaBase` prefix and then re-add it back it here, but it's possible it handles an edge case I haven't considered. I'm ok with keeping the existing behavior for now.

chibongho (approved):
Looks good.

PR #2761, opened 2026-09-19
(fix) O3-5912: Clear dependent address fields when a parent level changes

Faisal77666:
## Requirements

- [x] My work is based on designs, which are linked or shown either in the Jira ticket or the description below. N/A, no design change.

## Summary

With `searchAddressByLevel` enabled, changing a parent address level left the child values from the previous selection in place, so the form could end up with inconsistent addresses (for example, Uganda with Nairobi as the province).

`useAddressEntryFetchConfig` took `setFieldValue` from `usePatientRegistrationContext()`. Every address level renders inside `AddressComponentContainer`, which provides its own `PatientRegistrationContextProvider` whose `setFieldValue` is `async () => {}`. `updateChildElements()` therefore ran on every selection and did nothing.

This change takes `setFieldValue` from Formik (`useFormikContext`) in the hook instead. Formik is already reachable at that depth, since the hook calls `useField('address')`. The container's stubbed context is left as it is, so the embedded address widget stays isolated, and behaviour without `searchAddressByLevel` is unchanged.

### Tests

New tests in `address-hierarchy-levels.test.tsx` render the real `AddressComponent`, drive the real combo inputs, and assert on the resulting Formik values. Hierarchy data comes from a mocked `openmrsFetch`, so the hierarchy hooks themselves are not mocked.

- Changing the country clears every dependent level.
- Changing a middle level clears only the levels below it and keeps its parent and the newly selected value.
- With `searchAddressByLevel` disabled, changing the country leaves other fields alone and makes no hierarchy entry requests.

The first two tests fail without the fix. Selecting Kenya, Nairobi and Westlands, then changing the country to Uganda, produced:
country: "Uganda"
stateProvince: "Nairobi" (expected "")
cityVillage: "Westlands" (expected "")

## Screenshots

N/A. I couldn't reproduce this in the browser: the dev3 backend returns no address hierarchy levels, so the tests drive the same interaction with mocked hierarchy data.

## Related Issue

https://openmrs.atlassian.net/browse/O3-5912

## Other

Verified locally with the registration package's tests, `tsc` and `eslint`.


Faisal77666:
@denniskigen @VeronicaMuthee @ibacher @NethmiRodrigo  PTAL

